PR #32, opened 2022-06-07
feat(server): reimplement bundling, watcher, script runner

StarpTech:
- reimplement bundling, watcher and script execution. Move logic to different modules.
- Track uncaught Node.js via https://nodejs.org/api/process.html#event-uncaughtexceptionmonitor
- Graceful shutdown hooks server 

#### Checklist

- [x] run `make test`


cli/commands/start.go
```diff
@@ -9,10 +9,10 @@ import (
 	"path/filepath"
 	"time"
 
+	cmd2 "github.com/go-cmd/cmd"
```
jensneuse, line 12: maybe `gocmd`?

cli/commands/start.go
```diff
@@ -48,7 +48,12 @@ If used without --exclude-server, make sure the server is available in this dire
 			return err
 		}
 
-		var hooksBundler *bundleconfig.Bundler
+		quit := make(chan os.Signal, 1)
+		signal.Notify(quit, os.Interrupt)
```
jensneuse, line 52: Should we add `syscall.SIGTERM` as well?
https://stackoverflow.com/a/16538055

cli/commands/start.go
```diff
@@ -60,27 +65,48 @@ If used without --exclude-server, make sure the server is available in this dire
...
+				fmt.Sprintf("HOOKS_TOKEN=%s", hooksJWT),
+				fmt.Sprintf("WG_MIDDLEWARE_PORT=%d", middlewareListenPort),
+				fmt.Sprintf("WG_LISTEN_ADDR=%s", listenAddr),
+			)
+
+			hooksServerDone = make(chan struct{})
+
+			go func() {
+				defer close(hooksServerDone)
+				// Done when both channels have been closed
```
jensneuse, line 88: are you sure?

cli/commands/start.go
StarpTech: Yes :+1:

cli/commands/start.go
StarpTech: Will reiterate on this, thanks.

cli/commands/start.go
StarpTech: Works and sounds reasonable with the attached references. Code from https://github.com/go-cmd/cmd/blob/master/examples/blocking-streaming/main.go#L28

cli/commands/start.go
StarpTech: Added

PR #50, opened 2022-06-22
fix: anonymous query 'operationName' wrong parsing value when null

manini-i3hub:
FIX anonymous query 'operationName' wrong parsing value when null


StarpTech (approved):
Great catch!

pkg/apihandler/apihandler.go
```diff
@@ -660,9 +660,13 @@ func (h *GraphQLHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	body := buf.Bytes()
 
 	requestQuery, _ := jsonparser.GetString(body, "query")
-	requestOperationName, _, _, _ := jsonparser.Get(body, "operationName")
+	requestOperationName, parsedOperationNameDataType, _, _ := jsonparser.Get(body, "operationName")
 	requestVariables, _, _, _ := jsonparser.Get(body, "variables")
 
+	if parsedOperationNameDataType == jsonparser.Null {
```
StarpTech, line 666: Could you please add a comment like

> 	// An operationName set to { "operationName": null } has a byte length. In that case, we need to reset the length
	// to ensure that the operationName is only set, if it is not null.

StarpTech (changes requested):

StarpTech:
Please sync the branch with main.

PR #54, opened 2022-06-29
feat: improve error message when graphql introspection fails

jensneuse:
This PR improves error messages when running `wunderctl up`.
Originally, we've just thrown an Error / Rejected Promise, which results in a very verbose error message where it's hard to find the real issue.

With this change, we'll catch the error, do a simple console.log of the error message and exit with code 1, stopping `wunderctl up` immediately. This gives a simple and clear error message.

### DIFF
--- a/packages/sdk/src/definition/index.ts
+++ b/packages/sdk/src/definition/index.ts
@@ -14,7 +14,7 @@ import { mergeApis } from './merge';
 import * as fs from 'fs';
 import { openApiSpecificationToRESTApiObject } from '../v2openapi';
 import { renameTypeFields, renameTypes } from '../graphql/renametypes';
-import axios, { AxiosRequestConfig } from 'axios';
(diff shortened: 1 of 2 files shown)


packages/sdk/src/definition/index.ts
```diff
@@ -928,9 +928,20 @@ const introspectGraphQLAPI = async (
 	}
 
 	// TODO single place where a HTTP client is configured for all data sources
-	const res = await axios.post(resolveVariable(introspection.url), data, opts);
+	let res: AxiosResponse | undefined;
+	try {
+		res = await axios.post(resolveVariable(introspection.url), data, opts);
+	} catch (e: any) {
+		console.log(`introspection failed, error: ${e.message}`);
```
StarpTech, line 935: Could you add which introspection has failed?

PR #56, opened 2022-06-29
fix: don't (re-)fetch user server & client-side when authentication is disabled

jensneuse:
This PR improves the nextjs package.

When no authentication provider is configured (cookie based auth), there's no need to fetch the user server-side as well as client-side, incl. re-fetching. This PR disables all user fetches in this case.

### DIFF
--- a/packages/nextjs-testapp/.wundergraph/operations/ProtectedWeather.graphql
+++ b/packages/nextjs-testapp/.wundergraph/operations/ProtectedWeather.graphql
@@ -1,25 +1,25 @@
-query ($forCity: String!) {
-  getCityByName: weather_getCityByName(name: $forCity) {
-    name
-    coord {
-      lat
-      lon
-    }
-    weather {
-      summary {
-        title
-        description
-        icon
-      }
-      temperature {
-        actual
-        min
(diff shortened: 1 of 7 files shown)


StarpTech:
Is it possible to implement it in the same way for the wundergraph client?

jensneuse:
> Is it possible to implement it in the same way for the wundergraph client?

The client by itself doesn't refetch. It's the nextjs wrapper that does by handling the application lifecycle with hooks.

StarpTech:
I mean to avoid [this](https://github.com/wundergraph/wundergraph/blob/main/packages/sdk/src/codegen/templates/typescript/web.client.template.ts#L266) when no auth was configured.

jensneuse:
> I mean to avoid [this](https://github.com/wundergraph/wundergraph/blob/main/packages/sdk/src/codegen/templates/typescript/web.client.template.ts#L266) when no auth was configured.

It's a different client btw. has nothing to do with nextjs. Won't fix in this PR.

StarpTech:
ok

StarpTech (approved):
LGTM

PR #63, opened 2022-07-04
feat: add introspection caching & DataSource polling

jensneuse:
- makes wunderctl faster because for each file change (e.g. Operation) introspection of all DataSources is not necessary anymore
- allows for synchronization of DataSources and the generated Schema: E.g. if you continuously modify a Database, it can be introspected every 5 seconds to auto-update the Virtual Graph


jensneuse:
@StarpTech please review

cli/commands/up.go
```diff
@@ -49,13 +50,32 @@ var upCmd = &cobra.Command{
...
 			abstractlogger.String("commit", BuildInfo.Commit),
 			abstractlogger.String("date", BuildInfo.Date),
 			abstractlogger.String("builtBy", BuildInfo.BuiltBy),
 		)
 
+		introspectionCacheDir := path.Join(wundergraphDir, "generated", "introspection", "cache")
+		_, errIntrospectionDir := os.Stat(introspectionCacheDir)
+		if errIntrospectionDir == nil {
+			if clearIntrospectionCache {
+				err = os.RemoveAll(introspectionCacheDir)
```
StarpTech, line 66: Could you add an info statement here?

packages/default-testapp/.wundergraph/wundergraph.config.ts
```diff
@@ -12,6 +12,9 @@ import operations from './wundergraph.operations';
 const spaceX = introspect.graphql({
 	apiNamespace: 'spacex',
 	url: 'https://api.spacex.land/graphql/',
+	introspection: {
```
StarpTech, line 15: Please revert the example so it keeps the "most default" in the repo.

pkg/scriptrunner/scriptrunner.go
```diff
@@ -84,7 +84,7 @@ func (b *ScriptRunner) Run(ctx context.Context) chan struct{} {
 			)
 		case <-b.cmd.Done():
 			status := b.cmd.Status()
-			if status.Error != nil {
+			if status.Error != nil && status.Exit > 0 {
```
StarpTech, line 87: Could you add a comment?

cli/commands/up.go
```diff
@@ -97,11 +121,31 @@ var upCmd = &cobra.Command{
 		<-configBundler.BuildDoneChan
 
 		<-configRunner.Run(ctx)
+		// only start watching in the builder once the initial config was built and written to the filesystem
+		close(deferWatchConfigBuilder)
```
StarpTech, line 125: I think it would be simpler if you split Bundle and watch in two different methods. You could still have a bundle method that do both.

StarpTech (changes requested):

cli/commands/up.go
jensneuse: I don't want to over-engineer this.

StarpTech: That's what my request is about. Using a channel, in that case, is not needed and is hard to grasp.

packages/default-testapp/.wundergraph/wundergraph.config.ts
jensneuse: I'd like this to always run so that we notice when it's broken.

PR #64, opened 2022-07-04
fix(codegen): detect internal input correctly

StarpTech:
#### Checklist

- [ ] run `make test`


jensneuse (changes requested):
I assume that the nextjs package needs this fix as well, right?

StarpTech:
> I assume that the nextjs package needs this fix as well, right?

Looks fine. It uses `VariablesSchema` to filter for operation with input.

jensneuse (approved):

PR #65, opened 2022-07-04
fix(start): error when config cant be found

StarpTech:
#### Checklist

- [ ] run `make test`


jensneuse (approved):
It's not idiomatic go to have a "helpers" package. Instead of "helpers.FileExists", it should be e.g. "files.Exists(filePath). You can chose to refactor or merge as is. I just recommend to follow idiomatic go patterns.

PR #68, opened 2022-07-06
fix: restart hooks server, ensure __wg exists

StarpTech:
#### Checklist

- [ ] run `make test`

### DIFF
--- a/cli/commands/up.go
+++ b/cli/commands/up.go
@@ -42,6 +42,11 @@ var upCmd = &cobra.Command{
 		quit := make(chan os.Signal, 2)
 		signal.Notify(quit, os.Interrupt, syscall.SIGTERM)
 
+		wd, err := os.Getwd()
+		if err != nil {
+			log.Fatal("Could not get your current working directory")
+		}
+
 		hooksJWT, err := apihandler.CreateHooksJWT(secret)
 		if err != nil {
 			return err
@@ -73,23 +78,8 @@ var upCmd = &cobra.Command{
 		}
 
 		configJsonPath := path.Join(wundergraphDir, "generated", "wundergraph.config.json")
-
 		configOutFile := path.Join(wundergraphDir, "generated", "bundle", "config.js")
-		configBundler := bundler.NewBundler(bundler.Config{
-			Name:       "config-bundler",
-			EntryPoint: entryPoint,
-			OutFile:    configOutFile,
-			Logger:     log,
-			WatchPaths: []string{
-				path.Join(wundergraphDir, "operations"),
-				// a new cache entry is generated as soon as the introspection "poller" detects a change in the API dependencies
-				// in that case we want to rerun the script to build a new config
-				introspectionCacheDir,
-			},
-			IgnorePaths: []string{
-				"node_modules",
-			},
-		})
+		serverOutFile := path.Join(wundergraphDir, "generated", "bundle", "server.js")
 
 		configRunner := scriptrunner.NewScriptRunner(&scriptrunner.Config{
 			Name:       "config-runner",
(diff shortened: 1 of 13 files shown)


cli/commands/up.go
```diff
@@ -205,39 +131,77 @@ var upCmd = &cobra.Command{
...
-				}
-			}()
 
-			hooksBundler.BundleAndWatch(ctx)
+			onAfterBuild = func() {
+				log.Debug("Configuration change detected")
+				// generate new config
+				<-configRunner.Run(ctx)
+
+				// bundle server
```
jensneuse, line 147: bundle hooks

cli/commands/up.go
```diff
@@ -205,39 +131,77 @@ var upCmd = &cobra.Command{
...
+				hooksBundler.Bundle()
+				go func() {
+					// run or restart server
+					<-hookServerRunner.Run(ctx)
```
jensneuse, line 151: does Run kill and restart automatically?

cli/commands/up.go
```diff
@@ -268,8 +232,18 @@ var upCmd = &cobra.Command{
 			log.Info("Received interrupt signal. Initialize WunderNode shutdown ...",
 				abstractlogger.String("signal", signal.String()),
 			)
+			configRunner.Stop()
```
jensneuse, line 235: Can you not just move the Stop logic below the select case? no need to duplicate it I think. select case will block until one of signal or ctx done returns. So this logic can go above n.Close (249)

packages/sdk/src/middleware/server.ts
```diff
@@ -193,19 +191,19 @@ export const startServer = async (
...
 			req.ctx = {
 				log: req.log.child({ plugin: 'hooks' }),
-				user: req.body.__wg.user,
+				user: req.body.__wg?.user,
 				// clientRequest represents the original client request that was sent initially to the server.
 				clientRequest: {
 					headers: new Headers(req.body.__wg.clientRequest?.headers),
-					requestURI: req.body.__wg.clientRequest?.requestURI || '',
-					method: req.body.__wg.clientRequest?.method || 'GET',
+					requestURI: req.body.__wg?.clientRequest?.requestURI || '',
```
jensneuse, line 203: why __wg? when the definition in line 194 says that __wg will always be set? Make it optional in 194 or remove the '?'?

pkg/apihandler/apihandler.go
```diff
@@ -2114,7 +2114,7 @@ func getOperationMetaData(r *http.Request) *OperationMetaData {
 
 func hookBaseData(r *http.Request, buf []byte, variables []byte, response []byte) []byte {
 	buf = buf[:0]
-	buf = append(buf, []byte("{}")...)
+	buf = append(buf, []byte(`{ "__wg": {} }`)...)
```
jensneuse, line 2117: no need for whitespace in JSON

pkg/bundler/bundler.go
```diff
@@ -41,6 +41,7 @@ type Config struct {
 	WatchPaths            []string
 	IgnorePaths           []string
 	OutFile               string
+	OnAfter               func()
```
jensneuse, line 44: Can we be a bit more specific here? e.g. OnAfterBuild if that's the intention?

pkg/scriptrunner/scriptrunner.go
```diff
@@ -87,6 +87,14 @@ func (b *ScriptRunner) Run(ctx context.Context) chan struct{} {
 			// exit code == -1 means the script was killed by a signal
 			// this is intentional and not an error and happens when we re-start the process after a watched file has changed
 			if status.Exit == -1 {
+				b.log.Error("Script exited",
```
jensneuse, line 90: Should this be logged as an error? It's intentional. Maybe Debug?

jensneuse (changes requested):

cli/commands/up.go
StarpTech: Yes

pkg/scriptrunner/scriptrunner.go
StarpTech: I wanted to handle it as debug. Good catch.

cli/commands/up.go
StarpTech: I like.

jensneuse (approved):

PR #72, opened 2022-07-12
feat: extract typescript client from nextjs

Pagebakers:
#### Checklist

- [x] run `make test`


Pagebakers:
Will move the client in the sdk package as discussed with Dustin.

Pagebakers:
This is ready for review now.

The client can be imported from `@wundergraph/sdk/client`. The SDK doesn't export the client directly yet, since there are some conflicting types with the internal client.

StarpTech (changes requested):

packages/sdk/src/client/index.ts
```diff
@@ -0,0 +1,777 @@
+import hash from 'object-hash';
```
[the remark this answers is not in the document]
Pagebakers: Oh you're right! I forgot to copy this over.

Pagebakers:
Added and update the Readme.

`typesVersions` seems to fix the type issues with exports. Imports from 'sdk/dist/...' work still as well.

@StarpTech  I'd recommend you try this branch as well, to see if the new exports work on your machine too.

StarpTech:
Great, I'll try it out.

Pagebakers:
@jensneuse I've added types to the exports to make sure the ts compiler can find them.

Running make won't fail anymore, since the nextjs-testapp has been removed from the build script.

But.. after a fresh clone and running make, trying to build the nextjs-testapp will fail, complaining about not being able to find the client types. Once you run `pnpm i` inside the testapp, it does work as expected. To me this seems more like an issue with the nextjs-testapp in the monorepo then the types not being exported correctly from the sdk.

Pagebakers:
@jensneuse 

It's working fine from NPM, I used a local verdaccio server to publish @wundergraph/nextjs and @wundergraph/sdk and installed it in the nextjs starter without any issues.

If the types and response handling refactoring is OK, we are good to go.

packages/sdk/src/client/README.md
```diff
@@ -0,0 +1,106 @@
...
+});
+```
+
+### Run a query
+
+```ts
+const result = await client.query(
+  {
+    operationName: 'Hello',
+    requiresAuthentication: false,
```
jensneuse, line 36: we don't actually make use of "requiresAuthentication" in the pure ts client
I think we should remove it from the config and also the docs, it seems confusing
I've just checked the code, we don't do anything with it, it's only used by the nextjs package

jensneuse (changes requested):

packages/sdk/src/client/README.md
Pagebakers: Makes sense! Since 'operationName' is the only param in this case. We can change the client to
```client.query('Hello', {input: { hello: 'world' }})```

Like my proposal in https://github.com/wundergraph/wundergraph/discussions/80

Comment (Pagebakers) [reply packages/sdk/src/client/README.md]: Removed requiresAuthentication and moved isLiveQuery to InternalSubscriptionArgs.

I kept the client api the same, as it would introduce too many changes and I'd need to rewrite a lot of the Next.js api, which makes no sense, since I will be creating a new pure React hook client anyway.

File: packages/sdk/src/client/README.md [hunk 3]
```diff
@@ -0,0 +1,118 @@
...
+      name: 'WunderGraph',
+    },
+  }
+);
+```
+
+### LiveQuery
+
+```ts
+const result = await client.subscribe(
```
jensneuse, line 63: I think this is wrong. It's not returning a result. Also, it needs a callback when a message comes back, right?

packages/sdk/src/client/README.md
```diff
@@ -0,0 +1,118 @@
...
+      name: 'World',
+    },
+  }
+);
+```
+
+### Subscription
+
+```ts
+const result = await client.subscribe(
```
jensneuse, line 79: I think this is the same problem as with the live query.

jensneuse (changes requested):

packages/sdk/src/client/README.md
```diff
@@ -0,0 +1,120 @@
+# Wundergraph TypeScript Client
```
StarpTech, line 1: This client is built for Web and Server correct? It would be good to mention that here.

packages/sdk/src/client/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+![wundergraph-client](https://img.shields.io/npm/v/@wundergraph/sdk.svg)
+
+This is the base implementation of the WunderGraph HTTP protocol in TypeScript.
+It's used as the base interface for the Web client, React and Next.js implementations.
+
+Use it to build your own 3rd party integration with WunderGraph.
+
+## Getting Started
```
StarpTech, line 10: Could you also add an example to pass a custom fetch implementation? Ideally Web and Server example.

packages/sdk/src/client/index.ts
```diff
@@ -0,0 +1,668 @@
...
+					cb(this.errorResult(response));
+					return;
+				}
+
+				const reader = response.body.getReader();
+				const decoder = new TextDecoder();
+				let message: string = '';
+				while (true) {
+					const { value, done } = await reader.read();
+					if (done) break;
```
StarpTech, line 240: I think as a consumer it is valuable to know when the stream ends. Should we pass an object `{ value, done }` to the callback?

packages/sdk/src/client/index.ts
```diff
@@ -0,0 +1,668 @@
...
+					if (!value) continue;
+					message += decoder.decode(value);
+					if (message.endsWith('\n\n')) {
+						const responseJSON = message.substring(0, message.length - 2);
+						cb(this.jsonToSubscriptionResult(responseJSON));
```
StarpTech, line 245: By convention, the first argument of a callback is the error.  We should work with one object `{ value, done, error }` or passing it as two arguments `(err, { value, done }) =>`

packages/sdk/src/client/index.ts
```diff
@@ -0,0 +1,668 @@
...
+					},
+					method: 'GET',
+					credentials: 'include',
+					mode: 'cors',
+					signal: abortSignal,
+				}
+			);
+			if (this.isOK(response)) {
+				return response.json();
+			}
```
StarpTech, line 450: As a consumer, I'd expect to get an error rather than nothing. Let's return the response or some information for debugging?

packages/sdk/src/client/index.ts
```diff
@@ -0,0 +1,668 @@
...
+		} catch {}
+		return null;
+	};
+
+	public login = (authProviderID: string, redirectURI?: string) => {
+		const query = this.queryString({
+			redirect_uri: redirectURI || window.location.toString(),
+		});
+		window.location.assign(`${this.baseURL}/${this.applicationPath}/auth/cookie/authorize/${authProviderID}${query}`);
```
StarpTech, line 459: This will throw on the server right? We should check the existence.

packages/sdk/src/client/index.ts
```diff
@@ -0,0 +1,668 @@
...
+		);
+		return this.isOK(response);
+	};
+
+	private queryString = (input?: Object): string => {
+		if (!input) {
+			return '';
+		}
+		const query = (Object.keys(input) as Array<keyof typeof input>)
+			// @ts-ignore
```
StarpTech, line 484: Can you get rid of this?

StarpTech (changes requested):

StarpTech:
How / did you test the whole implementation?

packages/sdk/src/client/index.ts
```diff
@@ -0,0 +1,668 @@
...
+						message: e.message,
+					},
+				],
+			};
+		}
+	}
+
+	public subscribe = <S extends SubscriptionProps, Input, Data>(
+		subscription: S,
+		cb: (response: SubscriptionResult<Data>) => void,
```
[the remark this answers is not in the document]
Pagebakers: I considered this, but I didn't want to introduce breaking changes in this PR.

packages/sdk/src/client/index.ts
Pagebakers: This is more like an event handler, so I don't feel like this pattern applies here.

packages/sdk/src/client/index.ts
Pagebakers: I don't think we should add this to the callback parameters, as this is the result object, rather keep it consistent with all other methods.

Perhaps we can bind it to the scope of the function.

```ts
function callback() {
  console.log(this.state) // done // streaming // initializing?
}
```

But perhaps subscribe

packages/sdk/src/client/index.ts
Pagebakers: I suppose we should throw an error in this case. But this brings the next question, should we throw every where instead of returning a result with the error message.

Pagebakers:
> How / did you test the whole implementation? (Upload, Subscribe, ...)

I tested all functionality (except subscriptions) using the nextjs-testapp

packages/sdk/src/client/index.ts
StarpTech: After reading it a second time, I think you're right. However, I forget we use the same response shape as for graphql errors.

packages/sdk/src/client/index.ts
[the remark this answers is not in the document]
StarpTech: As far as I see, the consumer does not use the WunderGraph Client directly in the Next.js package. You can easily swap the parameter order.

packages/sdk/src/client/index.ts
Pagebakers: Done

packages/sdk/src/client/index.ts
StarpTech: Generally, yes. But for now, I'd try to do it only for low-level methods like the client and handle it in the same way as now in the consumer packages (next.js, react ...). So if someone uses this client package, he has full control over handling an error.

Pagebakers: Introducing this will break a thing or two in the next.js client. I wanted to prevent that in this PR. And only introduce  any breaking changes in the new React package.

StarpTech: Could you point me to the relevant places?

pnpm-lock.yaml
```diff
@@ -11324,6 +11324,7 @@ packages:
       react: ^16.8.0 || ^17.0.2 || ^18.0.0
       react-dom: ^16.8.0 || ^17.0.2 || ^18.0.0
     dependencies:
+      '@wundergraph/client': link:packages/client
```
StarpTech, line 11327: Do you need to rerun `pnpm i`?

packages/sdk/src/client/index.ts
Pagebakers: https://github.com/wundergraph/wundergraph/blob/main/packages/nextjs/src/index.ts#L797

This is also mentioned in a blog post somewhere, so that should be updated too if we change this.

But.. it already catches any error errors in this code, perhaps the change is as easy as `setUser(null)` in the catch, in case the call fails.

StarpTech: Blog post is irrelevant. We can't update it after every change :smile: 

> But.. it already catches any error errors in this code, perhaps the change is as easy as setUser(null) in the catch, in case the call fails.

Yes, that should do the trick.

pnpm-lock.yaml
Pagebakers: Yeah because we added @wundergraph/client

I also had to resolve a type issue because there are multiple versions of @types/react installed.

packages/sdk/src/client/index.ts
Pagebakers: Let's keep this as is for now, and open a new issue once there is a need for having access to a 'done' event.

StarpTech (approved):
LGTM

jensneuse (approved):

PR #73, opened 2022-07-12
fix: nextjs client fails hard when a page misses "withWunderGraph"

Pagebakers:
#### Checklist

- [x] run `make test`

I think we should still throw an error when the WunderGraph context is missing, otherwise developers might not realise what's going on.

This PR will make sure the Nextjs client throws an error when the client in the WunderGraphContext is undefined with a clear message to instruct the developer how to fix the problem.

Screenshot 2022-07-12 at 16 06 42


packages/nextjs/src/index.ts
```diff
@@ -1013,6 +1013,33 @@ export interface MutationResultError {
...
+	name?: string;
+}
+
+function useWunderGraphContext<Role>(
+	context: Context<WunderGraphContextProperties<Role>>,
+	options: UseWunderGraphContextOptions = {}
+) {
+	const {
+		strict = true,
+		name = 'useQuery',
```
jensneuse, line 1028: should it be useQuery by default? Should there be a default value? Or would it not make more sense to force the user of useWunderGraphContext to set a name?

packages/nextjs/src/index.ts
```diff
@@ -1013,6 +1013,33 @@ export interface MutationResultError {
...
+		errorMessage = `${name} 'client' is undefined. Did you forget to wrap your app or page with 'withWunderGraph'?`,
```
jensneuse, line 1029: Is this message accurate? Is the main problem that client is undefined  or the missing Context?

jensneuse (changes requested):

packages/nextjs/src/index.ts
Pagebakers: `useWunderGraph` would be a better default, since that is the most basic context consumer. I don't think we need to force this.

packages/nextjs/src/index.ts
Pagebakers: The main problem is that the ContextProvider is missing. The context has a default value, so the best way to check if the the app/page is wrapped with withWundreGraph or WunderGraphProvider (for React apps) is by checking if the client has been set.

Pagebakers:
The default name is `useWunderGraph` now.

Pagebakers:
@jensneuse this can be merged

StarpTech:
There is a failing job.

Pagebakers:
> There is a failing job.

Hm i had this locally too in the other branch, because there are multiple @types/react versions installed.

StarpTech:
Are you able to fix it?

Pagebakers:
> Are you able to fix it?

Yep it was fixed in the other PR,  fixed now with the merge

jensneuse (approved):

PR #76, opened 2022-07-13
fix(auth): pass raw access token to hook

StarpTech:
- [X] Pass the raw access token to the hooks
- [X]  Adds examples that make use of the access token.

#### Checklist

- [ ] run `make test`


examples/rbac/.wundergraph/wundergraph.server.ts
```diff
@@ -0,0 +1,43 @@
...
+				const { data } = await client.queries.HasWunderGraphStarred();
+
+				// Only allow the user to access the server if they have starred the repo
+				if (
+					// Error are communicated via GraphQL unions
+					data.github_checkRepoIsStarred.__typename === 'github_BasicErrorRequiresAuthentication'
+				) {
+					log.error(`Could not authorize user ${user.name}`, data.github_checkRepoIsStarred.message);
+				} else {
+					roles.push('stargazer');
```
jensneuse, line 28: I think this is wrong. Should check if the response is 204, not just "not" auth error

pkg/authentication/authentication.go
```diff
@@ -300,7 +303,7 @@ func bearerTokenToJSON(token string) ([]byte, error) {
 	return payload, nil
 }
 
-func mustBearerTokenToJSON(token string) []byte {
+func tryParseJWTPayload(token string) []byte {
```
jensneuse, line 306: rename to: "tryParseJWT"?

jensneuse (changes requested):

pkg/authentication/authentication.go
StarpTech: Done

jensneuse (approved):

PR #78, opened 2022-07-13
feat: replace the legacy client with the new implementation

Pagebakers:
#### Checklist

- [x] run `make test`

## What's new

The legacy client now uses the new client internally and is backwards compatible, so this should not introduce breaking changes.

## Tested methods

- [x] useQuery
- [x] useMutation
- [x] useLiveQuery
- [ ] useSubscription
- [x] uploadFiles
- [x] login
- [x] logout
- [x] fetchUser

This pr is still based of #72 , I can also open a new PR once 72 is merged into main.


jensneuse (changes requested):
@Pagebakers do you think we can make the client code less repetitive in terms of handling response status codes? Currently, we're doing it multiple times in slightly different ways, but similar for queries and subscriptions as well as uploads. Maybe we can have a generic Response Object with enums for the different states, like error, partial, unauthorized, etc... The generic type will contain content that's specific to the operation. It would be slightly breaking, but I think it's worth cleaning this up. WDYT? cc @StarpTech

Pagebakers:
> @Pagebakers do you think we can make the client code less repetitive in terms of handling response status codes? Currently, we're doing it multiple times in slightly different ways, but similar for queries and subscriptions as well as uploads. Maybe we can have a generic Response Object with enums for the different states, like error, partial, unauthorized, etc... The generic type will contain content that's specific to the operation. It would be slightly breaking, but I think it's worth cleaning this up. WDYT? cc @StarpTech

We could have a base interface with all shared statuscodes that query/mutate/subscribe extend with their own properties. This wont break anything.

The response handling in the pr is to not break the legacy client, eg translating the new result to the old response type. But this will be deprecated and replaced with a new client with the new Result structure.

StarpTech:
I'm always for simplifying things :smile: @Pagebakers If you can do it without breaking that would be nice. Maybe just create a proposal and show us in zoom?

StarpTech:
I can't run `make` it looks like `nextjs-testapp` can't find the client.

```
│ Failed to compile.
│ .wundergraph/generated/wundergraph.client.ts:27:8
│ Type error: Cannot find module '@wundergraph/sdk/client' or its corresponding type declarations.
│   25 |        MutationResult,
│   26 |        SubscriptionResult,
│ > 27 | } from "@wundergraph/sdk/client";
│      |        ^
│   28 | 
│   29 | export const WUNDERGRAPH_S3_ENABLED = true;
│   30 | export const WUNDERGRAPH_AUTH_ENABLED = true;
└─ Failed in 4.6s

```

Pagebakers:
> Running into this error now too.
> 
> ```shell
> uncaught exception, origin: uncaughtException, error: Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './dist/codegen/templates/typescript/web.client' is not defined by "exports" in /Users/eelco/Development/wundergraph/wundergraph/packages/nextjs-testapp/node_modules/@wundergraph/sdk/package.json
> node:internal/modules/cjs/loader:488
>       throw e;
>       ^
> ```
> 
> I was expecting this would happen when introduction `exports` in the package.json. You can no longer import from `@wundergraph/sdk/dist...` which I suspect might break a few things.

Added /dist/* to exports to resolve this.

packages/sdk/src/client/index.ts
```diff
@@ -0,0 +1,681 @@
...
+	custom_attributes?: string[];
+	custom_claims?: {
+		[key: string]: any;
+	};
+	access_token?: JSONObject;
+	id_token?: JSONObject;
+	raw_id_token?: string;
+}
+
+export class WunderGraphClient<Role> {
```
StarpTech, line 45: This is the same implementation as in PR #72 ?

StarpTech (changes requested):

packages/sdk/src/client/index.ts
Pagebakers: This PR is based of #72

Pagebakers:
Fully up to date with #72 now

packages/sdk/src/client/index.ts
```diff
@@ -530,30 +510,36 @@ export interface QueryArgsWithInput<Input> extends QueryArgs {
 	input: Input;
 }
 
-export interface QueryProps {
+// export interface QueryProps<Input, Args = QueryArgs> {
```
StarpTech, line 513: Any reason to stay it here?

packages/sdk/src/client/index.ts
```diff
@@ -0,0 +1,671 @@
...
+
+export interface InternalQueryArgsWithInput<Input> extends InternalQueryArgs {
+	input: Input;
+}
+
+export interface QueryArgsWithInput<Input> extends QueryArgs {
+	input: Input;
+}
+
+// export interface QueryProps<Input, Args = QueryArgs> {
```
StarpTech: Any reason to keep it?

StarpTech (approved):
LGTM

packages/sdk/src/client/index.ts
Pagebakers: I actually dreamed about this last night, I forgot to remove it :D

jensneuse (approved):

PR #81, opened 2022-07-14
fix: handle cli paths correctly

### LINKED ISSUE #17: up mode, file path error.
### Bug description

`up` mode, file path error.

Expect add `AbsWorkingDir` at `initialBuild` function in `bundleconfig.go` file.

image

### How to reproduce

1. Go to '/cmd/wunderctl'
2. Change 'nothing'
3. Run 'go run main.go up --debug true --wundergraph-dir ../../examples/simple/.wundergraph'

```sh
2022-05-28T15:12:38+08:00       debug   starting without env file
2022-05-28T15:12:38+08:00       info    starting WunderNode     {"version": "dev", "commit": "unknown", "date": "unknown", "builtBy": "unknown"}
2022-05-28T15:12:38+08:00       fatal   Config Bundler: initial build failed    {"bundler": "config", "outFile": "../../examples/simple/.wundergraph/generated/bundle/config.js", "errors": [{"PluginName":"","Text":"The entry point \"wundergraph.config.ts\" cannot be marked as external","Location":null,"Notes":null,"Detail":null}]}
exit status 1
```

### Expected behavior

_No response_

### Environment & setup

- OS: Mac OS
- Go version: go version go1.17.5 darwin/amd64
- Node.js version: v16.13.1

### WunderCtl Version

2b16f968f7c18e1f3bb9dd205de21336220070bb

StarpTech:
This PR fix #17. All paths are properly resolved based on the configured base directory. Required files are validated before a command is run. We no longer search recursively for `.wundergraph` to avoid issues in monorepo and .env handling. The proposed implementation can deal with 3 ways.

- Run any wunderctl command on the same directory as `.wundergraph`
- Run any wunderctl command inside `.wundergraph`
- Run any wunderctl command and pass the base directory explicitly with `--wundergraph-dir=my/path/to/.wundergraph`

I also moved some global flags to the specific commands.

Breaking change: No longer possible to add a full path for `-c ../wundergraph.config.json`. Users can overwrite the base directory and config json filename separately.

#### Checklist

- [ ] run `make test`


cli/commands/add.go
```diff
@@ -15,6 +17,10 @@ var addCmd = &cobra.Command{
 	Long:    `wunderctl add spacex/spacex jens/weather stripe/stripe`,
 	Args:    cobra.MinimumNArgs(1),
 	Run: func(cmd *cobra.Command, dependencies []string) {
+		if !files.DirectoryExists(wundergraphDir) {
+			log.Fatal(`could not find base directory`, abstractlogger.String("dir", wundergraphDir))
```
jensneuse, line 21: Is the intention to stop and exit here with code 1? If so, you can use "RunE" instead of "Run" (line 19) and return an error.

cli/commands/generate.go
```diff
@@ -24,17 +25,25 @@ var generateCmd = &cobra.Command{
 	Long: `In contrast to wunderctl up, it only generates all files in ./generated but doesn't start WunderGraph or the hooks.
 Use this command if you only want to generate the configuration`,
 	RunE: func(cmd *cobra.Command, args []string) error {
+		entryPoints, err := files.GetWunderGraphEntryPoints(wundergraphDir, configEntryPointFilename, serverEntryPointFilename)
+		if err != nil {
+			log.Fatal(`could not find file or directory`,
```
jensneuse, line 30: same as comment on add.go

cli/commands/publish.go
```diff
@@ -22,6 +24,10 @@ The APIs to publish need to be generated into the .wundergraph/generated directo
 	Example: `wunderctl publish organization/api`,
 	Args:    cobra.MinimumNArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
+		if !files.DirectoryExists(wundergraphDir) {
+			log.Fatal(`could not find base directory`, abstractlogger.String("dir", wundergraphDir))
```
jensneuse, line 28: same as comment on add.go

cli/commands/remove.go
```diff
@@ -14,6 +16,10 @@ var removeCmd = &cobra.Command{
 	Example: `wunderctl remove spacex/spacex`,
 	Args:    cobra.MinimumNArgs(1),
 	Run: func(cmd *cobra.Command, dependencies []string) {
+		if !files.DirectoryExists(wundergraphDir) {
+			log.Fatal(`could not find base directory`, abstractlogger.String("dir", wundergraphDir))
```
jensneuse, line 20: same as comment on add.go

cli/commands/root.go
```diff
@@ -79,6 +79,10 @@ You can opt out of this by setting the following environment variable: WUNDERGRA
 		}
 	},
 	Run: func(cmd *cobra.Command, args []string) {
+		if !files.DirectoryExists(wundergraphDir) {
+			_, _ = red.Printf(`unable to find "%s" directory`, wundergraphDir)
```
jensneuse, line 83: same as comment on add.go

cli/commands/start.go
```diff
@@ -38,15 +37,24 @@ just running the engine as efficiently as possible without the dev overhead.
 If used without --exclude-server, make sure the server is available in this directory:
 {entrypoint}/bundle/server.js or override it with --server-entrypoint.`,
 	RunE: func(cmd *cobra.Command, args []string) error {
-		ctx, cancel := context.WithCancel(context.Background())
-		defer cancel()
+		entryPoints, err := files.GetWunderGraphEntryPoints(wundergraphDir, configEntryPointFilename, serverEntryPointFilename)
+		if err != nil {
+			log.Fatal(`could not find file or directory`,
```
jensneuse, line 42: this uses already "RunE", so an error can be returned

pkg/bundler/bundler.go
```diff
@@ -121,11 +124,12 @@ func (b *Bundler) BundleAndWatch(ctx context.Context) {
...
-		EntryPoints: []string{b.entryPoint},
-		Bundle:      true,
-		Incremental: true,
-		Platform:    api.PlatformNode,
+		Outfile:       b.outFile,
+		EntryPoints:   []string{b.entryPoint},
+		Bundle:        true,
+		Incremental:   true,
+		Platform:      api.PlatformNode,
+		AbsWorkingDir: b.absWorkingDir,
```
jensneuse, line 132: :thumbs_up:

pkg/files/files.go
```diff
@@ -14,34 +15,50 @@ func DirectoryExists(path string) bool {
...
+	files := &WunderGraphEntryPoints{
+		WunderGraphDirAbs: wundergraphAbsDir,
+	}
+
+	configEntryPoint := path.Join(wundergraphDir, configEntryPointFilename)
+
+	if FileExists(configEntryPoint) {
+		files.ConfigEntryPointAbs = configEntryPoint
+	} else {
+		return nil, fmt.Errorf(`code file "%s" not found`, configEntryPoint)
```
jensneuse, line 52: what happens if the generated dir does not yet exist? will it always fail? e.g. during "up"

pkg/scriptrunner/scriptrunner.go
```diff
@@ -130,12 +133,13 @@ func (b *ScriptRunner) Run(ctx context.Context) chan struct{} {
 // newCmd creates a new command to run the bundler script.
 // it returns a channel that is closed when the command is done.
 // this is necessary to make sure the IO was flushed after the script is stopped.
-func newCmd(executable string, scriptArgs []string, scriptEnv []string) (*gocmd.Cmd, chan struct{}) {
+func newCmd(executable string, cmdDir string, scriptArgs []string, scriptEnv []string) (*gocmd.Cmd, chan struct{}) {
```
jensneuse, line 136: >3 args indicates a smell

cli/commands/up.go
```diff
@@ -36,8 +33,14 @@ var upCmd = &cobra.Command{
 	Short: "Start the WunderGraph application in the current dir",
 	Long:  `Make sure wundergraph.config.json is present or set the flag accordingly`,
 	RunE: func(cmd *cobra.Command, args []string) error {
+		entryPoints, err := files.GetWunderGraphEntryPoints(wundergraphDir, configEntryPointFilename, serverEntryPointFilename)
+		if err != nil {
+			log.Fatal(`could not find file or directory`,
```
jensneuse, line 38: similar to add.go comment, with one addition: will this not fail if the generated files don't yet exist? e.g. wundergraph.config.js doesn't exist on the first run

jensneuse (changes requested):

cli/commands/add.go
StarpTech: Good point

pkg/files/files.go
StarpTech: We are only searching for the base dir not generated things.

cli/commands/up.go
StarpTech: No, in `GetWunderGraphEntryPoints` we are looking for 

- .wundergraph directory
- wundergraph.config.ts
- wundergraph.server.ts

But it will fail when `wundergraph.server.ts` don't exist. I need to change that.

StarpTech: Forget that place.

StarpTech: Done.

jensneuse (approved):

PR #85, opened 2022-07-18
feat: use web std headers implementation

StarpTech:
The current [headers-polyfill](https://github.com/mswjs/headers-polyfill) exposes more than the standard. I like the idea to follow the Web Standard IO API where it makes sense. The proposed lib is also used internally by Remixjs. 

#### Checklist

- [ ] run `make test`


StarpTech:
I'm going to open a new pr.

PR #109, opened 2022-08-02
feat: add grpc datasource support

spetrunin:
#### Checklist

- [ ] run `make test`


jensneuse:
@spetrunin can you please add intructions to default test app README on how to start the dev grpc service

jensneuse (changes requested):

pkg/datasources/grpc/planner.go
```diff
@@ -0,0 +1,112 @@
...
+		p.v.Walker.StopWithInternalErr(err)
+		return nil
+	}
+	return src
+}
+
+// connection - temporary until lazy connection added
+func (p *Planner) connection() *grpc.ClientConn {
+	conn, err := grpc.DialContext(p.ctx, p.config.Server.Target,
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
```
StarpTech, line 86: I think this must be configurable or secure by default.

StarpTech (changes requested):
Only small points except for the insecure grpc connection. Great overall PR :+1:

pkg/datasources/grpc/planner.go
StarpTech: Would also be fine to me to address that in a separate PR:

PR #121, opened 2022-08-09
DRAFT: extract hooks from Nextjs into new React package

Pagebakers:
#### Checklist

- [x] run `make test`


Pagebakers:
I removed all 'contextWrapper' related code, because it was hard to read and add a lot of unneeded complexity and size to the overall package. Instead I opted for an optional 'context' option parameter, so you can pass in a custom context. The templates use this to add the local context to the React hooks.

Right now all code is backwards compatible and there are no breaking changes for now. But I'm not 100% satisfied with the API yet, since we are introducing a new React package, this is also the best time to introduce any API changes from the original hooks.

Here is an example of a generated hook.
```ts
export const useHelloQuery = (args: QueryArgsWithInput<HelloInput>) =>
	useQuery<HelloInput, HelloResponseData, Role>(
		{
			operationName: "Hello",
		},
		args,
		{ context: wunderGraphContext, requiresAuthentication: false } // internal options, but could be consolidated with args.
	);

// the next package just re-exports this.

export const useQuery {
  HelloQuery: useHelloQuery
}
```

My personal preference is to change the api as followed. Separate the input from the args/options and use a string for operationName instead of another object. Also move `result` one level up.

```
const { data, status, refetch } = useQuery('Hello', { name: 'Eelco' }, options)

// data.hello.name instead of result.data.hello.name

const { data, mutate } = useMutation('SetName', options)
const { data } = useSubscription('LiveWeather', { forCity: 'Berlin' }, options)
```

- More in line with other libraries like RTK query and React Query
- Cleaner code and smaller size and easier to write (see #80)

We should also consider adding some more request state variables, a couple of suggestions.

```
const { isFetching } = useQuery('Hello', { name: 'Eelco' }, options)

isFetching when data is being fetched. 

const { isDone } = useSubscription('LiveWeather', { forCity: 'Berlin' }, { isLiveQuery: true })

isDone = true when the subscription has ended.
```

## SSR

The new React hooks have SSR functionality build in. But currently the package has no helpers to make SSR work (eg with react-ssr-prepass in Next.js). So it's not a plug and play in Vite for example. I think it's out of the scope for this issue, but one question:

- Should the React hooks have SSR code by default? or should we have non SSR hooks as well.

StarpTech:
> const { data, status, refetch } = useHelloQuery({ name: 'Eelco' }, options)

I like this! For state variables a great place for inspiration is https://swr.vercel.app/docs/options#return-values

> Should the React hooks have SSR code by default? or should we have non SSR hooks as well.

I'd prefer that they work for SSR by default.

With this preference, can we still ship it without breaking customers?

Pagebakers:
Cool, I go ahead with these changes if @jensneuse agrees too.

SSR will still work as before with Next.js, no changes needed.

Pagebakers:
Closing this for now as will we implement the learnings into SWR and React Query hooks

PR #122, opened 2022-08-11
feat: add node managed mode

spetrunin:
#### Checklist

- [ ] run `make test`

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -182,7 +182,7 @@ func (n *Node) StartBlocking(opts ...Option) error {
 
 		if options.staticConfig != nil {
 			n.log.Info("Api config: loading cached config")
-			n.configCh <- *options.staticConfig
+			go n.startServer(*options.staticConfig)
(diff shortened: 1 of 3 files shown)


pkg/node/node.go
```diff
@@ -164,18 +174,28 @@ func (n *Node) StartBlocking(opts ...Option) error {
...
-	if options.staticConfig != nil {
+	switch {
+	case options.enableManagedMode:
+		n.log.Info("Api config: managed mode")
+		go n.reconfigureOnConfigUpdate()
+		go n.netPollConfig()
+
+		if options.staticConfig != nil {
+			n.log.Info("Api config: loading cached config")
+			n.configCh <- *options.staticConfig
```
jensneuse, line 185: can we do this synchronously instead of async through channel on first load?

jensneuse (changes requested):

jensneuse:
canceled, might pick up later

PR #127, opened 2022-08-16
feat: implement webhook reverse proxy

jensneuse:


pkg/apihandler/apihandler.go
```diff
@@ -99,6 +101,7 @@ type BuilderConfig struct {
 	EnableIntrospection        bool
 	GitHubAuthDemoClientID     string
 	GitHubAuthDemoClientSecret string
+	HookServerURL              string
```
jensneuse, line 104: @spetrunin I've had to add this already to be able to "inject" a test hook server

PR #139, opened 2022-08-24
fix: s3 upload client

YuriBuerov:
Bug description
------
 - It's impossible to upload small files, like the ssh public key in my case. 
 - It's impossible to upload files when the boundary prefix is followed by a double dash, space, tab, cr, nl, or end of input. 
 
In both these cases, EOF error returned within the response.  This is not correct, EOF should not be treated as an error in this case. 

There are a few issues with the existing implementation: 

```golang
contentTypeBuff := make([]byte, 512) // too much
_, err := part.Read(contentTypeBuff)
if err != nil {
        // io.EOF error should be handled  
	return nil, err
}
filetype := http.DetectContentType(contentTypeBuff) 
recycled := io.MultiReader(bytes.NewReader(contentTypeBuff), part)
```

```multipart.Part``` ```Read```  reads the body of a part, after its headers (!), and could return EOF.  Which is not handled, but if it would be handled like: ```if err != io.EOF { return nil, err }```  uploaded data could be corrupted (first case in bug description) because of preallocated bytes in ```contentTypeBuff```,  and it will not be possible to detect the correct content type. 

Solution
------

 - ```multipart.Part``` comes with prepopulated headers, in most cases, it should be enough to get content type.
 -  if we couldn't find the content type based on the file header, try based on the extension
 -  additionally, I removed the extra step with temp file creation, which makes upload a bit faster. 

TODO:
 - tests?
  

#### Checklist

- [x] run `make test`

### DIFF
--- a/.github/workflows/engine-ci.yaml
+++ b/.github/workflows/engine-ci.yaml
@@ -36,6 +36,15 @@ env:
 jobs:
   build_test:
     runs-on: ubuntu-latest
+    services:
+      minio:
+        image: minio/minio:edge-cicd
+        ports:
+          - 9000:9000
+        env:
+          MINIO_ROOT_USER: minio
+          MINIO_ROOT_PASSWORD: minio123
+        options: --name=minio --health-cmd "curl http://localhost:9000/minio/health/live"
     steps:
       - uses: actions/checkout@v3
         with:
@@ -55,6 +64,14 @@ jobs:
           version: '3.x'
           repo-token: ${{ secrets.GITHUB_TOKEN }}
 
+      - run: wget https://dl.min.io/client/mc/release/linux-amd64/mc
+      - run: chmod +x ./mc
+      - run: ./mc alias set minio http://127.0.0.1:9000 minio minio123
+      - run: ./mc admin user add minio test 12345678
+      - run: ./mc admin policy set minio readwrite user=test
+      - run: ./mc admin user info minio test
+      - run: ./mc mb --ignore-existing minio/uploads
+      - run: ./mc policy set public minio/uploads
       - run: pnpm install --ignore-scripts
       - run: make engine-dev
       - run: make test-go
--- a/.github/workflows/engine-ci.yaml
+++ b/.github/workflows/engine-ci.yaml
@@ -64,14 +64,7 @@ jobs:
           version: '3.x'
           repo-token: ${{ secrets.GITHUB_TOKEN }}
 
-      - run: wget https://dl.min.io/client/mc/release/linux-amd64/mc
-      - run: chmod +x ./mc
-      - run: ./mc alias set minio http://127.0.0.1:9000 minio minio123
-      - run: ./mc admin user add minio test 12345678
-      - run: ./mc admin policy set minio readwrite user=test
-      - run: ./mc admin user info minio test
-      - run: ./mc mb --ignore-existing minio/uploads
-      - run: ./mc policy set public minio/uploads
+      - run: make bootstrap-minio
       - run: pnpm install --ignore-scripts
       - run: make engine-dev
       - run: make test-go
--- a/pkg/s3uploadclient/s3uploadclient.go
+++ b/pkg/s3uploadclient/s3uploadclient.go
@@ -1,13 +1,13 @@
 package s3uploadclient
 
 import (
-	"bytes"
 	"encoding/hex"
 	"encoding/json"
+	"fmt"
 	"io"
-	"mime"
 	"mime/multipart"
 	"net/http"
+	"os"
 	"path/filepath"
 	"strconv"
 	"time"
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: pkg/s3uploadclient/s3uploadclient.go [hunk 1]
```diff
@@ -81,47 +80,34 @@ func (s *S3UploadClient) createBucket() error {
...
-	recycled := io.MultiReader(bytes.NewReader(contentTypeBuff), part)
+	contentType := "application/octet-stream"
 
+	// unless we can find type based on file header, try based on extension
 	extension := filepath.Ext(part.FileName())
-	// creates a temporary unique file in the temp folder of the OS
-	dst, err := ioutil.TempFile("", fmt.Sprintf("wundergraph-upload.*.%s", extension))
-	if err != nil {
-		return nil, err
+	if t := mime.TypeByExtension(extension); t != "" {
```
StarpTech, line 87: We can get rid of this when we use `FPutObject` from the minio client library. See https://github.com/minio/minio-go/blob/174b4c070788c7c1eed4ba4bf6d5497198aaf2e6/api-put-object-file-context.go#L58-L62

StarpTech (changes requested):
Hi @YuriBuerov thank you for opening a PR :100: I have left some comments.

pkg/s3uploadclient/s3uploadclient.go
```diff
@@ -81,47 +80,34 @@ func (s *S3UploadClient) createBucket() error {
...
-	_, err = io.Copy(dst, recycled)
-	if err != nil {
-		return nil, err
+	// find type based on file header, populated when *multipart.Part created
+	if t := part.Header.Get("Content-Type"); t != "" {
+		contentType = t
 	}
 
-	_, err = dst.Seek(0, io.SeekStart)
+	body, err := io.ReadAll(part)
```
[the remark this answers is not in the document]
YuriBuerov: @StarpTech Thanks for your comments. I do agree that ```ReadAll``` is not efficient, in the case of large files, supposing that ```MaxUploadSize``` will become configurable someday.  Good point, I missed this. At the same time temporary file creation and multiple ```io.Copy()``` calls could be quite costly as well. Copy does not take out all the data at one time (which is a big advantage compared to ```ReaadAll```), but keeps reading and writing. 

What if we pass ```part *multipart.Ppart``` directly to minio client?  Streaming approach, without extra steps, fast, and less code. We will not be able to precalculate size, but minio can do this for us if we pass -1 as the size parameter. 

The only drawback is that in this case, we lose de-dup feature ```filename := hex.EncodeToString(hasher.Sum(nil)) + extension```, which prevents uploading the same file twice and could save some storage space, but it still updates the existing file even if the content was not changed. 

What do you think? The only alternative I see is using ```FPutObject``` as you've suggested and keeping temp file creation.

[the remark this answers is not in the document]
StarpTech: I'd always favor reliability over performance when performance is still good. Passing `-1` will according to https://github.com/minio/minio-go/blob/master/docs/API.md#putobjectctx-contextcontext-bucketname-objectname-string-reader-ioreader-objectsize-int64opts-putobjectoptions-info-uploadinfo-err-error consume a lot of memory. The de-dup feature is pretty important for space efficiency. I suggest `FPutObject`.

YuriBuerov:
@StarpTech I implemented tests, using docker. Would you like me to push tests as well? If so, I will have to make some changes. 

There are 3 options:
- run docker in tests using dockertest (may be slow on ci?)
- wrap minio client to be able to mock it
- pass transport within ```Options``` to ```NewS3UploadClient``` and rewrite it in tests

StarpTech:
Hi @YuriBuerov I like the idea of an simple integration test. What do you think about using minio image with Github CI service containers https://docs.github.com/en/actions/using-containerized-services/about-service-containers#creating-service-containers? Locally, we could boostrap minio with a simple script like https://github.com/wundergraph/wundergraph/blob/main/packages/nextjs-testapp/minio/setup.sh? I think this works well enough before we pull a library like dockertest into the codebase.

.github/workflows/engine-ci.yaml
```diff
@@ -55,6 +64,14 @@ jobs:
           version: '3.x'
           repo-token: ${{ secrets.GITHUB_TOKEN }}
 
+      - run: wget https://dl.min.io/client/mc/release/linux-amd64/mc
```
StarpTech, line 67: Please put it in a script so it can be reproduced locally.

StarpTech (changes requested):

pkg/s3uploadclient/s3uploadclient_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+	resp := make([]s3uploadclient.UploadedFile, 0)
+	err := json.NewDecoder(res.Body).Decode(&resp)
+	require.NoError(t, err)
+
+	require.NotEmpty(t, resp)
+	assert.Equal(t, "cec8f2d4d95a43d0.json", resp[0].Key)
+}
+
+func initClient() *s3uploadclient.S3UploadClient {
```
StarpTech, line 52: Could you put the test helper in a separate file? This makes reading the test cases easier. Ideally, you can put the testdata under fixtures directory.

StarpTech (changes requested):

pkg/s3uploadclient/s3uploadclient_test.go
YuriBuerov: @StarpTech Would it be better to put testdata.go under a ```testdata``` directory so the build will ignore it? Either put a json file under  ```fixtures``` directory and open this file in the test.

StarpTech: Sounds good! I like the `testdata` idea.

scripts/minio-setup.sh
```diff
@@ -0,0 +1,21 @@
...
+	wget https://dl.min.io/client/mc/release/linux-amd64/mc -P scripts/
+elif [[ "$OSTYPE" == "darwin"* ]]; then
+	if [[ $(uname -p) == "arm" ]]; then
+		wget https://dl.min.io/client/mc/release/darwin-arm64/mc -P scripts/
+  elif [[ $(uname -p) == "i386" ]]; then  
+  	wget https://dl.min.io/client/mc/release/darwin-amd64/mc -P scripts/
+  fi
+fi
+
+chmod +x ./scripts/mc
```
StarpTech, line 13: I think you need to add "mc" binary to .gitignore?

scripts/minio-setup.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+if [[ "$OSTYPE" == "linux-gnu"* ]]; then
```
StarpTech, line 3: Can we check if the binary exists to avoid redownloading it?

StarpTech (changes requested):

scripts/minio-setup.sh
YuriBuerov: The script deletes it at the end

scripts/minio-setup.sh
YuriBuerov: Do you mean redownloading if we rerun ci?

StarpTech: I mean if you have already installed it locally. I think for CI this is negotiable.

scripts/minio-setup.sh
StarpTech: I'd still add it because when one command fails the binary is not deleted.

scripts/minio-setup.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
```
StarpTech, line 2: Always add `set -e`

StarpTech (approved):
LGTM

StarpTech:
@YuriBuerov ready to merge?

YuriBuerov:
> @YuriBuerov ready to merge?

@StarpTech Ready

PR #143, opened 2022-08-28
fix: unify .wundergraph directory resolution

StarpTech:
#### Checklist

- [ ] run `make test`

### DIFF
--- a/cli/commands/generate.go
+++ b/cli/commands/generate.go
@@ -29,7 +29,7 @@ Use this command if you only want to generate the configuration`,
 	RunE: func(cmd *cobra.Command, args []string) error {
 		wgDir, err := files.FindWunderGraphDir(wundergraphDir)
 		if err != nil {
-			return fmt.Errorf("unable to find .wundergraph dir: %w", err)
+			return fmt.Errorf(files.WunderGraphDirNotFoundErrorMsg, err)
 		}
 
 		// only validate if the file exists
--- a/cli/commands/generate.go
+++ b/cli/commands/generate.go
@@ -38,11 +38,8 @@ Use this command if you only want to generate the configuration`,
 			return fmt.Errorf(files.CodeFileNotFoundErrorMsg, configEntryPointFilename)
 		}
 
-		// optional
+		// optional, no error check
 		codeServerFilePath, _ := files.CodeFilePath(wgDir, serverEntryPointFilename)
-		if err != nil {
-			return fmt.Errorf(files.CodeFileNotFoundErrorMsg, serverEntryPointFilename)
-		}
 
 		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
 		defer cancel()
--- a/cli/commands/generate.go
+++ b/cli/commands/generate.go
@@ -27,21 +27,21 @@ var generateCmd = &cobra.Command{
 	Long: `In contrast to wunderctl up, it only generates all files in ./generated but doesn't start WunderGraph or the hooks.
 Use this command if you only want to generate the configuration`,
 	RunE: func(cmd *cobra.Command, args []string) error {
-		wgDir, err := files.GetWunderGraphDir(wundergraphDir)
+		wgDir, err := files.FindWunderGraphDir(wundergraphDir)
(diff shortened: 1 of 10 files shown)


cli/commands/generate.go
```diff
@@ -27,9 +27,21 @@ var generateCmd = &cobra.Command{
...
+		}
+
+		// only validate if the file exists
+		_, err = files.GetWunderGraphConfigFilePath(wgDir, configEntryPointFilename)
+		if err != nil {
+			return err
+		}
+
+		// optional
+		codeServerFilePath, _ := files.GetWunderGraphServerFilePath(wgDir, configEntryPointFilename)
```
spetrunin, line 42: I believe it should use `serverEntryPointFilename`

cli/commands/generate.go
```diff
@@ -138,8 +150,8 @@ Use this command if you only want to generate the configuration`,
 
 		configBundler := bundler.NewBundler(bundler.Config{
 			Name:          "config-bundler",
-			AbsWorkingDir: entryPoints.WunderGraphDirAbs,
-			EntryPoints:   []string{configEntryPointFilename},
+			AbsWorkingDir: wgDir,
+			EntryPoints:   []string{serverEntryPointFilename},
```
spetrunin, line 154: is it correct that here we use server entryPoint? previously it was a config

pkg/files/files.go
```diff
@@ -24,56 +24,58 @@ func FileExists(filePath string) bool {
...
-	} else {
-		files.WunderGraphDirAbs = wundergraphAbsDir
+		return dir, nil
 	}
 
-	configEntryPoint := path.Join(files.WunderGraphDirAbs, configEntryPointFilename)
+	return absWgDir, nil
+}
+
+// GetWunderGraphConfigFilePath returns the absolute path to the wundergraph.config.ts file.
```
spetrunin, line 61: These 2 methods look the same, probably we could use only 1 unified
cause they produce errors which differed only in filename from param

cli/commands/up.go
```diff
@@ -35,9 +35,21 @@ var upCmd = &cobra.Command{
...
+		}
+
+		// only validate if the file exists
+		_, err = files.GetWunderGraphConfigFilePath(wgDir, configEntryPointFilename)
+		if err != nil {
+			return err
+		}
+
+		// optional
+		codeServerFilePath, _ := files.GetWunderGraphServerFilePath(wgDir, configEntryPointFilename)
```
spetrunin, line 50: should be server also?

spetrunin (changes requested):

cli/commands/generate.go
StarpTech: I got rid of the `EntryPoint` naming schema.

cli/commands/generate.go
StarpTech: Good catch!

pkg/files/files.go
StarpTech: Done.

cli/commands/generate.go
```diff
@@ -27,9 +27,21 @@ var generateCmd = &cobra.Command{
...
+
+		// only validate if the file exists
+		_, err = files.GetValidFilePath(wgDir, configEntryPointFilename)
+		if err != nil {
+			return fmt.Errorf(`code file "%s" not found`, configEntryPointFilename)
+		}
+
+		// optional
+		codeServerFilePath, _ := files.GetValidFilePath(wgDir, serverEntryPointFilename)
+		if err != nil {
```
spetrunin, line 43: which error are we checking here?

cli/commands/generate.go
```diff
@@ -27,9 +27,21 @@ var generateCmd = &cobra.Command{
...
+			return fmt.Errorf(`code file "%s" not found`, serverEntryPointFilename)
```
spetrunin, line 44: do we need to repeat the error shape from the helper?

StarpTech: What do you mean?

spetrunin (approved):

PR #145, opened 2022-08-29
feat: use web std header implementation

StarpTech:
#### Checklist

- [ ] run `make test`

### DIFF
--- a/packages/default-testapp/.wundergraph/webhooks/github.ts
+++ b/packages/default-testapp/.wundergraph/webhooks/github.ts
@@ -9,8 +9,13 @@ const webhook: Webhook<InternalClient, { a: number }, { hello: string; a: number
 		// and don't bundle them into the webhook.
 		buildSchema(`scalar DateTime`);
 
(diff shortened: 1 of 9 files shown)


packages/default-testapp/.wundergraph/webhooks/github.ts
```diff
@@ -9,8 +9,13 @@ const webhook: Webhook<InternalClient, { a: number }, { hello: string; a: number
 		// and don't bundle them into the webhook.
 		buildSchema(`scalar DateTime`);
 
+		console.log(event.headers);
```
spetrunin, line 12: should we prefix here with 'Webhooks.Github'?

StarpTech: yes

StarpTech: My mistake, let's log with the webhook longer which is already scoped.

spetrunin (approved):

PR #149, opened 2022-08-30
fix(types): make webhooks optional

StarpTech:
#### Checklist

- [ ] run `make test`

### DIFF
--- a/pkg/s3uploadclient/s3uploadclient_test.go
+++ b/pkg/s3uploadclient/s3uploadclient_test.go
@@ -2,6 +2,7 @@ package s3uploadclient_test
 
 import (
 	"encoding/json"
+	"os"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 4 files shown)


jensneuse:
should the title be "make webhooks tests optional"?

StarpTech:
> should the title be "make webhooks tests optional"?

No, this PR makes webhooks in ConfigureWunderGraph types optional. I could not run the tests locally because integration tests are run by default. That behavior is also addressed in this PR.

jensneuse (approved):

jensneuse (approved):

pkg/s3uploadclient/s3uploadclient_test.go
```diff
@@ -10,6 +11,10 @@ import (
 )
 
 func TestS3UploadClient_UploadFile(t *testing.T) {
+	if os.Getenv("INT") != "true" {
```
spetrunin, line 14: does it make sense to use [short](https://pkg.go.dev/testing#hdr-Skipping)?



example:
https://github.com/wundergraph/graphql-go-tools/pull/382/commits/2b606d9bfb33992ad28c54f57b6ce6df1861d7d2

StarpTech: Good point. Would you like to create a PR? Otherwise, I'll tackle this next time.

PR #153, opened 2022-08-31
feat: add dev mode so that cookie secrets stay static

jensneuse:
This allows you to run `wunderctl up` and never get logged out of the application during development.

### DIFF
--- a/pkg/apihandler/apihandler.go
+++ b/pkg/apihandler/apihandler.go
@@ -87,6 +87,7 @@ type Builder struct {
 	forceHttpsRedirects bool
 	enableDebugMode     bool
 	enableIntrospection bool
+	devMode             bool
 
 	renameTypeNames []resolve.RenameTypeName
 
@@ -102,6 +103,7 @@ type BuilderConfig struct {
 	GitHubAuthDemoClientID     string
 	GitHubAuthDemoClientSecret string
 	HookServerURL              string
+	DevMode                    bool
 }
 
 func NewBuilder(pool *pool.Pool,
@@ -122,6 +124,7 @@ func NewBuilder(pool *pool.Pool,
 		enableIntrospection:        config.EnableIntrospection,
 		githubAuthDemoClientID:     config.GitHubAuthDemoClientID,
 		githubAuthDemoClientSecret: config.GitHubAuthDemoClientSecret,
+		devMode:                    config.DevMode,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


Pagebakers:
CSRF tokens can still be tested in devmode? These are now static too?

spetrunin (approved):

jensneuse:
> CSRF tokens can still be tested in devmode? These are now static too?

Yes, we're just not invalidating CSRF cookies because the secret is not random anymore.
So a CSRF session will be valid across restarts of the wundernode/config changes.

pkg/apihandler/apihandler.go
```diff
@@ -1928,6 +1937,14 @@ func (r *Builder) registerAuth(pathPrefix string, insecureCookies bool) {
 	r.registerCookieAuthHandlers(cookieBasedAuth, cookie, pathPrefix)
 }
 
+func (r *Builder) staticInsecureSecret(length int) []byte {
```
StarpTech, line 1940: This is odd 😄  Why not passing just working with a fixed secret?

pkg/apihandler/apihandler.go
```diff
@@ -1867,22 +1870,28 @@ func (r *Builder) registerAuth(pathPrefix string, insecureCookies bool) {
...
 
-	if b := loadvariable.String(r.api.AuthenticationConfig.CookieBased.BlockKey); b != "" {
-		blockKey = []byte(b)
-	} else {
-		blockKey = r.generateSecret(32)
-	}
+		if b := loadvariable.String(r.api.AuthenticationConfig.CookieBased.BlockKey); b != "" {
+			blockKey = []byte(b)
+		} else {
+			blockKey = r.generateSecret(32)
```
StarpTech, line 1887: I don't understand the fact that we need to regenerate the secret on every restart. How does it work in production? Does the user has login after every deployment?

StarpTech (changes requested):

pkg/apihandler/apihandler.go
StarpTech: The secret should be mandatory in production.

jensneuse: I disagree. Here's my reasoning: By default, it should always work. That's most important for a great DX.
So on every deployment, in theory, you get a new secret, which invalidates existing user sessions.
If a user wants to avoid this, they can define the secrets in their config or use environment variables.
In our cloud, we should manage and rotate secrets automatically.

By making this mandatory, you a) make deployments harder and b) it's irrelevant for our cloud.
We need a deployment guide anyways, but by default, deployments should work without any additional config.

pkg/apihandler/apihandler.go
jensneuse: I don't manually count the number of characters of a string, lol.

pkg/apihandler/apihandler.go
StarpTech: Correct me, If I'm wrong, but in order to enable auth you need to configure it in wundergraph.config.ts. My idea is that you "must" pass a secret to fulfill the type. This avoids pitfalls in production. In development you could just pass any random string.

pkg/apihandler/apihandler.go
StarpTech: Why do we need a method to generate a random fixed string if we can work with any here?

jensneuse: it must be a specific length

pkg/apihandler/apihandler.go
jensneuse: In production (cloud) we manage this.

pkg/apihandler/apihandler.go
StarpTech: Then you could just define const string?

PR #154, opened 2022-08-31
chore: Update CODEOWNERS

jensneuse:


.github/CODEOWNERS
```diff
@@ -2,4 +2,4 @@
 
 # These owners will be the default owners for everything in
 # the repo. Unless a later match takes precedence,
-*                                @starptech @jensneuse
\ No newline at end of file
+*                                @starptech @jensneuse @spetrunin @YuriBuerov
```
StarpTech, line 5: I'd like to keep it as it is because you need some profound knowledge about the code base to be a code owner.

PR #156, opened 2022-08-31
fix(config): set subscription url from config

manini-i3hub:
this is a fix for issue in subscription url config that is not reflected in the server


StarpTech:
Hi, @manini-i3hub thank you for the PR. I will look into it soon.

jensneuse:
Merged with #162

PR #158, opened 2022-09-01
feat: refactor introspection runner

StarpTech:
This PR will fix the situation that the user will always see up-to-date introspection results when they restart wunderctl. But in case of an API outage, we will fall back to an old cache result if possible. I also ensured that the poller could never throw and exit due to an introspection error.

#### Checklist

- [ ] run `make test`


cli/commands/up.go
```diff
@@ -106,6 +106,12 @@ var upCmd = &cobra.Command{
 			AbsWorkingDir: wgDir,
 			ScriptArgs:    []string{configOutFile},
 			Logger:        log,
+			FirstRunEnv: []string{
+				// when the user runs `wunderctl up` for the first time, we revalidate the cache
+				// so the user can be sure that the introspection is up to date. In case of an API is not available
+				// we will fallback to the cached introspection (when available)
+				"WG_ENABLE_INTROSPECTION_CACHE=false",
```
jensneuse, line 113: smart

packages/default-testapp/.wundergraph/wundergraph.config.ts
```diff
@@ -26,6 +26,9 @@ const countries = introspect.graphql({
 const weather = introspect.graphql({
 	apiNamespace: 'weather',
 	url: 'https://graphql-weather-api.herokuapp.com/',
+	introspection: {
+		pollingIntervalSeconds: 15,
```
jensneuse, line 30: for testing, I'd suggest 5s

pkg/scriptrunner/scriptrunner.go
```diff
@@ -14,6 +14,7 @@ type Config struct {
 	Executable    string
 	ScriptArgs    []string
 	ScriptEnv     []string
+	FirstRunEnv   []string
```
jensneuse, line 17: Can you add comments to distinguish the different behaviours of FirstRunEnv and ScriptEnv?
When to use what? Why does it exist?

jensneuse (approved):
small minor requests
one thing i'd like to add is that we store the last time we've updated the cache for each cache item. This way, we can actually print out this information in the debug log, so you see exactly when this API cache item was updates last time for each introspection.

cli/commands/up.go
StarpTech: Thanks

StarpTech:
> small minor requests
one thing i'd like to add is that we store the last time we've updated the cache for each cache item. This way, we can actually print out this information in the debug log, so you see exactly when this API cache item was updates last time for each introspection.

I'd love to do this in a separate PR. COuld you create an issue?

PR #160, opened 2022-09-02
feat(node): implement api config validation for cookie-based auth

jensneuse:
This PR adds API config validation which detects wrong config for cookie-based auth configs.


packages/sdk/CHANGELOG.md
```diff
@@ -7,24 +7,24 @@ See [Conventional Commits](https://conventionalcommits.org) for commit guideline
 
 ### Features
 
-* implement userId for fromClaim directive ([#152](https://github.com/wundergraph/wundergraph/issues/152)) ([51df6e5](https://github.com/wundergraph/wundergraph/commit/51df6e50244bee9f5f8d579ff6f604e1a1c853d9)) (@jensneuse)
```
StarpTech, line 10: Why is it changed? This is auto generated.

pkg/node/node.go
```diff
@@ -308,6 +310,34 @@ func (n *Node) startServer(nodeConfig wgpb.WunderNodeConfig) {
...
+			// in dev mode, ignore missing cookie secrets
+			// we automatically override them
+			IgnoreMissingCookieAuthenticationSecrets: n.options.devMode,
+		})
+
+		if !valid {
+			n.log.Error("API config invalid",
+				abstractlogger.Strings("errors", messages),
+			)
+			os.Exit(1)
```
StarpTech, line 323: or use `log.fatal`? I think this ensures that the log buffer is printed before the process dies.

pkg/node/node.go
```diff
@@ -308,6 +310,34 @@ func (n *Node) startServer(nodeConfig wgpb.WunderNodeConfig) {
...
+			return
+		}
+
+		n.setApiConfigDefaults(api)
+
+		// revalidate after setting defaults
+		// just to make sure that defaults are validated
+		valid, messages = validate.DefaultApiConfig(api, validate.ApiConfigValidationConfig{})
```
StarpTech, line 331: Why do we need to call the validator twice? This is confusing to me.

pkg/validate/apiconfig.go
```diff
@@ -0,0 +1,87 @@
...
+type ApiConfigValidationConfig struct {
+	IgnoreMissingCookieAuthenticationSecrets bool
+}
+
+type ApiConfigValidationRule interface {
+	Validate(a *ApiConfigValidator)
+}
+
+type ApiConfigValidator struct {
+	api      *wgpb.Api
```
StarpTech, line 36: Is it really necessary that the validator has to know the `*wgpb.Api` object? The validator only needs to know: What is a valid value for a cookie secret / csrf.

```go
err := Validate(a.api.AuthenticationConfig.CookieBased.CsrfSecret, Validate.CookieCSRF)
err := Validate(a.api.AuthenticationConfig.CookieBased.BlockKey, Validate.CookieSecret)
```

I think this is easier to write and maintain.

StarpTech (changes requested):

packages/sdk/CHANGELOG.md
jensneuse: I don't know

pkg/node/node.go
jensneuse: the comment says it

pkg/validate/apiconfig.go
jensneuse: subjective

jensneuse: I've investigated a few other solutions. The issue remains that somewhere needs to sit the mapping.
I'll abandon the idea of improving the code at this point.
If you'd like to refactor it, ok.

packages/sdk/CHANGELOG.md
jensneuse: fixed

pkg/validate/apiconfig.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/wundergraph/wundergraph/pkg/apiconfig"
+	"github.com/wundergraph/wundergraph/pkg/loadvariable"
+	"github.com/wundergraph/wundergraph/types/go/wgpb"
+)
+
+// ApiConfig validates the Api
+func ApiConfig(api *wgpb.Api) (valid bool, messages []string) {
+
+	if !apiconfig.HasCookieAuthEnabled(api) {
+		// skip following tests if cookie auth is not enabled
```
StarpTech, line 13: If you put this here, we can only validate `AuthenticationConfig` settings, but `ApiConfig` is a place to validate the whole config.

jensneuse: For each rule, we should call one function which can return early. Currently, with just one rule, it's fine.

StarpTech: ok

PR #167, opened 2022-09-05
feat: ensure that user is always set in auth hooks, disable introspection cache in `wunderctl generate`

StarpTech:
#### Checklist

- [x] run `make test`

- [x] This PR ensures that `user` is always set otherwise, we will not request the authentication hooks. This also includes improvements on the typescript types.
- [x]  Ensure that default-testapp can be generated without type issues.
- [X] It also includes a fix for the introspection cache. We accidentally use the cache in `wunderctl generate`.


packages/sdk/src/middleware/types.ts
```diff
@@ -34,7 +34,7 @@ export interface BaseRequestContext<User = any, IC = InternalClient> {
 	internalClient: IC;
 }
 
-export interface AuthRequestContext<User = any> {
+export interface AuthenticationRequestContext<User = any> {
```
jensneuse, line 37: User = any or User = User?

StarpTech: As written `User = any` because we don't have `User` generated type in the SDK.

PR #170, opened 2022-09-07
feat: add golang client

jensneuse:


packages/default-testapp/.wundergraph/wundergraph.server.ts
```diff
@@ -47,18 +47,17 @@ export default configureWunderGraphServer<HooksConfig, InternalClient, WebhooksC
 				onOriginRequest: {
 					enableForAllOperations: true,
 					hook: async ({ request }) => {
-						request.headers.set('X-Wundergraph-Test', 'test');
 						console.log('onOriginRequest', request.headers);
 						return request;
 					},
 				},
-				onOriginResponse: {
+				/*onOriginResponse: {
```
StarpTech, line 54: Please enable it before merging.

packages/default-testapp/.wundergraph/wundergraph.server.ts
```diff
@@ -47,18 +47,17 @@ export default configureWunderGraphServer<HooksConfig, InternalClient, WebhooksC
 				onOriginRequest: {
 					enableForAllOperations: true,
 					hook: async ({ request }) => {
-						request.headers.set('X-Wundergraph-Test', 'test');
```
StarpTech, line 50: Please enable it before merging.

packages/default-testapp/golang/client/client.go
```diff
@@ -0,0 +1,95 @@
...
+
+func (c *Client) Subscriptions() *Subscriptions {
+	return &Subscriptions{c}
+}
+
+func (c *Client) LiveQueries() *LiveQueries {
+	return &LiveQueries{c}
+}
+
+type Queries struct {
```
StarpTech, line 36: Should we generate an interface for easier mocking?

packages/golang-client/src/index.ts
```diff
@@ -0,0 +1,288 @@
...
+		// check if gofmt is installed
+		const formatter = execa.sync('gofmt', {
+			input: code,
+			encoding: 'utf8',
+		});
+		if (formatter.exitCode === 0) {
+			return formatter.stdout;
+		}
+	} catch (e: any) {
+		console.error(e);
```
StarpTech, line 34: This can be quite annoying and confusing because most people arent go users. I would make it silent.

packages/golang-client/src/index.ts
```diff
@@ -0,0 +1,288 @@
...
+
+export interface GolangClientTemplateConfig {
+	packageName: string;
+}
+
+const defaultTemplateConfig: GolangClientTemplateConfig = {
+	packageName: 'client',
+};
+
+const gofmt = (code: string) => {
```
StarpTech, line 23: Nice.

packages/golang-client/src/index.ts
```diff
@@ -0,0 +1,288 @@
...
+	try {
+		// check if gofmt is installed
+		const formatter = execa.sync('gofmt', {
```
StarpTech, line 26: Can we make it part of the SDK? it would be great to have it consistent. This also validates the AST right?

packages/golang-client/src/index.ts
```diff
@@ -0,0 +1,288 @@
...
+				path: 'models.go',
+				content: gofmt(content),
+				header: golangHeader(this.config.packageName),
+			},
+		]);
+	}
+}
+
+const golangModelsBase = `
+type GraphQLError struct {
```
StarpTech, line 166: Do we really need to put this here? Can't we import it as a static go file?

packages/sdk/src/index.ts
```diff
@@ -76,6 +76,8 @@ export { configureWunderGraphServer } from './middleware/server';
 export { default as cors } from './cors';
 export { authProviders } from './configure/authentication';
 
+export { visitJSONSchema } from './codegen/index';
```
StarpTech, line 79: I don't like the fact that we expose internals to the user. I'd love to have a separate export only intended for plugin development.

StarpTech (changes requested):

packages/golang-client/src/index.ts
jensneuse: lol

packages/sdk/src/index.ts
jensneuse: can you contribute a better solution?

packages/golang-client/src/index.ts
jensneuse: It's easier to maintain this way and I'm not expecting it to change.

packages/golang-client/src/index.ts
jensneuse: Do you think we're going to use it anywhere else?

packages/default-testapp/golang/client/client.go
jensneuse: Isn't this a concern of the user?

packages/golang-client/src/index.ts
StarpTech: Probably not, but also in the SDK we follow the approach to putting static code into code files to benefit from all language tools/validation etc...

packages/golang-client/src/index.ts
StarpTech: rofl

packages/golang-client/src/index.ts
jensneuse: What exactly will be annoying? If you're not a go user, you don't add the go template to your code.
I can make it silent, but then we don't see why formatting failed. wdyt?

StarpTech: You're right. Then I would throw a more meaningful error like: "Formatting failed. Did you install gofmt?"

StarpTech: But it feels weird that the user has to install tooling to format code that he will never touch :smile:

jensneuse: Assuming you use go, because you're generating go code, you should have go installed, no? :D

packages/default-testapp/golang/client/client.go
StarpTech: No, otherwise the user has to update its interface every time a new operation is added.

packages/sdk/src/index.ts
StarpTech: Yes

packages/golang-client/src/index.ts
jensneuse: I've made it silent

StarpTech (approved):
LGTM

PR #178, opened 2022-09-12
feat: oidc,  allow passing additional query parameters to the IDP

YuriBuerov:
#### Checklist

- [x] run `make test`

Allow passing additional query parameters to the IDP. e.g. kc_idp_hint query


docs-website/src/pages/docs/examples/keycloak-openid-connect-authentication.md
```diff
@@ -0,0 +1,60 @@
...
+```shell
+npm install && npm start
+```
+
+### 3. Use the application
+
+On the NextJS frontend, click the "Login" button.
+Once the login is complete, the Frontend will automatically fetch the data and inject the bearer token into the origin request.
+
+Keycloak console: `http://localhost:8080/` user name: admin, password: admin
```
StarpTech, line 60: Move that into a list for better readability.

pkg/apihandler/apihandler.go
```diff
@@ -2036,11 +2037,21 @@ func (r *Builder) configureCookieProvider(router *mux.Router, provider *wgpb.Aut
 		if provider.OidcConfig == nil {
 			return
 		}
+
+		queryParams := make([]authentication.QueryParam, 0, len(provider.OidcConfig.QueryParams))
```
StarpTech, line 2041: If you create an array of proper size X, you should not append the values in the loop. Otherwise, you will allocate more than necessary. In that case, If you have ten params, your array will have a cap of 20.

StarpTech (changes requested):

pkg/apihandler/apihandler.go
StarpTech: https://go.dev/play/p/GtrpM-xzXbF

YuriBuerov: https://go.dev/play/p/cFaYJtVCGHk I set cap, not len

StarpTech: You're right :+1: I forgot about the semantics myself :smile: For code not in the hot path, I always prefer the shorter variant without preallocation.

YuriBuerov: Yeah, I'm mixing it from time to time [for example](https://github.com/wundergraph/wundergraph/pull/178/files#diff-b24fb1cbf3cd00bb7eed0771d1281eb4d4b0afd0e30a5f2fd62692482abb064fR158). But in terms of performance they are more or less equal, as far as I remember

types/protos/wundernode_config.proto
```diff
@@ -103,10 +103,16 @@ message GithubAuthProviderConfig {
 	ConfigurationVariable clientSecret = 2;
 }
 
+message OpenIDConnectQueryParam {
```
StarpTech, line 106: Don't use shortcuts. Write `OpenIDConnectQueryParameter` and `queryParameters`

StarpTech (changes requested):

StarpTech (approved):
LGTM

PR #179, opened 2022-09-13
feat: make typescript client typesafe

Pagebakers:
#### Checklist

- [x] run `make test`


Pagebakers:
The pure TypeScript client now accepts an `Operation` type that defines all operations with input and data types to make all operation methods typesafe.

Adding `TypeScriptClientTemplate` to a project will generate the `Operation` types and a `createClient` function.

Added an example with Vite and SWR in `package/vite-testapp`

packages/nextjs-testapp/pages/authentication.tsx
```diff
@@ -75,12 +76,12 @@ const ProtectedLiveWeather: FC<{ city: string }> = ({ city }) => {
 			{liveWeather.result.status === 'error' && <p>Error</p>}
 			{liveWeather.result.status === 'ok' && (
 				<div>
-					<h3>City: {liveWeather.result.data.getCityByName.name}</h3>
-					<p>{JSON.stringify(liveWeather.result.data.getCityByName.coord)}</p>
+					<h3>City: {liveWeather.result.data.getCityByName?.name}</h3>
+					<p>{JSON.stringify(liveWeather.result.data.getCityByName?.coord)}</p>
```
StarpTech, line 80: Why is it still required to access `getCityByName` when there is only one prop in response?

packages/vite-testapp/src/hooks/swr.ts
```diff
@@ -0,0 +1,97 @@
...
+>(
+	operationName: OperationName,
+	config?: SWRConfiguration
+) => {
+	const response = useSWR(operationName, null, config);
+
+	return {
+		...response,
+		async mutate(input: Input, options: MutatorOptions<Data>) {
+			const result = await client.mutate({
```
StarpTech, line 52: Here too: This doesn't look as designed.

StarpTech (changes requested):

packages/nextjs-testapp/pages/authentication.tsx
Pagebakers: I didn't change the nextjs interface in this PR.

Pagebakers: I didn't change the nextjs interface in this PR.

packages/vite-testapp/src/hooks/swr.ts
```diff
@@ -0,0 +1,97 @@
...
+	operationName: OperationName,
+	input: Input,
+	config: SWRConfiguration & { isLiveQuery: LiveQuery }
+): SWRResponse<Data, ResultError> => {
+	const [key] = useState({ operationName, input, isLiveQuery: config?.isLiveQuery });
+	const response = useSWR(key, null, config);
+	useEffect(() => {
+		const abort = new AbortController();
+		client.subscribe(
+			{ ...key, abortSignal: abort.signal } as unknown as SubscriptionProps<
```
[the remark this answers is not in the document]
Pagebakers: Yeah if you have ideas, since these hooks are only used for testing and no official release I didn't want to spent any more time on it for now.

packages/sdk/src/codegen/templates/typescript/react.ts
```diff
@@ -80,18 +81,18 @@ export class TypescriptReactHooks implements Template {
...
-	op.InternalVariablesSchema.properties !== undefined &&
-	Object.keys(op.InternalVariablesSchema.properties).length !== 0;
+// export const hasInternalInput = (op: GraphQLOperation): boolean =>
+// 	op.InternalVariablesSchema.properties !== undefined &&
+// 	Object.keys(op.InternalVariablesSchema.properties).length !== 0;
 
-export const hasInjectedInput = (op: GraphQLOperation): boolean =>
-	op.InjectedVariablesSchema.properties !== undefined &&
-	Object.keys(op.InjectedVariablesSchema.properties).length !== 0;
+// export const hasInjectedInput = (op: GraphQLOperation): boolean =>
```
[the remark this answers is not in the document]
Pagebakers: Oh forgot this!

packages/vite-testapp/src/hooks/swr.ts
```diff
@@ -0,0 +1,97 @@
...
+
+	return {
+		...response,
+		async mutate(input: Input, options: MutatorOptions<Data>) {
+			const result = await client.mutate({
+				operationName,
+				input,
+			} as unknown as MutationProps<OperationName, ClientMutationArgs<Input>>);
+
+			if (result.status === 'ok') {
```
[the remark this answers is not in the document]
Pagebakers: How should we handle that here? just throw the error?

packages/vite-testapp/src/hooks/swr.ts
Pagebakers: Doesn't look as as designed?

packages/vite-testapp/src/App.tsx
```diff
@@ -0,0 +1,49 @@
...
+const LiveWeather: React.FC<{ city: string }> = ({ city }) => {
+	const liveWeather = useSubscription('Weather', { forCity: city }, { isLiveQuery: true });
+
+	return (
+		<div>
+			{liveWeather.isValidating && <p>Loading...</p>}
+			{liveWeather.error && <p>Error</p>}
+			{liveWeather.data && (
+				<div>
+					<h3>City: {liveWeather.data.getCityByName?.name}</h3>
```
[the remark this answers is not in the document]
StarpTech: No I mean 

```ts
const { data } = useSubscription()
data.name
```

packages/vite-testapp/src/hooks/swr.ts
StarpTech: I mean this PR is about type safety, but this example has type conversions.

packages/vite-testapp/src/hooks/swr.ts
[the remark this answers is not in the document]
StarpTech: Same semantics as in SWR.
```ts
const { data, error } = useMutation()
```

> The returned value will be passed as data, and if it throws, it will be caught as error.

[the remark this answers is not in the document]
StarpTech: There should be no "partial" state

packages/vite-testapp/src/hooks/swr.ts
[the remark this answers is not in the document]
StarpTech: If this PR isn't about type safety then I don't understand the scope of this PR. The example represents the final SWR client, isn't it?

StarpTech (changes requested):

packages/vite-testapp/src/App.tsx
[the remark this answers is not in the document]
jensneuse: `getCityByName` is the root field of the Query, it cannot be omitted. An operation can have multiple root fields, therefore you must include them.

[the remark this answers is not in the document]
StarpTech: ah, you're right! I had the operation -> single root field image in my mind :face_with_spiral_eyes:

packages/vite-testapp/src/hooks/swr.ts
```diff
@@ -0,0 +1,149 @@
...
+	options: UseQueryOptions<LiveQuery> = {}
+): SWRResponse<Data, ResultError> => {
+	const { isLiveQuery, enabled = true, ...config } = options;
+	const key = { operationName, input };
+	const [_key] = unstable_serialize(key);
+	const response = useSWR(enabled ? key : null, !isLiveQuery ? queryFetcher : null, config);
+
+	useEffect(() => {
+		let unsubscribe: () => void;
+		if (isLiveQuery && enabled) {
```
jensneuse, line 60: Is it possible to move this outside of the effect and just use plain js?
Otherwise, all useQuery calls will have this effect, right?

jensneuse (approved):

packages/vite-testapp/src/hooks/swr.ts
Pagebakers: I thought about it, but otherwise we cannot cleanup the subscription if components get unmounted. Since it's a hook we cannot make the entire statement conditional.

the useEffect wont' do anything if it's not a liveQuery, so it does not add much overhead.

packages/sdk/src/client/index.ts
```diff
@@ -195,28 +232,30 @@ export class WunderGraphClient<Role> {
...
+						...this.extraHeaders,
+						'Content-Type': 'application/json',
+						'WG-SDK-Version': this.sdkVersion,
+					},
+					method: 'GET',
+					signal: subscription?.abortSignal,
+				});
 
 				if (!this.isOK(response) || response.body == null) {
 					cb(this.errorResult(response));
```
StarpTech, line 261: This currently swallows the original error. https://github.com/wundergraph/wundergraph/blob/main/packages/sdk/src/client/index.ts#L314

The should be able to access the original error.

StarpTech (changes requested):

packages/sdk/src/client/index.ts
StarpTech: This is a bit more complicated. `errorResult` is fine because it only executes for requests with statusCode != `200` so we are talking about gateway errors. The issue with this implementation is that the client maps statusCodes to gateway errors. We should just pass the gateway error to the client.

Pagebakers: This issue existed already in the client. I would say we create another issue to improve this, so we can ship the improved types.

StarpTech: ok

PR #183, opened 2022-09-15
feat: allow templates in baseURL for openapi

jensneuse:


packages/sdk/src/v2openapi/index.ts
```diff
@@ -674,10 +700,14 @@ class RESTApiBuilder {
...
 		}
 		const secure = this.spec.servers.find((server) => server.url.startsWith('https'));
-		if (secure) {
-			return secure.url;
-		}
-		return this.spec.servers[0].url;
+		return secure ? secure.url : this.spec.servers[0].url;
+	};
+	private cleanupBaseURL = (url: string): string => {
+		return url.replace(/{[a-zA-Z]+}/, (str) => {
```
StarpTech, line 706: This will fail `{ tld }` (spaces)

packages/sdk/src/v2openapi/index.ts
```diff
@@ -48,6 +48,7 @@ import {
 	applyNameSpaceToTypeFields,
 } from '../definition/namespacing';
 import { mapInputVariable } from '../configure';
+import { resolveConfigurationVariable } from '../../dist/configure/variables';
```
StarpTech, line 51: You cant reference to dist :smile:

packages/sdk/src/v2openapi/index.ts
```diff
@@ -226,6 +245,8 @@ class RESTApiBuilder {
 			ChildNodes: [],
 			Directives: [],
 		});
+		const baseURL = resolveConfigurationVariable(this.dataSources[this.dataSources.length - 1].Custom.Fetch.baseUrl!);
```
StarpTech, line 248: We should throw an error when the user passes `new EnvironmentVariables` in combination with base URL variables.

StarpTech (changes requested):

packages/sdk/src/v2openapi/index.ts
jensneuse: that's ok, it's invalid. we can add validation later

StarpTech (changes requested):
Could you also update the docs?

StarpTech (approved):
LGTM

PR #187, opened 2022-09-16
chore: switch baseURL to be string

jensneuse:


packages/protobuf/wgpb/wundernode_config.ts
```diff
@@ -857,7 +857,7 @@ export interface FetchConfiguration {
 	 */
 	urlEncodeBody: boolean;
 	mTLS: MTLSConfiguration | undefined;
-	baseUrl: ConfigurationVariable | undefined;
+	baseUrl: string;
```
StarpTech, line 860: baseURL is required now?

StarpTech (changes requested):

packages/protobuf/wgpb/wundernode_config.ts
jensneuse: No

StarpTech (approved):
LGTM

packages/protobuf/wgpb/wundernode_config.ts
jensneuse: In the generated config, it can be empty string. In the user submitted config, it can be optional (undefined).

PR #188, opened 2022-09-18
refactor: rewrite base typescript client

StarpTech:
## Testing

- [X] File uploads with Next.js testapp
- [X] SSR + Caching with Next.js testapp
- [X] CSRF with Next.js testapp
- [X] Fetch User with Next.js testapp
- [X] Login / Logout Github with Next.js testapp
- [X] Make csrf calls only when authentication was enabled
- [X] Realtime subscriptions with SEE
- [X] re-add subscribeOnce for subscriptions
- [x] Realtime subscriptions with Fetch
- [ ] More testing of the Next.js package (State handling is a bit hard. What means `refetching` as part of the result?)
- [x] Test subscriptions (Next.js and SWR)
- [ ] Validate compatibility with `WunderGraphContext`

## TODO

- [x] Move SWR hooks to a dedicated package. https://github.com/wundergraph/wundergraph/blob/dustin/eng-360-make-typescript-client-typesafe/packages/swr-hooks/README.md

## Highlights

- Rewritten the base client.
- Simplified response types. An operation either succeeds with `data` or fail with `errors`.
- Improved @Pagebakers type-safe client to avoid `any`. I could also drastically improve the types.
- Added E2E tests
- Allow the client to make decisions based on the operation e.g don't make CSRF calls for operations that don't require auth.
- Use AsyncGenerator instead of blocking through `while(true)` for real-time subscriptions over long polling.
- Removed the old `web.client.ts` client and ported the `next.js` package to the new one. The Next.js package still uses the old interface. In the future, we should refactor the Next.js implementation as well.


Pagebakers:
I agree there was a lot of room for improvement in the client, but the goal was to add a generic `Operations` type to have (optional) typesafety of the client which is now removed again, I'm not sure why, other than simplifying the types.

I don't think we have to rewrite the current Next.js client, as we will be replacing it SWR/React Query clients. Both will need to have their own withWunderGraph HOC for Next.js SSR support.

We also need a standalone WunderGraphProvider context provider with authentication / client.

packages/swr-hooks/README.md
```diff
@@ -0,0 +1,61 @@
...
+```
+
+Second, run `wunderctl generate` to generate the code.
+
+Now you can use the hooks.
+
+```ts
+import { createHooks } from '@wundergraph/swr-hooks';
+import { createClient, Mutations, Queries, Subscriptions } from './components/generated/client';
+const { useQuery, useMutation } = createHooks<Queries, Mutations, Subscriptions>(createClient());
```
Pagebakers, line 39: I like the createHooks factory, I had the same idea for the final swr package. But can we simplify it a bit by just passing `Operations` instead?

```ts
createHooks<Operations>()
```

We should probably also move the client to `<WunderGraphProvider client={createClient()}) />`, the hooks can then access the client using useWunderGraph / useClient, what do you think?

packages/swr-hooks/src/hooks.ts
```diff
@@ -0,0 +1,192 @@
...
+			},
+		};
+	};
+
+	/**
+	 * This is will subscribe to an operation and mutate the SWR state on result.
+	 *
+	 * @usage
+	 * ```ts
+	 * const unsubscribe = subscribeTo('Hello', {world: 'World'})
```
Pagebakers, line 136: The api changed to one options object

StarpTech:
> the goal was to add a generic Operations type to have (optional) typesafety of the client which is now removed again

Why do you think like that? It's still there.

> I don't think we have to rewrite the current Next.js client

The wundergraph client was overly complex. The next.js package still use parts of that (state handling) . During reading the implementation I've seen a lot of places to improve. My confidence to manage this package is also very low because there are no tests.

Pagebakers:
> > the goal was to add a generic Operations type to have (optional) typesafety of the client which is now removed again
> 
> Why do you think like that? It's still there.

Since this doesn't include typesafety for operations out of the box now.
```ts
import {createClient} from 'generated/client'
const client = createClient() // simply new Client() instead of new Client<Operations>()

client.query({
  operationName: 'Weather'
})
```

But now you need to annotate each operation call.

```ts
client.query<OperationRequestOptions<'Weather', { lat: number }>, { id: number }>({
  operationName: 'Weather'
})
```

> 
> > I don't think we have to rewrite the current Next.js client
> 
> The wundergraph client was overly complex. The next.js package still use parts of that (state handling) . During reading the implementation I've seen a lot of places to improve. My confidence to manage this package is also very low because there are no tests.

I was referring this this comment: `In the future, we should refactor the Next.js implementation as well.`
I did a lot of improvement in #120 but I most of the code can be dropped now anyway in favor of SWR

packages/swr-hooks/README.md
StarpTech: Good, less to type for the user. I forgot the context. I like the idea but it should be required like swr allows you to configure defaults.

StarpTech:
> Since this doesn't include typesafety for operations out of the box now.

Let's discuss this. I don't think this is a big deal to change that or going with this approach. In my opinion, the client will as always serve as a foundation and is never used directly.

> I did a lot of improvement in https://github.com/wundergraph/wundergraph/pull/120 but I most of the code can be dropped now anyway in favor of SWR

Sounds like a plan.

packages/swr-hooks/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "@wundergraph/swr-hooks",
```
StarpTech, line 2: Rename it to `swr`

packages/swr-hooks/src/hooks.ts
```diff
@@ -0,0 +1,192 @@
...
+	>(
+		options: UseMutationOptions<OperationName>
+	) => {
+		const { operationName, ...config } = options;
+		const response = useSWR<Data, GraphQLResponseError>(operationName, null, config);
+
+		return {
+			...response,
+			async mutate(options?: MutateOptions<Input, Data>): Promise<Data | undefined> {
+				const _options = {
```
StarpTech, line 120: Simplify?

packages/swr-hooks/src/hooks.ts
```diff
@@ -0,0 +1,192 @@
...
+	 */
+	const subscribeTo = (options: SubscriptionRequestOptions) => {
+		const abort = new AbortController();
+
+		options.abortSignal = abort.signal;
+
+		client.subscribe(options, (result: ClientResponse) => {
+			// Promise is not handled because we are not interested in the result
+			// Errors are handled by SWR internally
+			mutate({ operationName: options.operationName, input: options.input }, async () => {
```
StarpTech, line 147: Async not necessary

packages/nextjs/src/index.ts
```diff
@@ -487,6 +512,7 @@ function useSubscriptionContextWrapper<Input, Data, Role>(
 					result: ssrCache[cacheKey] as SubscriptionResult<Data>,
 				};
 			}
+			// TODO: What's the difference between "subscribeOnce" this and a single query call?
 			const promise = client.query({ ...subscription, ...args, subscribeOnce: true });
```
StarpTech, line 516: Convert to a single query call.

jensneuse:
@StarpTech do we need changes to lerna.json?

packages/nextjs/src/index.ts
jensneuse: Subscribe once makes the subscription/live query "immediately" return one single response.
This is useful when you want to server-side-render a subscription/live query.

jensneuse: Keep in mind that subscribe-once will make the "subscription/live query" synchronous. It will behave like a query.

StarpTech:
@jensneuse yes, good point. Added!

PR #189, opened 2022-09-19
DRAFT: legacy client fallback

Pagebakers:
#### Checklist

- [ ] run `make test`

### DIFF
--- a/packages/sdk/src/codegen/templates/index.ts
+++ b/packages/sdk/src/codegen/templates/index.ts
@@ -15,6 +15,7 @@ import { WunderGraphHooksPlugin } from './typescript/hooks';
 import { AuthenticationProviderConfiguration } from './markdown/authentication';
 import { WunderGraphWebhooksPlugin } from './typescript/webhooks';
 import { TypeScriptClient } from './typescript/client';
+import { TypeScriptLegacyWebClient } from './typescript/web.client';
 
 const typescriptModels = [
 	new TypeScriptInputModels(),
@@ -47,6 +48,7 @@ const templates = {
 		react: [
 			...typescriptAll,
 			new TypeScriptClient(),
+			new TypeScriptLegacyWebClient(),
 			new TypescriptReactProvider(),
 			new TypescriptReactHooks(),
 			new Forms(),
@@ -54,6 +56,7 @@ const templates = {
 		reactNative: [
 			...typescriptAll,
 			new TypeScriptClient(true),
+			new TypeScriptLegacyWebClient(),
 			new TypescriptReactNativeProvider(),
 			new TypescriptReactHooks(true),
 		],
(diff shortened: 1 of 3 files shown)


packages/sdk/src/codegen/templates/typescript/web.client.ts
```diff
@@ -0,0 +1,108 @@
...
+			hasSubscriptionsOrLiveQueries: _subscriptions.length + _liveQueries.length !== 0,
+			authProviders: config.authentication.cookieBased.map((provider) => provider.id),
+			hasAuthProviders: config.authentication.cookieBased.length !== 0,
+			hasS3Provider: config.application.S3UploadProvider.length > 0,
+			s3Provider: config.application.S3UploadProvider,
+			reactNative: this.reactNative,
+		});
+		return Promise.resolve([
+			{
+				path: 'wundergraph.client.ts',
```
StarpTech, line 48: That name is confusing. Can we rename it to e.g. `legacy.client.ts` ?

packages/sdk/src/codegen/templates/index.ts
```diff
@@ -47,13 +48,15 @@ const templates = {
 		react: [
 			...typescriptAll,
 			new TypeScriptClient(),
+			new TypeScriptLegacyWebClient(),
```
StarpTech, line 51: So previously, you could generate only the client. Is that intentional?

packages/sdk/src/codegen/templates/typescript/web.client.template.ts
```diff
@@ -0,0 +1,199 @@
+//language=handlebars
+export const template = `
+import type { {{ modelImports }} } from "./models";
+import type { RequestOptions, UserListener, Response } from "@wundergraph/sdk";
+import type { User, Role } from "./wundergraph.server";
```
StarpTech, line 5: This means you can only generate the client if you also generate the server. Do we somehow enforce that the basics things are generated like models and this? :thinking:

packages/sdk/src/codegen/templates/typescript/web.client.ts
```diff
@@ -0,0 +1,108 @@
...
+				doNotEditHeader: true,
+			},
+		]);
+	}
+}
+
+const filteredOperations = (application: ResolvedApplication, includeInternal: boolean) =>
+	includeInternal ? application.Operations : application.Operations.filter((op) => !op.Internal);
+
+export const operations = (application: ResolvedApplication, operationType: OperationType, includeInternal: boolean) =>
```
StarpTech, line 59: Those helper functions are already in helper.ts packages/sdk/src/codegen/templates/typescript/helpers.ts but are careful there might be minor differences.

StarpTech (changes requested):

packages/sdk/src/codegen/templates/index.ts
Pagebakers: I'm not sure what you mean.

packages/sdk/src/codegen/templates/typescript/web.client.template.ts
Pagebakers: It's always generated with templates.typescript.react. 

Same setup as previously.

StarpTech (approved):
LGTM

PR #196, opened 2022-09-20
feat: add post logout hook

jensneuse:
This hook allows you to take actions, e.g. call APIs/DBs, when the user logged out.
Example use case: Use 3rd party API to end session.


packages/sdk/src/middleware/plugins/hooks.ts
```diff
@@ -88,6 +88,23 @@ const FastifyHooksPlugin: FastifyPluginAsync<FastifyHooksOptions> = async (fasti
 			}
 		}
 	});
+	fastify.post('/authentication/postLogout', async (request, reply) => {
+		reply.type('application/json').code(200);
```
StarpTech, line 92: This is incorrect. If the user is not set the endpoint can't be called. We should return non-2xx. E.g 400 (bad request and only 200 after postLogout). This would also make our metrics in the future wrong.

StarpTech (changes requested):
Looks good expect one minor thing.

packages/sdk/src/middleware/plugins/hooks.ts
jensneuse: it's copy pasta from the other endpoints, are they wrong as well?

pkg/authentication/authentication.go
```diff
@@ -737,18 +756,20 @@ func (_ *CSRFTokenHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-	if !logoutOpenIDConnectProvider {
-		return
-	}
 	user := UserFromContext(r.Context())
 	if user == nil {
 		return
 	}
+	u.Hooks.handlePostLogout(r.Context(), user)
+	logoutOpenIDConnectProvider := r.URL.Query().Get("logout_openid_connect_provider") == "true"
+	if !logoutOpenIDConnectProvider {
```
YuriBuerov, line 770:
```
if !r.URL.Query().Has("logout_openid_connect_provider") {}
```

jensneuse: fails for "logout_openid_connect_provider=false"

packages/sdk/src/middleware/plugins/hooks.ts
StarpTech: This

```ts
	fastify.post('/authentication/postLogout', async (request, reply) => {
		if (!config.authentication?.postLogout || !request.ctx.user) {
			return reply.code(400)
		}
			try {
				const out = await config.authentication.postLogout(request.ctx);
				return reply.type('application/json').code(200).send({
					hook: 'mutatingPostAuthentication',
					response: out,
					setClientRequestHeaders: headersToObject(request.ctx.clientRequest.headers),
				})
			} catch (err) {
				request.log.error(err);
				return reply.code(500).send( { hook: 'mutatingPostAuthentication', error: err });
			}
	});
```

pkg/authentication/authentication.go
YuriBuerov: Btw, you can use existence for boolean parameter like ?logout_openid_connect_provider

like elasticsearch: `GET _cat/master?v` 

for example https://go.dev/play/p/wOysiK2j5gz

this way you don't need  param=true

StarpTech: But still ?x=false is interpreted as true.

YuriBuerov: yes, but the idea is to get rid of `?x=false` or `?x=true`  and have just `?x` which is valid URL param. if it's present - for us it's true, if not - false.

YuriBuerov: like in docs I linked above:
```
GET _cat/<operation_name>?h=<header_name_1>,<header_name_2>&v
Typically, for any operation you can find out what headers are available using the ?help parameter, and then use the ?h parameter to limit the output to only the headers that you care about.
```

jensneuse: I'm 100% sure that we always strip nullable args, so flag=false will not be the case anywhere. I think it's safe to make this change.

packages/sdk/src/middleware/plugins/hooks.ts
jensneuse: as we discussed, will be fixed in another PR

PR #198, opened 2022-09-21
fix: use node url in graphiql apiUrl

spetrunin:
#### Checklist

- [ ] run `make test`

### DIFF
--- a/pkg/apihandler/apihandler.go
+++ b/pkg/apihandler/apihandler.go
@@ -297,6 +297,7 @@ func (r *Builder) BuildAndMountApiHandler(ctx context.Context, router *mux.Route
 			log:           r.log,
 			html:          graphiql.GetGraphiqlPlaygroundHTML(),
 			apiPathPrefix: api.PathPrefix,
+			nodeUrl:       api.Options.NodeUrl,
 		}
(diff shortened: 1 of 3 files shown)


pkg/apihandler/apihandler.go
```diff
@@ -297,6 +297,7 @@ func (r *Builder) BuildAndMountApiHandler(ctx context.Context, router *mux.Route
 			log:           r.log,
 			html:          graphiql.GetGraphiqlPlaygroundHTML(),
 			apiPathPrefix: api.PathPrefix,
+			nodeUrl:       api.Options.NodeUrl,
```
jensneuse, line 300: What is the default value of NodeUrl? Would it be possible that it's empty? In this case, wouldn't it make sense to keep the old logic that checks for TLS and dynamically creates the apiURL?

spetrunin: it is never empty, as it has a fallback to default

it works like this:
1) no options provided
`http://localhost:9991`
2) port provided
`http://localhost:4444`
3) nodeUrl provided
uses nodeUrl

jensneuse (approved):

Issue #201, opened 2022-09-21
Uploading files pollutes the logs


jensneuse:
Currently, when running `wunderctl up --debug` we're logging all requests including req+res.
That's problematic in case of file uploads, because we're printing the file to the std out.
Ideally, we could detect if the request is a file upload and prevent logging the body to keep the logs clean.

jensneuse:
fixed

PR #203, opened 2022-09-21
fix: avoid logging file contents during uploads (#201)

fiam:
If the request is multipart, don't log the body contents. This doesn't prevent logging the file in absolutely all cases (e.g. a PUT with the contents as the body), but it will omit the messages as long as the upload comes from the WunderGraph SDK. Added notes in both places to make sure they're kept in sync.

#### Checklist

- [x] run `make test`
- [x ] commit message and code follows the [Code of conduct](https://github.com/wundergraph/wundergraph/blob/main/CODE_OF_CONDUCT.md)

### DIFF
--- a/packages/sdk/src/client/client.ts
+++ b/packages/sdk/src/client/client.ts
@@ -307,6 +307,9 @@ export class Client {
 	 * of file IDs that can be used to download the files from your S3 bucket.
 	 */
 	public async uploadFiles(config: UploadRequestOptions): Promise<UploadResponse> {
+		// XXX: apihandler.go uses the Content-Type header to detect file uploads
+		// and omit them from the logging messages. Any changes to the upload
+		// mechanism should also update apihandler.go accordingly.
 		const formData = new FormData();
 		for (const [_, file] of Object.entries(config.files)) {
 			if (file instanceof Blob) {
--- a/packages/sdk/src/client/client.ts
+++ b/packages/sdk/src/client/client.ts
@@ -307,9 +307,6 @@ export class Client {
 	 * of file IDs that can be used to download the files from your S3 bucket.
 	 */
 	public async uploadFiles(config: UploadRequestOptions): Promise<UploadResponse> {
-		// XXX: apihandler.go uses the Content-Type header to detect file uploads
-		// and omit them from the logging messages. Any changes to the upload
-		// mechanism should also update apihandler.go accordingly.
 		const formData = new FormData();
 		for (const [_, file] of Object.entries(config.files)) {
 			if (file instanceof Blob) {
--- a/pkg/apihandler/apihandler.go
+++ b/pkg/apihandler/apihandler.go
@@ -206,10 +206,18 @@ func (r *Builder) BuildAndMountApiHandler(ctx context.Context, router *mux.Route
 	r.router.Use(func(handler http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, request *http.Request) {
 			if r.enableDebugMode {
-				requestDump, err := httputil.DumpRequest(request, true)
+				// If the request looks like a file upload, avoid printing the whole
+				// encoded file as a debug message.
+				// File upload is handled in packages/sdk/client.ts, make sure to keep both in sync.
+				isMultipart := strings.HasPrefix(request.Header.Get("Content-Type"), "multipart/")
+				suffix := ""
+				if isMultipart {
+					suffix = "<multipart content omitted>"
+				}
+				requestDump, err := httputil.DumpRequest(request, !isMultipart)
 				if err == nil {
-					fmt.Printf("\n\n--- ClientRequest start ---\n\n%s\n\n\n\n--- ClientRequest end ---\n\n",
-						string(requestDump),
+					fmt.Printf("\n\n--- ClientRequest start ---\n\n%s%s\n\n\n\n--- ClientRequest end ---\n\n",
+						string(requestDump), suffix,
 					)
 				}
 			}
--- a/pkg/apihandler/apihandler.go
+++ b/pkg/apihandler/apihandler.go
@@ -306,6 +306,7 @@ func shouldLogRequestBody(request *http.Request) bool {
 	return !strings.HasPrefix(request.Header.Get("Content-Type"), "multipart/")
 }
 
+// returns a middleware that logs all requests to the given io.Writer
 func logRequestMiddleware(logger io.Writer) mux.MiddlewareFunc {
 	return func(handler http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, request *http.Request) {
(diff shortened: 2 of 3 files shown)


packages/sdk/src/client/client.ts
```diff
@@ -307,6 +307,9 @@ export class Client {
 	 * of file IDs that can be used to download the files from your S3 bucket.
 	 */
 	public async uploadFiles(config: UploadRequestOptions): Promise<UploadResponse> {
+		// XXX: apihandler.go uses the Content-Type header to detect file uploads
```
StarpTech, line 310: I like the initiative to document this but I don't think it is worth it. We use a compatible `fetch` implementation that always uses `multipart/form-data` for file uploads.

pkg/apihandler/apihandler.go
```diff
@@ -206,10 +206,18 @@ func (r *Builder) BuildAndMountApiHandler(ctx context.Context, router *mux.Route
 	r.router.Use(func(handler http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, request *http.Request) {
 			if r.enableDebugMode {
-				requestDump, err := httputil.DumpRequest(request, true)
+				// If the request looks like a file upload, avoid printing the whole
+				// encoded file as a debug message.
+				// File upload is handled in packages/sdk/client.ts, make sure to keep both in sync.
+				isMultipart := strings.HasPrefix(request.Header.Get("Content-Type"), "multipart/")
```
StarpTech, line 212: Could we put this in a helper `isMultipart` ?

I think it is safe to compare `multipart/form-data` instead of `multipart/`.

Do you think we should also ignore `x-www-form-urlencoded`?

StarpTech (changes requested):
Great PR. Only small suggestions.

pkg/apihandler/apihandler.go
```diff
@@ -206,10 +206,18 @@ func (r *Builder) BuildAndMountApiHandler(ctx context.Context, router *mux.Route
 	r.router.Use(func(handler http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, request *http.Request) {
 			if r.enableDebugMode {
-				requestDump, err := httputil.DumpRequest(request, true)
+				// If the request looks like a file upload, avoid printing the whole
```
StarpTech, line 209: I would love to see a simple test. You could export this as a middleware, pass a buffer and assert it after the request. WDYT?

```go
fmt.Fprintf(os.Stdout or buf, ....
```

packages/sdk/src/client/client.ts
fiam: Using a `PUT` with the file contents as the request body is not unheard of (e.g. S3 API works like that), but I agree the upload mechanism is unlikely to change and that comment might be "overdocumenting" a bit. I'll remove it.

StarpTech (changes requested):

packages/sdk/src/client/client.ts
StarpTech: Thank you.

pkg/apihandler/apihandler.go
fiam: That seems like a good idea, totally missed that possibility.

pkg/apihandler/apihandler.go
StarpTech: Sounds good!

pkg/apihandler/apihandler.go
```diff
@@ -318,6 +300,31 @@ func (r *Builder) BuildAndMountApiHandler(ctx context.Context, router *mux.Route
 	return streamClosers, err
 }
 
+func shouldLogRequestBody(request *http.Request) bool {
+	// If the request looks like a file upload, avoid printing the whole
+	// encoded file as a debug message.
+	return !strings.HasPrefix(request.Header.Get("Content-Type"), "multipart/")
```
StarpTech, line 306: As far as I know, there is only "multipart/form-data"

StarpTech (changes requested):

pkg/apihandler/apihandler_test.go
```diff
@@ -264,3 +266,36 @@ func TestQueryHandler_Caching(t *testing.T) {
 
 	assert.Equal(t, 2, resolver.invocations)
 }
+
+func TestLogMiddleware_Debug(t *testing.T) {
```
StarpTech, line 270: As a newbie, I'd not understand what this is about. Could you structure it in test cases or add meaningful messages to your asserts?

```go
assert.NotContains(buf.String, testFileContents, "Should not contain binary data")
```

StarpTech (changes requested):

pkg/apihandler/apihandler.go
fiam: There are more subtypes of `multipart`, but they're rather uncommon. I will change it to test for `multipart/form-data` if you prefer it that way, since for this use case it should work fine with any of those approaches.

pkg/apihandler/apihandler_test.go
fiam: I haven't had time to familiarise myself with the way you write tests (I'm a bit tight on time this week, sorry). I myself prefer to only use `t.Run()` for table-driven tests and not depend on 3rd party packages for testing values. Thanks for the clarifications, I'll redo them as you suggested.

pkg/apihandler/apihandler.go
```diff
@@ -203,19 +203,9 @@ func (r *Builder) BuildAndMountApiHandler(ctx context.Context, router *mux.Route
...
-				if err == nil {
-					fmt.Printf("\n\n--- ClientRequest start ---\n\n%s\n\n\n\n--- ClientRequest end ---\n\n",
-						string(requestDump),
-					)
-				}
-			}
-			handler.ServeHTTP(w, request)
-		})
-	})
+	if r.enableDebugMode {
```
fiam, line 206: Not sure if you noticed this. After checking that `r.enableDebugMode` never changes after initialisation, I moved the check from the middleware to `BuildAndMountApiHandler()`. Just wanted to make sure you don't plan to introduce any means to alter it at a later point, which would indeed require checking the value in every invocation.

pkg/apihandler/apihandler_test.go
StarpTech: We only use 3rd party packages if it makes sense. We use testify already in the codebase. I agree, for simple asserts its not needed but for `assert.Contains` it can increase readability.

pkg/apihandler/apihandler.go
StarpTech: Make sense.

StarpTech (approved):
LGTM

StarpTech:
PR titles have to start lowercase. I have fixed it for you.

jensneuse (approved):

pkg/apihandler/apihandler_test.go
fiam: Sorry, I was a bit too terse on my previous comment. Having test-only dependencies (or any kind of dependency for that matter) is perfectly fine, I’m just not used to use them. I agree that in this case it makes a lot of sense and ‘assert’ does make the intent much clearer.

fiam:
Thanks a lot for the reviews and clarifications. Much appreciated!

PR #204, opened 2022-09-21
fix(sdk): add fastify route only if the hook is configured

JivusAyrus:
#### Checklist

- [x] run `make test`


jensneuse:
Please always make sure to merge main into your PR branch.

jensneuse:
@JivusAyrus how can we verify that hooks still work? we should somehow test this. Ideas?

packages/sdk/src/middleware/plugins/hooks.ts
```diff
@@ -39,72 +39,105 @@ const FastifyHooksPlugin: FastifyPluginAsync<FastifyHooksOptions> = async (fasti
...
-				return { hook: 'revalidateAuthentication', error: err };
+			return {
+				hook: 'postAuthentication',
+			};
+		});
+	}
+
+	if (config.authentication?.mutatingPostAuthentication) {
+		fastify.post('/authentication/mutatingPostAuthentication', async (request, reply) => {
+			reply.type('application/json');
```
StarpTech, line 68: This is set automatically when we return json.

packages/sdk/src/middleware/plugins/hooks.ts
```diff
@@ -39,72 +39,105 @@ const FastifyHooksPlugin: FastifyPluginAsync<FastifyHooksOptions> = async (fasti
...
+			if (request.ctx.user === undefined) {
+				reply.code(400);
```
StarpTech, line 70: I recommend the chained syntax

```ts
reply.code(400).send({ hook: 'mutatingPostAuthentication', error: "User context doesn't exist" })
```

JivusAyrus:
Used postman to check if only the hooks which are used have the fastify routes and made sure that is the case.
To check if they are working appropriately, should we use the hooks in the test app and verify?

packages/sdk/src/middleware/plugins/hooks.ts
```diff
@@ -39,72 +39,105 @@ const FastifyHooksPlugin: FastifyPluginAsync<FastifyHooksOptions> = async (fasti
...
+			return {
+				hook: 'postAuthentication',
+			};
+		});
+	}
+
+	if (config.authentication?.mutatingPostAuthentication) {
+		fastify.post('/authentication/mutatingPostAuthentication', async (request, reply) => {
+			reply.type('application/json');
+			if (request.ctx.user === undefined) {
```
StarpTech, line 69: To avoid repetition, we could move this part into a [pre-handler](https://www.fastify.io/docs/latest/Reference/Hooks/#prehandler). 

You could archive that very easy with fastify [encapsulation](https://www.fastify.io/docs/latest/Reference/Encapsulation/) concept:

```ts
childServer.register(async function grandchildContext (fastify) {
	fastify.addHook('preHandler', (request, reply, done) => {
 	 	// validate user
  		done()
	})

	// register the hooks routes
})
```

StarpTech (changes requested):

StarpTech:
> To check if they are working appropriately, should we use the hooks in the test app and verify?

Ideally, we set up a simple test with fastify inject feature. https://www.fastify.io/docs/latest/Guides/Testing/#testing-with-http-injection

We have already created a server factory for that https://github.com/wundergraph/wundergraph/blob/suvij/eng-402-dont-add-fastify-route-when-route-isnt/packages/sdk/src/middleware/server.ts#L126

packages/sdk/src/middleware/plugins/hooks.ts
JivusAyrus: Removed reply.type().

packages/sdk/src/middleware/plugins/hooks.ts
JivusAyrus: Used preHandler and fixed it in the latest commit.

packages/sdk/src/middleware/plugins/hooks.ts
JivusAyrus: Done. Used chained syntax

JivusAyrus:
> > To check if they are working appropriately, should we use the hooks in the test app and verify?
> 
> Ideally, we set up a simple test with fastify inject feature. https://www.fastify.io/docs/latest/Guides/Testing/#testing-with-http-injection
> 
> We have already created a server factory for that https://github.com/wundergraph/wundergraph/blob/suvij/eng-402-dont-add-fastify-route-when-route-isnt/packages/sdk/src/middleware/server.ts#L126

I wrote tests to verify the working of the routes, verify if the routes have user context check and verify if only the hooks used have the routes configured.

packages/sdk/src/middleware/server.ts
```diff
@@ -184,22 +187,19 @@ export const createServer = async ({
...
-					clientRequest: {
-						headers: new Headers(req.body.__wg.clientRequest?.headers),
-						requestURI: req.body.__wg.clientRequest?.requestURI || '',
-						method: req.body.__wg.clientRequest?.method || 'GET',
-					},
-					internalClient: clientFactory({}, req.body.__wg.clientRequest),
-				};
-			}
-		);
+		fastify.addHook<{ Body: FastifyRequestBody }>('preHandler', async (req, reply) => {
```
StarpTech, line 190: In the future, I'd like to keep this file very thin and move middleware, handlers into different files with a test file.

packages/sdk/src/middleware/plugins/hooks.test.ts
```diff
@@ -0,0 +1,250 @@
...
+		method: 'POST',
+		url: '/authentication/postLogout',
+		payload: {
+			__wg: {
+				user: {},
+				clientRequest: {},
+			},
+		} as FastifyRequestBody,
+	});
+	expect(postLogoutResponse.statusCode).toEqual(200);
```
StarpTech, line 109: We should also expect the response payload.

StarpTech (changes requested):

JivusAyrus:
Have tested the hooks using the test apps(default and nextjs) manually.

StarpTech (approved):
LGTM

PR #207, opened 2022-09-23
fix: allow to configure internal and public node urls separately

spetrunin:
#### Checklist

- [ ] run `make test`


jensneuse (approved):

packages/sdk/src/configure/index.ts
```diff
@@ -817,17 +817,17 @@ export const configureWunderGraphApplication = (config: WunderGraphConfigApplica
 
 			done();
 
-			let nodeUrl = resolveConfigurationVariable(resolved.nodeOptions.nodeUrl);
+			let publicNodeUrl = resolveConfigurationVariable(resolved.nodeOptions.publicNodeUrl);
 			// trim trailing slash if any
-			nodeUrl = nodeUrl.endsWith('/') ? nodeUrl.slice(0, -1) : nodeUrl;
```
StarpTech, line 822: Could you put this in a helper fn?

packages/sdk/src/configure/options.ts
```diff
@@ -99,10 +103,15 @@ export const resolveNodeOptions = (options?: NodeOptions): ResolvedNodeOptions =
 		? new EnvironmentVariable(WgEnv.NodeUrl, `http://localhost:${options?.listen?.port}`)
 		: DefaultNodeOptions.nodeUrl;
 
+	const fallbackPublicNodeUrl = options?.listen?.port
```
StarpTech, line 106: Why is the `port` significant for the condition? I'd say the whole URL is relevant.

StarpTech (changes requested):

packages/sdk/src/configure/options.ts
```diff
@@ -94,15 +102,19 @@ export interface ResolvedServerLogger {
...
+export const resolveNodeOptions = (options?: NodeOptions): ResolvedNodeOptions => {
 	let nodeOptions = isCloud
 		? DefaultNodeOptions
 		: {
-				nodeUrl: options?.nodeUrl || fallbackNodeUrl,
+				nodeUrl:
+					options?.nodeUrl || new EnvironmentVariable(WgEnv.NodeUrl, fallbackUrl(defaultNodePort, options?.listen)),
+				publicNodeUrl:
+					options?.publicNodeUrl ||
+					new EnvironmentVariable(WgEnv.PublicNodeUrl, fallbackUrl(defaultNodePort, options?.listen)),
```
StarpTech, line 117: Shouldn't it be?

```ts
const nodeUrl = options?.nodeUrl || new EnvironmentVariable(WgEnv.NodeUrl, fallbackUrl(defaultNodePort, options?.listen))
const publicNodeUrl = options?.publicNodeUrl || nodeUrl,
```

StarpTech (changes requested):

packages/sdk/src/configure/options.ts
spetrunin: than potentially it will share the same env var, which is not desired I believe

StarpTech: Make sense, thanks.

StarpTech (approved):
LGTM

pkg/node/node.go
```diff
@@ -416,7 +416,7 @@ func (n *Node) startServer(nodeConfig WunderNodeConfig) error {
 	}
 
 	n.log.Debug("node url",
-		abstractlogger.String("nodeUrl", nodeConfig.Api.Options.NodeUrl),
+		abstractlogger.String("nodeUrl", nodeConfig.Api.Options.PublicNodeUrl),
```
StarpTech, line 419: Does it make sense to print here a URL that is not controlled by the node?

StarpTech (changes requested):

PR #209, opened 2022-09-23
fix(codegen): resolve template dependencies recursively

StarpTech:
#### Checklist

- [ ] run `make test`

We forgot to collect all template dependencies recursively. I run into an issue that specifying the `fastifyServer` template was not enough because it has template dependencies that requires other templates. 

```ts
	codeGenerators: [
		{
			templates: [templates.typescript.client, templates.typescript.fastifyServer],
			path: '../../web/src/generated',
		},
	],
```

I also added a max template depth to protect against the infinite loop.


packages/sdk/src/codegen/index.ts
```diff
@@ -37,33 +37,31 @@ const ensurePath = (filePath: string) => {
 	}
 };
 
-export const doNotEditHeader = '// Code generated by wunderctl. DO NOT EDIT.\n\n';
+export const collectAllTemplates = (templates: Template[], level = 0) => {
+	if (level > 10) {
+		throw new Error('Template dependency loop detected');
```
jensneuse, line 42: Does it make sense to throw? or just stop? Recursive dependencies might happen, but shouldn't be a problem I'd say.

jensneuse (changes requested):

packages/sdk/src/codegen/index.ts
StarpTech: Definitely, if you have more than 10 levels, you are definitely doing something wrong.

StarpTech: We can also return it and log an error when a specific depth has been reached.

jensneuse (approved):

jensneuse:
LGTM

PR #211, opened 2022-09-24
fix: remove invalid hasDotWunderGraphDirectory option

StarpTech:
#### Checklist

- [ ] run `make test`


jensneuse:
So, this code is correct and we could merge it.
However, I think that the real problem is that we don't know in which directory the user installed `.wundergraph`.
Is the WunderGraph dir in the root of a project, or are we in a monorepo and have some nesting?

For VSCode it's crucial that the `.graphqlconfig` sits at the root of the project. Otherwise, IntelliSense won't work.
Ideally, we could configure an absolute or relative path where the `.graphqlconfig` should be generated and updated.

docs-website/src/pages/docs/wundergraph-config-ts-reference/configure-graphqlconfig.md
```diff
@@ -4,27 +4,23 @@ pageTitle: 'WunderGraph - Configure .graphqlconfig'
 description:
 ---
 
-This section describes how to set up the configuration for generating the `.graphqlconfig` file.
-
-This file is very important for WunderGraph to work.
+This `.graphqlconfig` file is very important for WunderGraph to work.
```
jensneuse, line 7: Updated proposal:

This `.graphqlconfig` file is very important for an ideal developer experience.
IDEs like VSCode and Jetbrains recognise the file and use it to configure IntelliSense.
It's important to place it in the right directory.
E.g. in VSCode, it's important that you place it at the root of the project.
With the `.graphqlconfig` properly configured, your IDE helps you to define GraphQL Operations.

jensneuse (changes requested):

StarpTech:
> Ideally, we could configure an absolute or relative path where the .graphqlconfig should be generated and updated.

I tried this already, but it's not that easy to do because you user can only specify a relative path.

Example:

```ts
configureWunderGraphApplication({
  dotGraphQLConfig: {
    disable: true,
	path: '../../.graphqlconfig'
  },
})
```

How should we calculate the schema path? We would need to calculate the target directory and search for the nearest `.wundergraph` directory.

jensneuse:
@StarpTech update from main+merge or discard?

bmcilw1:
> So, this code is correct and we could merge it. However, I think that the real problem is that we don't know in which directory the user installed `.wundergraph`. Is the WunderGraph dir in the root of a project, or are we in a monorepo and have some nesting?
> 
> For VSCode it's crucial that the `.graphqlconfig` sits at the root of the project. Otherwise, IntelliSense won't work. Ideally, we could configure an absolute or relative path where the `.graphqlconfig` should be generated and updated.

Can both of these challenges be addressed by setting an ENV var on the startup of the CLI? I believe both VSCode plugins and the running CLI process could reference an ENV var if set by something similar to #300

jensneuse:
@fiam what should we do with this? I think you were working on a similar issue.

fiam:
@jensneuse I think this one can be closed now, superseded by #537

StarpTech:
Yeah, let's close it.

PR #214, opened 2022-09-26
feat:  add graphql introspection middleware

spetrunin:
#### Checklist

- [ ] run `make test`


jensneuse:
Don't forget to update the docs

packages/sdk/src/definition/federation.ts
```diff
@@ -13,15 +14,26 @@ export const isFederationService = (schema: GraphQLSchema): boolean => {
...
+
+	let opts: AxiosRequestConfig = {
 		headers: {
 			'Content-Type': 'application/json',
 			Accept: 'application/json',
 		},
-	});
+	};
+
+	if (requestMiddleware) {
```
StarpTech, line 32: I read the issue, but I'm still not convinced why we need to provide a middleware mechanism. Could you elaborate?

packages/sdk/src/definition/index.ts
```diff
@@ -248,6 +248,12 @@ export interface DataSource<Custom = unknown> {
 	Directives: DirectiveConfiguration[];
 }
 
+export type RequestMiddleware = (requestConfig: AxiosRequestConfig) => AxiosRequestConfig;
```
StarpTech, line 251: Axios is an implementation detail. In the future, we might switch the client. Don't expose the type to the user.

StarpTech (changes requested):

packages/sdk/src/definition/index.ts
spetrunin: I need a detailed specification in this case - like what should be possible

packages/sdk/src/definition/federation.ts
spetrunin: At the moment there is only one way to authenticate an introspection request - add a StaticHeader 
but this means that such header will be used also for all requests

So we have a need in some way to provide various auth options for the introspection calls
open to any proposals

packages/sdk/src/definition/index.ts
jensneuse: I actually think that at this moment, a function that takes the headers and allows us to override them should be enough.

spetrunin: could be there a basic auth?

jensneuse: yes, it's possible, but currently we don't have this use case
maybe you find a way to keep it flexible enough but keeping it simple?

packages/sdk/src/definition/federation.ts
StarpTech: What about?

```ts
const spacex = introspect.graphql({
	apiNamespace: 'spacex',
	headers: (builder) => builder.addClientRequestHeader('X-Authorization', 'Authorization'),
	introspection: {
		headers: (builder) => builder.addClientRequestHeader('X-Authorization', 'Authorization'),
	}
});
```

My favorite API would look like this :smile: Exposing a builder or middleware is powerful but I see no reason to add this complexity here. It's also not simple to use.

```ts
const spacex = introspect.graphql({
	apiNamespace: 'spacex',
	headers: [
		Headers.static('X-Authorization', 'Authorization'),
		Headers.client('X-Authorization', 'Authorization')],
	introspection: {
		headers: [
			Headers.static('X-Authorization', 'Authorization'),
			Headers.auth('username', 'password')],
	},
});
```

jensneuse (approved):
LGTM. I've only added a few minor changes in the docs.

docs-website/config/tags.js
```diff
@@ -281,7 +281,7 @@ const tags = {
 	'configure mTLS':
 		'/docs/wundergraph-config-ts-reference/configure-mtls-for-http-based-data-sources',
 	'configure introspection for protected API':
-		'/docs/wundergraph-config-ts-reference/configure-introspection#introspection-of-protected-graphql-api',
+		'/docs/wundergraph-config-ts-reference/configure-introspection#introspection-of-protected-graph-ql-ap-is',
```
StarpTech, line 284: `graph-ql`

StarpTech (changes requested):

docs-website/config/tags.js
spetrunin: it is how nextjs generated an anchor from `## heading`

spetrunin: https://github.com/wundergraph/wundergraph/pull/214/files#diff-a88d471cf72fc2ee1fffbe5d4cf67db38d8c45ffc3e5e702cfd9281a598361eeR81

header is here

StarpTech (approved):
Great work!

PR #216, opened 2022-09-27
fix(authentication): userInfoEndpoint should be optional for token auth providers

xzyfer:
Currently token auth providers without a `userInfoEndpoint` will always result in an `undefined` user in the wundergraph context object.

#### Checklist

- [x] run `make test`


jensneuse (approved):

jensneuse:
@StarpTech please verify and merge. Looks good to me.

PR #217, opened 2022-09-27
feat: sse subscriptions configuration

YuriBuerov:
#### Checklist

- [ ] run `make test`


packages/sdk/src/definition/index.ts
```diff
@@ -406,6 +406,7 @@ export interface GraphQLUpstream extends HTTPUpstream {
 	baseUrl?: InputVariable;
 	path?: InputVariable;
 	subscriptionsURL?: InputVariable;
+	useSSEForSubscription?: InputVariable;
```
jensneuse, line 409: as we've used the subscriptions prefix for URL, I'd suggest we call this `subscriptionsUseSSE`.

jensneuse: this also means that you can use `true` instead of `"true"`.

jensneuse (changes requested):

packages/sdk/src/definition/index.ts
YuriBuerov: makes sense

packages/protobuf/wgpb/wundernode_config.ts
```diff
@@ -793,6 +793,7 @@ export interface DataSourceCustomStatic {
 export interface GraphQLSubscriptionConfiguration {
 	enabled: boolean;
 	url: ConfigurationVariable | undefined;
+	useSSE: boolean;
```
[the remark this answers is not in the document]
StarpTech: nevermind, we use it :smile:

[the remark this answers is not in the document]
jensneuse: https://www.youtube.com/watch?v=VnT7pT6zCcA

packages/sdk/src/definition/index.ts
```diff
@@ -406,6 +406,7 @@ export interface GraphQLUpstream extends HTTPUpstream {
 	baseUrl?: InputVariable;
 	path?: InputVariable;
 	subscriptionsURL?: InputVariable;
+	subscriptionUseSSE?: boolean;
```
jensneuse, line 409: should be plural ("subscriptions") because we're using plural for subscriptionsURL too

pkg/apihandler/apitransport.go
```diff
@@ -118,6 +118,9 @@ func (t *ApiTransport) roundTrip(request *http.Request) (res *http.Response, err
 	// if you're doing this after calling the onRequest hook, it won't work
 	isUpgradeRequest := websocket.IsWebSocketUpgrade(request)
 
+	// on stream request we shouldn't read response body
+	isStreamingRequest := request.Header.Get("Accept") == "text/event-stream"
```
jensneuse, line 122: we discussed that we don't have to check here, instead we set a config param on the transport when we create the "subscription" http client (see comment below)

pkg/engineconfigloader/engineconfigloader.go
```diff
@@ -48,12 +48,17 @@ func NewDefaultFactoryResolver(transportFactory ApiTransportFactory, baseTranspo
 		Timeout:   time.Second * 10,
 		Transport: transportFactory(baseTransport),
 	}
+	streamingClient := &http.Client{
+		Transport: transportFactory(baseTransport),
```
jensneuse, line 52: Here you should add a param to set this transport to "streaming" mode.
This way, you know in the transport without checking for the response headers (might vary) that the transport is used within the context of a streaming client.

jensneuse (changes requested):

pkg/apihandler/apitransport.go
YuriBuerov: I do remember that we discussed it, but are there any other use cases for this config param apart from SSE? 
As for me, it looks consistent because this helper `isUpgradeRequest := websocket.IsWebSocketUpgrade(request)` also just checks headers. I'm ok if you want me to change this.

StarpTech (changes requested):
LGTM, but I'm missing an example in the testapps and examples.

YuriBuerov:
@StarpTech  I'm going to add examples with docs for examples in the separate PR after release. Not sure about testapps though, I'm testing with the default test app,  do you want me to include it in this PR? (I'll need to add graphql servers, one for WS, and one for SSE respectively). All of this we will have in examples later.

jensneuse (approved):

jensneuse:
Changes LGTM.
Resolve conflicts an we can merge.

StarpTech:
@YuriBuerov is fine for me if we have at least one example to test it. @jensneuse please don't be so rushed with merging, I had open threads.

PR #220, opened 2022-09-29
fix: sync path of prisma introspection and prisma runtime

jensneuse:
Previously, prisma introspection was started with working dir `project/.wundergraph` while prisma execution was started with working dir `project`. This wasn't really a problem for Postgres et al, because we don't care much about the filesystem. However, with SQLite, the database is a file. This meant that introspection worked, but the execution didn't, or vice versa. With this change, we sync both to start with working dir `project/.wundergraph`.

I've also added a few minor refactorings to make resolving the wundergraph dir simpler as well as wait for the lazy engine until it's ready in small intervals (10ms, previously 500) to start faster during development (not relevant for production).


cli/commands/root.go
```diff
@@ -61,18 +64,27 @@ Simply running "wunderctl" will check the wundergraph.manifest.json in the curre
 wunderctl is gathering anonymous usage data so that we can better understand how it's being used and improve it.
 You can opt out of this by setting the following environment variable: WUNDERGRAPH_DISABLE_METRICS
 `,
-	PersistentPreRun: func(cmd *cobra.Command, args []string) {
+	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+
+		var err error
+		WunderGraphDir, err = files.FindWunderGraphDir(_wunderGraphDirConfig)
```
StarpTech, line 70: This method already returns the absolute wundergraph dir. This method was also used for `AbsWorkingDir` :smile: 

image

StarpTech (changes requested):

cli/commands/install-prisma-dependencies.go
```diff
@@ -3,20 +3,14 @@ package commands
 import (
 	"github.com/spf13/cobra"
 	"github.com/wundergraph/wundergraph/pkg/datasources/database"
-	"github.com/wundergraph/wundergraph/pkg/files"
 )
 
 // introspectCmd represents the introspect command
 var ensurePrismaCmd = &cobra.Command{
 	Use:   "installPrismaDependencies",
```
spetrunin, line 10: where do we use this cli command?

jensneuse: the purpose is to run this in CI so that you don't have to download the prisma dependency at runtime/start of the process

cli/commands/root.go
jensneuse: removed

spetrunin:
as you already touching db introspection

does it make sense to move all subcommands of introspection under `introspection` folder?

jensneuse:
> as you already touching db introspection
> 
> does it make sense to move all subcommands of introspection under `introspection` folder?

It's a standard from viper to generate all cmd files in one directory (package) which also makes sharing some global variables like WunderGraphDir easier.

cli/commands/install-prisma-dependencies.go
spetrunin: we don't use it on our own and it is not present in docs on a website

so should it we add it to docs?

cli/commands/root.go
```diff
@@ -61,18 +62,25 @@ Simply running "wunderctl" will check the wundergraph.manifest.json in the curre
...
+		}
+
 		if enableDebugMode {
 			log = buildLogger(abstractlogger.DebugLevel)
 		} else {
 			log = buildLogger(findLogLevel(abstractlogger.ErrorLevel))
 		}
 
-		err := godotenv.Load(DotEnvFile)
+		err = godotenv.Load(DotEnvFile)
```
YuriBuerov, line 79:
```
if err := godotenv.Load(DotEnvFile); err != nil {
   ...
}
```
this way you can remove `var err error` (Line 67)

StarpTech (approved):
LGTM

PR #222, opened 2022-09-29
fix: synchronize prisma install for multiple prisma datasources

jensneuse:


spetrunin (approved):

packages/sdk/src/db/introspection.ts
```diff
@@ -23,10 +23,24 @@ export interface PrismaDatabaseIntrospectionResult {
...
+};
+
+let ensurePrisma: Promise<void> | undefined;
+
 const introspectPrismaDatabase = async (
 	databaseURL: string,
 	databaseSchema: DatabaseSchema
 ): Promise<PrismaDatabaseIntrospectionResult> => {
+	if (!ensurePrisma) {
+		ensurePrisma = _ensurePrisma();
```
StarpTech, line 41: How will it solve the issue? According to the code, `introspectPrismaDatabase` is called by all SDK introspection methods, and we run all asynchronously.

StarpTech (changes requested):

packages/sdk/src/db/introspection.ts
StarpTech: I missed the singleton here. Looks fine, although it would be better to do it once and not calling it after each "generate".

StarpTech (approved):
LGTM

PR #226, opened 2022-09-30
fix: re-enable to specify wundergraph dir via arg

jensneuse:
I'd still like to merge this. I think. it makes sense to make this configurable, especially for mono repos.
I don't see why it shouldn't be possible.
In the end, we git clone a repo, it contains a package.json in the root. This package.json could use the `--wundergraph-dir`arg or not, it's up to the user, and shouldn't make a difference for us.
Why dictate patterns if you don't have to?
Should we make it a convention? I think yes. Can we make it configurable? I think so too.


cli/commands/up.go
```diff
@@ -2,6 +2,7 @@ package commands
 
 import (
```
StarpTech, line 3: Did you test `wunderctl start`, `wunderctl node start`, `wunderctl server start` ?

packages/sdk/src/wunderctlexec/index.ts
```diff
@@ -26,6 +29,9 @@ export const wunderctlExecAsync = async (args: WunderCtlExecArgs): Promise<strin
 	if (!fs.existsSync(file)) {
 		throw new Error('wunderctl binary not found');
 	}
+	if (process.env.WG_DIR_ABS) {
+		args.cmd.push('--wundergraph-dir-abs', process.env.WG_DIR_ABS);
```
StarpTech, line 33: It feels odd to work with a flag instead setting the right process working directory :confused: 

Can't we do?

```ts
const subprocess = execa(file, args.cmd, {
		timeout: args.timeout,
		cwd: process.env.WG_DIR_ABS
```

StarpTech (changes requested):

StarpTech:
I checked all the other places. Looks good.

testapps/renamed/.gitignore
```diff
@@ -0,0 +1,41 @@
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
```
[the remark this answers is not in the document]
jensneuse: testing with a renamed project

cli/commands/root.go
```diff
@@ -150,6 +155,7 @@ func init() {
 	rootCmd.PersistentFlags().BoolVar(&enableDebugMode, "debug", false, "enables the debug mode so that all requests and responses will be logged")
 	rootCmd.PersistentFlags().BoolVar(&jsonEncodedLogging, "json-encoded-logging", false, "switches the logging to json encoded logging")
 	rootCmd.PersistentFlags().StringVar(&_wunderGraphDirConfig, "wundergraph-dir", files.WunderGraphDirName, "path to your .wundergraph directory")
```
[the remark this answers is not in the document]
jensneuse: this would be a breaking change

packages/sdk/src/wunderctlexec/index.ts
jensneuse: when seeting via cwd, we'd have to set `--wundergraph-dir .`, right?

testapps/renamed/.gitignore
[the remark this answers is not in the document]
StarpTech: We should not merge it.

cli/commands/root.go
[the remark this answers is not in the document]
StarpTech: Isn't it up to the user 

- `wunderctl --wundergraph-dir .wundergraph`
- `wunderctl --wundergraph-dir /users/dustin/myproject/.wundergraph` ?

packages/sdk/src/wunderctlexec/index.ts
StarpTech: I don't think so because the user has to pass `--wundergraph-dir` when he wants to overwrite the default.

```
"start": "wunderctl up --wundergraph-dir '/efer/fref'"
```

cli/commands/up.go
jensneuse: yes

jensneuse:
please check again

StarpTech (approved):
LGTM

PR #229, opened 2022-10-03
chore: add linux/arm64 wunderctl binaries to releases

fiam:
#### Checklist

- [x] run `make test`
- [-] tests and/or benchmarks are included
- [-] documentation is changed or added


fiam:
> Hi, is this PR only about adding ARM binaries to GitHub release or to make wunderctl work on arm? I ask this because it would require changes in other places.

Thanks for the heads up. I incorrectly assumed it already worked on `linux/arm64` because it does work on `darwin/arm64`. I'll dig and see what needs to be changed. Closing this PR for the time being.

PR #230, opened 2022-10-04
fix: introspection, infinite loop in file changes

YuriBuerov:
#### Checklist

- [ ] run `make test`

When we add a datasource that creates non-deterministic introspection results when the same data is in a different order. 
As a result, the introspection cache is always 'uniq' which causes infinite `watcher` calls .


StarpTech (approved):
Great solution!

PR #231, opened 2022-10-04
chore: add introspection fetcher with retries

spetrunin:
#### Checklist

- [ ] run `make test`

### DIFF
--- a/packages/sdk/src/definition/introspection-fetcher.ts
+++ b/packages/sdk/src/definition/introspection-fetcher.ts
@@ -1,5 +1,5 @@
 import axiosRetry from 'axios-retry';
-import axios, { AxiosError, AxiosInstance } from 'axios';
+import axios, { AxiosError, AxiosInstance, AxiosRequestConfig } from 'axios';
 
 let axiosInstance: AxiosInstance | undefined;
 
@@ -20,14 +20,9 @@ const initAxios = (): AxiosInstance => {
 		retryDelay: axiosRetry.exponentialDelay,
 		retryCondition: (error: AxiosError) => {
 			if (error.response) {
-				return error.response.status >= 400;
+				return error.response.status >= 500;
 			}
-			return true;
-		},
-		onRetry: (retryCount, error, requestConfig) => {
-			console.log(
-				`failed to perform request method: ${requestConfig.method} url: ${requestConfig.url}. Retry attempt #${retryCount}`
-			);
+			return axiosRetry.isNetworkError(error);
 		},
 	});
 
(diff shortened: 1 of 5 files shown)


packages/sdk/src/definition/introspection-fetcher.ts
```diff
@@ -0,0 +1,35 @@
...
+
+const initAxios = (): AxiosInstance => {
+	const instance = axios.create();
+
+	axiosRetry(instance, {
+		retries: 5,
+		retryDelay: axiosRetry.exponentialDelay,
+		retryCondition: (error: AxiosError) => {
+			if (error.response) {
+				return error.response.status >= 400;
```
StarpTech, line 23: I'd retry only on 5xx and network errors. 40xx means the request is not in the proper state.

packages/sdk/src/definition/introspection-fetcher.ts
```diff
@@ -0,0 +1,35 @@
...
+			}
+			return true;
+		},
+		onRetry: (retryCount, error, requestConfig) => {
+			console.log(
+				`failed to perform request method: ${requestConfig.method} url: ${requestConfig.url}. Retry attempt #${retryCount}`
```
StarpTech, line 29: Can we make it a bit clear to what this retry belongs? E.g. with a custom log message for the introspection part.

StarpTech (changes requested):

packages/sdk/src/definition/introspection-fetcher.ts
```diff
@@ -0,0 +1,30 @@
...
+	return axiosInstance;
+};
+
+const initAxios = (): AxiosInstance => {
+	const instance = axios.create();
+
+	axiosRetry(instance, {
+		retries: 5,
+		retryDelay: axiosRetry.exponentialDelay,
+		retryCondition: (error: AxiosError) => {
```
StarpTech, line 21: You can just use the default behaviour or `axiosRetry.isNetworkOrIdempotentRequestError`

StarpTech (changes requested):

packages/sdk/src/definition/introspection-fetcher.ts
spetrunin: According to this:
https://github.com/softonic/axios-retry/blob/master/es/index.mjs#L19

IsIdempotentRequestError - will return false for `post` request

spetrunin: yep, checked - no retries in case we use that

StarpTech: Good point.

StarpTech (approved):
LGTM

PR #232, opened 2022-10-05
feat: add configurable per-source timeouts

fiam:
- Add a new optional timeout field in HTTPUpstream
- Pass this new field to the node via FetchConfiguration, so it applies
  to all HTTP based sources
- In the server, create a dedicated http.Client with the specified
  timeout as needed

#### Checklist

- [x] run `make test`


jensneuse:
I think there should be a global default value we can pass to `configureWunderGraphApplication.options`.
Additionally, what's the behaviour for databases (Prisma).
We use an http client for prisma, too. We should also be able to configure timeouts there.

docs-website/src/pages/docs/wundergraph-config-ts-reference/configure-openapi-rest-data-source.md
```diff
@@ -19,6 +19,7 @@ const jsp = introspect.openApi({
   introspection: {
     pollingIntervalSeconds: 2,
   },
+  timeout: 10 * 1000, // in milliseconds, optional
```
StarpTech, line 22: Can we call it `timeoutInMs`? Very often, I see "timeout" in libraries, and it's not obvious about the unit.

docs-website/src/pages/docs/wundergraph-config-ts-reference/configure-graphql-data-source.md
```diff
@@ -14,6 +14,7 @@ The `GraphQL` data source allows you to introspect GraphQL services and add them
 const countries = introspect.graphql({
   apiNamespace: 'countries',
   url: 'https://countries.trevorblades.com/',
+  timeout: 10 * 1000, // in milliseconds, optional
```
StarpTech, line 17: Does `0` means unlimited?

StarpTech (changes requested):

docs-website/src/pages/docs/wundergraph-config-ts-reference/configure-openapi-rest-data-source.md
fiam: Definitely, that's a better idea. We have `pollingIntervalSeconds` somewhere else, though, so is `timeoutMilliseconds` better for consistency or would you prefer `timeoutMs` to save the user some typing?

docs-website/src/pages/docs/wundergraph-config-ts-reference/configure-graphql-data-source.md
fiam: No, zero means the default, which can be specified globally per node, or left as the 10 seconds default.

docs-website/src/pages/docs/wundergraph-config-ts-reference/configure-openapi-rest-data-source.md
StarpTech: `timeoutMilliseconds` is good.

StarpTech: Or why not use seconds too?

docs-website/src/pages/docs/wundergraph-config-ts-reference/configure-graphql-data-source.md
StarpTech: That's confusing. I'd prefer `0` to be unlimited and the default is only used when no value has been set.

fiam: I'd prefer to not have any means to set an unlimited timeout. Unlimited is an unintentional error 99% of the time, plus one can achieve the same result in practical (but quite explicit) terms setting it to a ridicuously high value.

StarpTech: I have no strong opinion here. Both are fine to me. Setting a zero explicitly is an act but I also understand your point.

fiam: I think I will change it to what you suggested. I feel a bit uneasy about not having a timeout, but maybe I'm being too paranoid. As you said, if they set it to zero manually without understanding the consequences, then it's their own fault.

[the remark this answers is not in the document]
fiam: It's a timeout for all requests sent from the node when servicing that data source. We could introduce more granular timeouts (e.g. introspection, time to connect/first bytes, time for whole response), but it's non-trivial and a general timeout would probably work for most use cases.

docs-website/src/pages/docs/wundergraph-config-ts-reference/configure-openapi-rest-data-source.md
StarpTech: You have to explain why seconds are not granular enough :smile:

packages/sdk/src/definition/graphql-introspection.ts
```diff
@@ -138,6 +138,7 @@ export const introspectGraphql = async (introspection: GraphQLIntrospection): Pr
 						UpstreamSchema: schemaSDL,
 					},
 					Directives: applyNamespaceToDirectiveConfiguration(schema, introspection.apiNamespace),
+					TimeoutMilliseconds: introspection.timeoutMilliseconds ?? -1,
```
StarpTech, line 141: What means `-1`?

packages/sdk/src/definition/index.ts
```diff
@@ -280,6 +282,7 @@ export interface HTTPUpstream extends IntrospectionConfiguration {
 	headers?: (builder: IHeadersBuilder) => IHeadersBuilder;
 	authentication?: HTTPUpstreamAuthentication;
 	mTLS?: HTTPmTlsConfiguration;
+	timeout?: number;
```
StarpTech, line 285: Why is it called "timeout" here?

docs-website/src/pages/docs/wundergraph-config-ts-reference/configure-openapi-rest-data-source.md
fiam: Imagine this use case. Let's say I want to generate a response in less than 1 second (which is quite high already) and I have 2 data sources than can do the more or less the same thing, but one is richer than the other (e.g. gas prices on a given area: one gives me all stations, another one just one brand) but sometimes it takes too much time. If the richer source doesn't return any data in let's say 500ms, I'd rather query the other one and still present some data to the user. That's a use case that I think we should support, specially when it takes so little code on our side.

packages/sdk/src/definition/index.ts
```diff
@@ -269,6 +270,7 @@ export interface DatabaseIntrospection extends IntrospectionConfiguration {
 }
 
 export interface IntrospectionConfiguration {
+	timeoutMilliseconds?: number;
```
StarpTech, line 273: I agree @jensneuse we should be clear about what this timeout means. What do you think about e.g.

```ts
const spacex = introspect.graphql({
	apiNamespace: 'spacex',
    requestTimeoutMilliseconds: 5000
});
```

StarpTech (changes requested):

packages/sdk/src/definition/graphql-introspection.ts
fiam: Any value <= 0 means "keep the default". That avoids having to represent a `null | undefined` on the Go side.

packages/sdk/src/definition/index.ts
fiam: Good catch. That's a leftover from the previous implementation. It's unused at this point and I need to delete it.

packages/sdk/src/definition/index.ts
fiam: That definitely makes it more explicit, although a bit too long. I will change it.

StarpTech: or use a nested object but I have no idea about the name.

docs-website/src/pages/docs/wundergraph-config-ts-reference/configure-openapi-rest-data-source.md
StarpTech: But you can't ensure that your service will respond in exact 500ms. Everything can happen in between. From my experience, timeouts are chosen as generous.

fiam: It depends on what your options are. If a timeout means all you can say is "Oooops, nothing for you to see here, please move along" then you will likely set it generously, because there's nothing else you can show if the upstream takes a long time. When you can have a fallback, you usually want to set it lower to still load your UI in a reasonable time.

packages/sdk/src/definition/graphql-introspection.ts
fiam: I've reverted it

packages/sdk/src/definition/index.ts
fiam: Changed to `requestTimeoutMilliseconds`

docs-website/src/pages/docs/wundergraph-config-ts-reference/configure-openapi-rest-data-source.md
fiam: There’s also a middle ground. We can make it take seconds and accept floats. The precision loss shouldn’t make any difference for this use case.

StarpTech: We specifically talk about request timeout :smile:

StarpTech: I just want to be consistent here. We have already "pollingIntervalSeconds" and I really don't see a reason to switch units.

fiam: I agree that having different units is less than ideal, but I think the mistake is in `pollingIntervalSeconds` being an integer. A second is just not granular enough for a lot of things, that's why most tools/SDKs accept either milliseconds or decimal seconds. If you want us to use `requestTimeoutSeconds` with integers, we can do it, but let it be noted that I would strongly recommend against it.

StarpTech: I don't understand your concern :smiley_cat:  Practically, polling an API isn't something where milliseconds matter. If you want to fetch every x second this is absolutely fine. The same goes for timeouts. There is no value in defining a request timeout with 2600ms instead 3s. It is an estimation of when a request is likely to be faulty.

jensneuse: Seconds for timeouts and polling intervals is fine.

fiam: I changed it to seconds in the last commit

StarpTech: :+1:

docs-website/src/pages/docs/wundergraph-config-ts-reference/configure-graphql-data-source.md
```diff
@@ -14,6 +14,7 @@ The `GraphQL` data source allows you to introspect GraphQL services and add them
 const countries = introspect.graphql({
   apiNamespace: 'countries',
   url: 'https://countries.trevorblades.com/',
+  timeoutMilliseconds: 10 * 1000, // optional
```
jensneuse, line 17: this needs an update

StarpTech, line 17: Should be "requestTimeoutSeconds"

StarpTech (changes requested):

jensneuse (changes requested):
update readme

pkg/engineconfigloader/engineconfigloader.go
```diff
@@ -111,42 +128,56 @@ func (d *DefaultFactoryResolver) tryCreateHTTPSClient(mTLS *wgpb.MTLSConfigurati
...
+func (d *DefaultFactoryResolver) newHTTPClient(ds *wgpb.DataSourceConfiguration, cfg *wgpb.FetchConfiguration) (*http.Client, error) {
+	// Timeout
+	timeout := d.transportFactory.DefaultTransportTimeout()
+	if ds != nil && ds.RequestTimeoutSeconds > 0 {
+		timeout = time.Duration(ds.RequestTimeoutSeconds) * time.Second
+	}
+	// TLS
+	var transport http.RoundTripper
+	var err error
+	if cfg != nil && cfg.MTLS != nil {
```
jensneuse, line 145: seems like config and mtls is checked twice

pkg/engineconfigloader/engineconfigloader.go
```diff
@@ -111,42 +128,56 @@ func (d *DefaultFactoryResolver) tryCreateHTTPSClient(mTLS *wgpb.MTLSConfigurati
...
 	case wgpb.DataSourceKind_GRAPHQL:
-		// when mTLS is enabled, we need to create a new client
-		if ds.CustomGraphql != nil && ds.CustomGraphql.Fetch != nil && ds.CustomGraphql.Fetch.MTLS != nil {
-			client, err := d.tryCreateHTTPSClient(ds.CustomGraphql.Fetch.MTLS)
+		if d.useCustomHTTPClient(ds, ds.CustomGraphql.Fetch) {
+			client, err := d.newHTTPClient(ds, ds.CustomGraphql.Fetch)
 			if err != nil {
 				return nil, err
 			}
+			// XXX: Implement WithHTTPClient?
```
jensneuse, line 167: no

jensneuse (changes requested):

pkg/engineconfigloader/engineconfigloader.go
```diff
@@ -74,8 +78,21 @@ func NewDefaultFactoryResolver(transportFactory ApiTransportFactory, baseTranspo
 	}
 }
 
-// tryCreateHTTPSClient creates a http client with the given options or defaults to the standard client if no mTLS options are given
-func (d *DefaultFactoryResolver) tryCreateHTTPSClient(mTLS *wgpb.MTLSConfiguration) (*http.Client, error) {
+// useCustomHTTPClient returns true iff the given FetchConfiguration requires a dedicated HTTP client
+func (d *DefaultFactoryResolver) useCustomHTTPClient(ds *wgpb.DataSourceConfiguration, cfg *wgpb.FetchConfiguration) bool {
```
StarpTech, line 82: This can be a function and doesn't need to be a method. The prefix `use` doesn't really express a true/false return value.

pkg/engineconfigloader/engineconfigloader.go
```diff
@@ -111,42 +128,56 @@ func (d *DefaultFactoryResolver) tryCreateHTTPSClient(mTLS *wgpb.MTLSConfigurati
...
-
 }
 
 func (d *DefaultFactoryResolver) Resolve(ds *wgpb.DataSourceConfiguration) (plan.PlannerFactory, error) {
 	switch ds.Kind {
 	case wgpb.DataSourceKind_GRAPHQL:
-		// when mTLS is enabled, we need to create a new client
-		if ds.CustomGraphql != nil && ds.CustomGraphql.Fetch != nil && ds.CustomGraphql.Fetch.MTLS != nil {
-			client, err := d.tryCreateHTTPSClient(ds.CustomGraphql.Fetch.MTLS)
+		if d.useCustomHTTPClient(ds, ds.CustomGraphql.Fetch) {
```
StarpTech, line 162: I think we can simplify this:

```go
// create a new client based on config etc...
client, err := d.newHTTPClient(ds, ds.CustomGraphql.Fetch)
if err != nil {
	return nil, err
}
d.graphql.HTTPClient = client
return d.graphql, nil
```

StarpTech (changes requested):

pkg/engineconfigloader/engineconfigloader.go
```diff
@@ -111,42 +128,56 @@ func (d *DefaultFactoryResolver) tryCreateHTTPSClient(mTLS *wgpb.MTLSConfigurati
...
+			client, err := d.newHTTPClient(ds, ds.CustomGraphql.Fetch)
 			if err != nil {
 				return nil, err
 			}
+			// XXX: Implement WithHTTPClient?
 			return &graphql_datasource.Factory{
```
StarpTech, line 168: We forgot to pass all properties from the original factory to this one. My suggestion above will this is.

StarpTech (changes requested):

pkg/engineconfigloader/engineconfigloader.go
fiam: I'd rather keep it as a method for namespacing. I had it named `shouldUseCustomHTTPClient` but then it felt a bit too Objective-C-y/Swift-y. I've changed it to `requiresCustomHTTPClient` which should convey the boolean nature of the function better.

StarpTech: Sounds good. Thank you.

pkg/engineconfigloader/engineconfigloader.go
fiam: I can't spot the other place where it gets checked. Could you please clarify where?

fiam:
> please update readme

I can't find any README files that mention timeouts. Could you please let me know which one specifically would you like me to update? Thanks!

StarpTech:
@fiam How did you test the implementation?

fiam:
> @fiam How did you test the implementation?

I configured sources from the test apps with 1ms timeouts and verified that the failed due to a deadline being exceeded. Then I did more or less the same for the global timeout.

StarpTech:
> I configured sources from the test apps with 1ms timeouts and verified that the failed due to a deadline being exceeded. Then I did more or less the same for the global timeout.

Thanks

packages/protobuf/wgpb/wundernode_config.ts
```diff
@@ -942,6 +943,7 @@ export interface NodeOptions {
 	publicNodeUrl: ConfigurationVariable | undefined;
 	listen: ListenerOptions | undefined;
 	logger: NodeLogging | undefined;
+	defaultRequestTimeoutSeconds: number;
```
StarpTech, line 946: Why is called here differently?

StarpTech (changes requested):

packages/protobuf/wgpb/wundernode_config.ts
fiam: That one is the default for the whole node, for sources that don't have an explicit one.

StarpTech: My point is why don't we call it `requestTimeoutSeconds` as well?

fiam: If you give them both the same name, I feel like it’s not immediately clear which one takes priority. Having one with the default prefix makes it more evident.

StarpTech: Ah got it.

jensneuse (approved):

jensneuse:
amazing work!

fiam:
> amazing work!

Thanks to your and Dustin’s help!

PR #234, opened 2022-10-05
chore: use global logger for sdk

spetrunin:
RFC: global logger for sdk

Before:
Screenshot 2022-10-05 at 12 30 56

After:
Screenshot 2022-10-05 at 12 29 56

- We could enable by default pino-prettier for non production commands
- Having global logger will allow to consume build logs by any tool which supports structured logger


StarpTech:
@spetrunin this PR can be dropped in favor of the other right?

spetrunin:
yeah
closed in favour of https://github.com/wundergraph/wundergraph/pull/240

PR #237, opened 2022-10-05
chore: add linux/arm64 support

fiam:
- add support for linux/arm64 in binarypaths.js, so it downloads the correct binary
- add linux/arm64 wunderctl binaries to releases

#### Checklist

- [x] run `make test`


.github/goreleaser/.goreleaser-linux.yml
```diff
@@ -12,7 +12,7 @@ before:
     - go generate ../...
 builds:
   - binary: wunderctl
-    id: linux
+    id: linux_amd64
```
StarpTech, line 15: We used it in some other places. We should be consistent.

StarpTech (changes requested):

.github/goreleaser/.goreleaser-linux.yml
fiam: I'm not sure what do you mean by that. Do you mean there are other places than use `x86_64` instead of `amd64`?

jensneuse: we should use the same id across all places is what @StarpTech means I think
this id has no real meaning, it's just about making it consistent

StarpTech: Yes

jensneuse (approved):

StarpTech (approved):
LGTM

PR #238, opened 2022-10-05
feat: refactor cache handling, move storage out of generated

fiam:
- feat: move introspection cache to .wundergraph/introspection/cache
- feat: prioritize cache during introspection

#### Checklist

- [x] run `make test`

### DIFF
--- a/cli/commands/generate.go
+++ b/cli/commands/generate.go
@@ -19,6 +19,7 @@ import (
 
 var (
 	generateAndPublish bool
+	cacheFallback      bool
 )
 
 // generateCmd represents the generate command
@@ -57,6 +58,8 @@ Use this command if you only want to generate the configuration`,
 				// Run scripts in prod mode
 				"NODE_ENV=production",
 				fmt.Sprintf("WUNDERGRAPH_PUBLISH_API=%t", generateAndPublish),
+				fmt.Sprintf("WG_ENABLE_INTROSPECTION_CACHE=%t", !disableCache),
+				fmt.Sprintf("WG_USE_INTROSPECTION_CACHE_EXCLUSIVELY=%t", !cacheFallback),
 				fmt.Sprintf("WG_DIR_ABS=%s", wunderGraphDir),
 			),
 		})
@@ -172,5 +175,6 @@ Use this command if you only want to generate the configuration`,
 
 func init() {
 	generateCmd.Flags().BoolVarP(&generateAndPublish, "publish", "p", false, "publish the generated API immediately")
+	generateCmd.Flags().BoolVar(&cacheFallback, "cache-fallback", false, "fallback to introspection if cache fails")
 	rootCmd.AddCommand(generateCmd)
 }
--- a/cli/commands/generate.go
+++ b/cli/commands/generate.go
@@ -19,7 +19,7 @@ import (
 
 var (
 	generateAndPublish bool
-	cacheFallback      bool
+	offline            bool
 )
 
 // generateCmd represents the generate command
@@ -59,7 +59,7 @@ Use this command if you only want to generate the configuration`,
 				"NODE_ENV=production",
 				fmt.Sprintf("WUNDERGRAPH_PUBLISH_API=%t", generateAndPublish),
 				fmt.Sprintf("WG_ENABLE_INTROSPECTION_CACHE=%t", !disableCache),
-				fmt.Sprintf("WG_USE_INTROSPECTION_CACHE_EXCLUSIVELY=%t", !cacheFallback),
+				fmt.Sprintf("WG_ENABLE_INTROSPECTION_OFFLINE=%t", offline),
 				fmt.Sprintf("WG_DIR_ABS=%s", wunderGraphDir),
 			),
 		})
@@ -175,6 +175,6 @@ Use this command if you only want to generate the configuration`,
 
 func init() {
 	generateCmd.Flags().BoolVarP(&generateAndPublish, "publish", "p", false, "publish the generated API immediately")
-	generateCmd.Flags().BoolVar(&cacheFallback, "cache-fallback", false, "fallback to introspection if cache fails")
+	generateCmd.Flags().BoolVar(&offline, "offline", false, "offline mode, return an error instead of hitting the network")
 	rootCmd.AddCommand(generateCmd)
 }
--- a/cli/commands/generate.go
+++ b/cli/commands/generate.go
@@ -178,6 +178,6 @@ var generateCmd = &cobra.Command{
 
 func init() {
 	generateCmd.Flags().BoolVarP(&generateAndPublish, "publish", "p", false, "publish the generated API immediately")
-	generateCmd.Flags().BoolVar(&offline, "offline", false, "offline mode, return an error instead of hitting the network")
+	generateCmd.Flags().BoolVar(&offline, "offline", false, "disables loading resources from the network")
 	rootCmd.AddCommand(generateCmd)
 }
--- a/cli/commands/generate.go
+++ b/cli/commands/generate.go
@@ -25,9 +25,12 @@ var (
 // generateCmd represents the generate command
 var generateCmd = &cobra.Command{
 	Use:   "generate",
-	Short: "Runs the code generation process once",
-	Long: `In contrast to wunderctl up, it only generates all files in ./generated but doesn't start WunderGraph or the hooks.
-Use this command if you only want to generate the configuration`,
+	Short: "Generate the production config",
+	Long: `Generate the production config to start the node and hook component
+ with 'wunderctl start' or individually with 'wunderctl node start', 'wunderctl
+ server start'. All files are stored to .wundergraph/generated. The local
+ introspection cache has precedence. You can overwrite this behavior by passing
+ --no-cache to the command`,
 	RunE: func(cmd *cobra.Command, args []string) error {
 		wunderGraphDir, err := files.FindWunderGraphDir(_wunderGraphDirConfig)
 		if err != nil {
--- a/cli/commands/generate.go
+++ b/cli/commands/generate.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"os"
 	"path"
-	"time"
 
 	"github.com/jensneuse/abstractlogger"
 	"github.com/spf13/cobra"
@@ -45,8 +44,7 @@ var generateCmd = &cobra.Command{
 		// optional, no error check
 		codeServerFilePath, _ := files.CodeFilePath(wunderGraphDir, serverEntryPointFilename)
 
-		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
-		defer cancel()
+		ctx := context.Background()
 
 		configOutFile := path.Join("generated", "bundle", "config.js")
 
--- a/cli/commands/root.go
+++ b/cli/commands/root.go
@@ -145,7 +145,7 @@ func init() {
 	rootCmd.PersistentFlags().BoolVar(&enableDebugMode, "debug", false, "enables the debug mode so that all requests and responses will be logged")
 	rootCmd.PersistentFlags().BoolVar(&jsonEncodedLogging, "json-encoded-logging", false, "switches the logging to json encoded logging")
 	rootCmd.PersistentFlags().StringVar(&_wunderGraphDirConfig, "wundergraph-dir", files.WunderGraphDirName, "path to your .wundergraph directory")
-	rootCmd.PersistentFlags().BoolVar(&disableCache, "no-cache", false, "Disable cache")
+	rootCmd.PersistentFlags().BoolVar(&disableCache, "no-cache", false, "disables local caches")
 }
 
 func buildLogger(level abstractlogger.Level) abstractlogger.Logger {
--- a/cli/commands/root.go
+++ b/cli/commands/root.go
@@ -145,7 +145,7 @@ func init() {
 	rootCmd.PersistentFlags().BoolVar(&enableDebugMode, "debug", false, "enables the debug mode so that all requests and responses will be logged")
 	rootCmd.PersistentFlags().BoolVar(&jsonEncodedLogging, "json-encoded-logging", false, "switches the logging to json encoded logging")
 	rootCmd.PersistentFlags().StringVar(&_wunderGraphDirConfig, "wundergraph-dir", files.WunderGraphDirName, "path to your .wundergraph directory")
-	rootCmd.PersistentFlags().BoolVar(&disableCache, "no-cache", false, "Enable cache")
+	rootCmd.PersistentFlags().BoolVar(&disableCache, "no-cache", false, "Disable cache")
 }
 
 func buildLogger(level abstractlogger.Level) abstractlogger.Logger {
--- a/cli/commands/root.go
+++ b/cli/commands/root.go
@@ -1,10 +1,12 @@
 package commands
 
 import (
+	"errors"
 	"fmt"
 	"io/fs"
 	"net/http"
 	"os"
+	"path/filepath"
 	"time"
 
 	"github.com/fatih/color"
@@ -39,6 +41,7 @@ var (
 	serviceToken          string
 	_wunderGraphDirConfig string
 	disableCache          bool
+	clearCache            bool
 
 	red    = color.New(color.FgHiRed)
 	green  = color.New(color.FgHiGreen)
@@ -71,16 +74,27 @@ You can opt out of this by setting the following environment variable: WUNDERGRA
 		if err != nil {
 			if _, ok := err.(*fs.PathError); ok {
 				log.Debug("starting without env file")
-				return nil
+			} else {
+				log.Fatal("error loading env file",
+					abstractlogger.Error(err))
 			}
-			log.Fatal("error loading env file",
-				abstractlogger.Error(err),
-			)
 		} else {
 			log.Debug("env file successfully loaded",
 				abstractlogger.String("file", DotEnvFile),
 			)
 		}
+
+		if clearCache {
+			wunderGraphDir, err := files.FindWunderGraphDir(_wunderGraphDirConfig)
+			if err != nil {
+				return err
+			}
+			cacheDir := filepath.Join(wunderGraphDir, "cache")
+			if err := os.RemoveAll(cacheDir); err != nil && !errors.Is(err, os.ErrNotExist) {
+				return err
+			}
+		}
+
 		return nil
 	},
 	RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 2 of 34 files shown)


docs-website/src/pages/docs/wundergraph-config-ts-reference/configure-introspection.md
```diff
@@ -25,7 +25,7 @@ That's why we've introduced introspection caching.
 
 {% callout type="warning" %}
 By default, every data source is introspected exactly once.
-The result is cached in the `.wundergraph/generated/introspection` directory.
+The result is cached in the `.wundergraph/introspection/cache` directory.
```
StarpTech, line 28: When we introduce a global persistent cache dir, we should not call it "introspection" maybe `.wundergraph/cache` and put the introspection cache in `.wundergraph/cache/introspection`. In the future, we probably have other things to cache.

packages/sdk/src/definition/introspection-cache.ts
```diff
@@ -122,48 +122,32 @@ export const introspectWithCache = async <Introspection extends IntrospectionCon
...
-		}
+	/*
+	 * At this point, we don't have a cache entry for this. Generate
+	 * introspection from scratch. Then cache it
+	 */
+	const api = await generator(introspection);
 
-		return api;
-	} catch (e) {
-		// fallback to old introspection result (only for development mode)
```
StarpTech, line 156: Why did you remove this? You have to differentiate between 

1. I run `wunderctl up`
2. I run `wunderctl generate`

Both commands should use the introspection cache but  `wunderctl up` calls it periodically. In case of an outage of one of the data sources, we want to fall back on the recent valid result. How do we ensure this now?

packages/sdk/src/definition/introspection-cache.ts
```diff
@@ -98,8 +97,9 @@ export const introspectWithCache = async <Introspection extends IntrospectionCon
 	introspection: Introspection,
 	generator: (introspection: Introspection) => Promise<Api<A>>
 ): Promise<Api<A>> => {
-	const isIntrospectionCacheEnabled =
-		WG_ENABLE_INTROSPECTION_CACHE && introspection.introspection?.disableCache !== true;
+	const isIntrospectionDisabledBySource = introspection.introspection?.disableCache === true;
+	const isIntrospectionDisabledByEnv = WG_DISABLE_INTROSPECTION_CACHE;
+	const isIntrospectionCacheEnabled = !isIntrospectionDisabledBySource && !isIntrospectionDisabledByEnv;
 	const cacheKey = objectHash(introspection);
```
StarpTech, line 103: I'd compute only when caching was enabled.

packages/sdk/src/definition/introspection-cache.ts
```diff
@@ -122,48 +122,32 @@ export const introspectWithCache = async <Introspection extends IntrospectionCon
...
-		}
-		throw e;
+	/*
+	 * We got a result. If the cache is enabled, populate it
+	 */
+	if (isIntrospectionCacheEnabled) {
+		try {
+			const cacheEntry = toCacheEntry<A>(api);
+			await writeIntrospectionCacheFile(cacheKey, JSON.stringify(cacheEntry));
+		} catch (e) {}
```
StarpTech, line 149: Never swallow without at least a log entry :smile:

cli/commands/up.go
```diff
@@ -88,18 +88,7 @@ var upCmd = &cobra.Command{
 			AbsWorkingDir: WunderGraphDir,
 			ScriptArgs:    []string{configOutFile},
 			Logger:        log,
-			FirstRunEnv: append(os.Environ(),
```
StarpTech, line 91: Do I understand correctly that the introspection cache is now enabled by default?

StarpTech (changes requested):

packages/sdk/src/definition/introspection-cache.ts
fiam: That block was just moved around, so I didn't question it that much, but you're right. I will add a log entry.

packages/sdk/src/definition/introspection-cache.ts
fiam: `cacheKey` is also used when the source requires polling, even if the cache is disabled. We can initialise it lazily, but then we run into potential incorrect uses without initialisation. It think the hashing is cheap enough to perform it on function entry.

docs-website/src/pages/docs/wundergraph-config-ts-reference/configure-introspection.md
fiam: I do agree that having `/cache/` at the first level is better than the at the last, but the initial issue specified exactly `.wundergraph/introspection/cache`. Should we change it or keep as initially requested?

fiam:
Made the requested changed and rebased

cli/commands/up.go
StarpTech: No, sounds good.

docs-website/src/pages/docs/wundergraph-config-ts-reference/configure-introspection.md
StarpTech: We should change it.

packages/sdk/src/definition/introspection-cache.ts
StarpTech: I mean, in case there is a cached introspection result, and the origin fails, the user should be able to continue with the cached result. Currently, the process will exit.

packages/sdk/src/definition/introspection-cache.ts
StarpTech: ok

packages/sdk/src/definition/introspection-cache.ts
fiam: By the time execution reaches that point, either the cache doesn't have the entry or either it was explicitly disabled. Shouldn't we fail and exit?

StarpTech (changes requested):

cli/commands/root.go
```diff
@@ -69,6 +70,12 @@ You can opt out of this by setting the following environment variable: WUNDERGRA
 			log = buildLogger(findLogLevel(abstractlogger.ErrorLevel))
 		}
 
+		if enableCache {
+			if err := os.Setenv("WG_ENABLE_INTROSPECTION_CACHE", "true"); err != nil {
```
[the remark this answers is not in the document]
fiam: Passing it individually to each command seems very error prone, at least a this stage. Unless we're expecting some commands will want to override the `--cache` flag in the near future, I think handling it globally is better.

cli/commands/generate.go
```diff
@@ -57,6 +58,8 @@ Use this command if you only want to generate the configuration`,
 				// Run scripts in prod mode
 				"NODE_ENV=production",
 				fmt.Sprintf("WUNDERGRAPH_PUBLISH_API=%t", generateAndPublish),
+				fmt.Sprintf("WG_ENABLE_INTROSPECTION_CACHE=%t", !disableCache),
+				fmt.Sprintf("WG_USE_INTROSPECTION_CACHE_EXCLUSIVELY=%t", !cacheFallback),
```
StarpTech, line 62: Can we simplify the flag? :smile: What means EXCLUSIVELY?

StarpTech (changes requested):

cli/commands/root.go
```diff
@@ -147,6 +148,7 @@ func init() {
 	rootCmd.PersistentFlags().BoolVar(&enableDebugMode, "debug", false, "enables the debug mode so that all requests and responses will be logged")
 	rootCmd.PersistentFlags().BoolVar(&jsonEncodedLogging, "json-encoded-logging", false, "switches the logging to json encoded logging")
 	rootCmd.PersistentFlags().StringVar(&_wunderGraphDirConfig, "wundergraph-dir", files.WunderGraphDirName, "path to your .wundergraph directory")
+	rootCmd.PersistentFlags().BoolVar(&disableCache, "no-cache", false, "Enable cache")
```
StarpTech, line 151: You mean "disable cache" ?

StarpTech (changes requested):

cli/commands/generate.go
StarpTech: `WG_ENABLE_INTROSPECTION_CACHE_FALLBACK` ?

cli/commands/up.go
```diff
@@ -62,19 +65,21 @@ var upCmd = &cobra.Command{
...
+		if !keepIntrospectionCache {
+			if err := os.RemoveAll(introspectionCacheDir); err != nil && !errors.Is(err, os.ErrNotExist) {
+				return err
 			}
 		}
-		err = os.MkdirAll(introspectionCacheDir, os.ModePerm)
-		if err != nil {
-			return err
+
+		if introspectionCacheClearInterval > 0 {
```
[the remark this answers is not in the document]
jensneuse: The typescript sdk automatically polls (if configured properly) and refreshes the cache. No need to clear it from the go side.

cli/commands/up.go
jensneuse: Is the switch from FirstRunEnv to ScriptEnv intentional?

cli/commands/up.go
```diff
@@ -300,5 +296,6 @@ var upCmd = &cobra.Command{
 
 func init() {
 	rootCmd.AddCommand(upCmd)
-	upCmd.Flags().BoolVar(&clearIntrospectionCache, "clear-introspection-cache", false, "clears the introspection cache")
+	upCmd.Flags().BoolVar(&keepIntrospectionCache, "keep-introspection-cache", false, "avoid clearing the introspection cache on startup")
```
jensneuse, line 299: this wasn't in the latest proposal we accepted

cli/commands/up.go
```diff
@@ -300,5 +296,6 @@ var upCmd = &cobra.Command{
...
+	upCmd.Flags().IntVar(&introspectionCacheClearInterval, "cache-clear-interval", 5, "interval to clear the instrospection cache periodically, in seconds - 0 to disable")
```
jensneuse, line 300: clearing the cache from go does not make sense this way

jensneuse (changes requested):

cli/commands/root.go
fiam: Indeed! Thanks for the heads up. Fixed.

cli/commands/generate.go
fiam: It's because "fallback" feels a bit weird as a name for this. When "cache" and "fallback" go together usually you hit the source of truth first (e.g. the remote server), then when that fails you query the cache. I'll see if I can come up with a better name.

fiam: Thinking more about this... Collins Dictionary says: "Exclusively is used to [refer](https://www.collinsdictionary.com/dictionary/english/refer) to situations or activities that involve only the thing or things [mentioned](https://www.collinsdictionary.com/dictionary/english/mention), and nothing else.", so maybe it's not that bad of a name.

cli/commands/root.go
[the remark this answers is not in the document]
fiam: I've changed it to pass the variable individually to each command.

cli/commands/up.go
StarpTech: I'd not add it unless requested.

jensneuse: I would also suggest to just always clear on "up"

fiam: Deleting code? I can get behind that!

fiam:
> One thing I'd add is to set all "introspection" configs to automatically refresh each cache item every 5 seconds if not otherwise configured. So, by default, all introspections should be cached and refreshed after 5 seconds of waiting. This will give the most pleasent DX as it will be fast and (almost) always up to date if a datasource changes. @StarpTech opinion?

So basically making `pollingIntervalSeconds` default to 5 while running from `wunderctl up`?

jensneuse:
> > One thing I'd add is to set all "introspection" configs to automatically refresh each cache item every 5 seconds if not otherwise configured. So, by default, all introspections should be cached and refreshed after 5 seconds of waiting. This will give the most pleasent DX as it will be fast and (almost) always up to date if a datasource changes. @StarpTech opinion?
> 
> So basically making `pollingIntervalSeconds` default to 5 while running from `wunderctl up`?

It can always default to 5s, not just for "up", because when using "start", we don't introspect.

StarpTech:
Assumption: Most people integrate an existing API into the service. They don't need to refresh it so often. If so, they can configure it. For me it makes more sense to opt-in.

jensneuse:
> Assumption: Most people integrate an ready API into the service. They don't need to refresh it so often. If so, they can configure it. For me, it makes more sense to opt in.

ok, makes sense

jensneuse (approved):

fiam:
@StarpTech GH says there's one remaining change requested from you, but then it links to a resolved one. Can we dismiss it?

jensneuse:
> @StarpTech GH says there's one remaining change requested from you, but then it links to a resolved one. Can we dismiss it?

@StarpTech if you're ok, please approve

cli/commands/generate.go
```diff
@@ -172,5 +175,6 @@ Use this command if you only want to generate the configuration`,
 
 func init() {
 	generateCmd.Flags().BoolVarP(&generateAndPublish, "publish", "p", false, "publish the generated API immediately")
+	generateCmd.Flags().BoolVar(&cacheFallback, "cache-fallback", false, "fallback to introspection if cache fails")
```
StarpTech, line 178: You mean "Fallback to introspection cache on error"?

StarpTech (changes requested):

StarpTech:
I'd like to test it once locally.

cli/commands/generate.go
fiam: No, it's the other way around. `generate` by default uses only the cache, if the entry is not in cache it will return an error. The option allows querying the network when the entry is not in cache.

StarpTech: ok but still makes no sense to me. "cache fallback" means I use cache but you said it will skip the cache.
Maybe

```go
generateCmd.Flags().BoolVar(&cacheFallback, "force-introspection", false, "Refresh introspection if cache fails")
```

jensneuse: > 

Sounds good, I'd extend it with:

3. `wunderctl generate --no-cache` will always make introspection call

This way, the user can decide between auto (default), online, or offline.

@StarpTech WDYT?

StarpTech: I like.

jensneuse: @fiam I think we have consensus if you agree on the 3 rules.

fiam: That's exactly how most things that work with a cache and provide an offline switch works. Both on by default, provide flags to turn them off individually. If anything, I'm happier with this approach that the previous ones.

jensneuse:
Once the implementation is done, the docs need updates as well:
https://docs.wundergraph.com/docs/wundergraph-config-ts-reference/configure-introspection

cli/commands/generate.go
```diff
@@ -175,6 +175,6 @@ Use this command if you only want to generate the configuration`,
 
 func init() {
 	generateCmd.Flags().BoolVarP(&generateAndPublish, "publish", "p", false, "publish the generated API immediately")
-	generateCmd.Flags().BoolVar(&cacheFallback, "cache-fallback", false, "fallback to introspection if cache fails")
+	generateCmd.Flags().BoolVar(&offline, "offline", false, "offline mode, return an error instead of hitting the network")
```
StarpTech, line 178: I always try to understand this from the user's perspective. You didn't explain it. I'd try to simplify this:

> Offline mode enforces the use of the local introspection cache

StarpTech (changes requested):

cli/commands/root.go
```diff
@@ -144,6 +145,7 @@ func init() {
 	rootCmd.PersistentFlags().BoolVar(&enableDebugMode, "debug", false, "enables the debug mode so that all requests and responses will be logged")
 	rootCmd.PersistentFlags().BoolVar(&jsonEncodedLogging, "json-encoded-logging", false, "switches the logging to json encoded logging")
 	rootCmd.PersistentFlags().StringVar(&_wunderGraphDirConfig, "wundergraph-dir", files.WunderGraphDirName, "path to your .wundergraph directory")
+	rootCmd.PersistentFlags().BoolVar(&disableCache, "no-cache", false, "Disable cache")
```
StarpTech, line 148: Here as well

> Disables the local introspection cache.

StarpTech (changes requested):

cli/commands/generate.go
```diff
@@ -19,6 +19,7 @@ import (
 
 var (
 	generateAndPublish bool
+	offline            bool
 )
 
 // generateCmd represents the generate command
```
StarpTech, line 25: Could you also update the cmd text?

	Short: "Generate the production config.",
    Long: "Generate the production config to start the node and hook component. All files are stored to .wundergraph/generated. The local introspection cache has precedence. You can overwrite this behavior by passing --no-cache to the command"

StarpTech (changes requested):

cli/commands/generate.go
fiam: It probably needs a better help string, but I don't think we should mention anything regarding the cache in there, they're orthogonal concepts. The way that the failure is handled by the code managing the cache is an implementation detail.

What about "disables loading resources from the network"?

cli/commands/root.go
fiam: Right now we only have a cache for introspection, but I think not far from now we'll have caches for other things. It's rare to need to disable just some caches - most of the time you either enable them all (for speed) or disable them altogether (to e.g. ensure the changes are applied, refetching resources or ensuring a cold start works). Think of web browsers and how virtually none of them expose the different caches individually, they just allow clearing/disabling everything at once.

I would, however, change the help to 3rd person since I've been checking all available flags in wunderctl and while they're not 100% consistent, 3rd person seems more common than 1st. What about "disables local caches"?

cli/commands/generate.go
fiam: The long description seems a bit too verbose to me, mentioning where the cache is stored as well as to how to disable it (which can be seen via --help), but I guess it's fine.

cli/commands/root.go
StarpTech: I'm okay with it, but I thought it wouldn't hurt to be a bit more clear about what's the impact of this flag.

cli/commands/generate.go
StarpTech: Fine for me as well.

StarpTech:
Let me test it locally and then we are good to go.

docs-website/src/pages/docs/wundergraph-config-ts-reference/configure-introspection.md
```diff
@@ -25,15 +25,17 @@ That's why we've introduced introspection caching.
...
+The result is cached in the `.wundergraph/cache/introspection` directory.
 {% /callout %}
 
-If a cache entry is found for a data source,
-it's always used instead of introspecting the data source again.
+When caching is enabled, if a cache entry is found for a data source,
+it's always used instead of introspecting the data source again. Cache is
+optional and might be turned off globally using the `--no-cache` flag.
 
 ## Clearing the Cache
```
StarpTech, line 35: Please add some docs about the different flags and use cases.
By default, we recommend not committing the cache directory unless you can't make introspection calls in your CI pipeline.

StarpTech (changes requested):

docs-website/src/pages/docs/wundergraph-config-ts-reference/configure-introspection.md
fiam: I've added some documentation about it, mentioning that it can be useful for CI. Please, let me know what you think. Thanks!

PR #239, opened 2022-10-05
refactor: start node and server in the same way as with separate comm…

StarpTech:
…ands

#### Checklist

- [ ] run `make test`


cli/commands/start.go
```diff
@@ -82,28 +80,39 @@ If used without --exclude-server, make sure the server is available in this dire
 			}()
 		}
 
-		configFileChangeChan := make(chan struct{})
-		n := node.New(nodeCtx, BuildInfo, WunderGraphDir, log)
+		data, err := ioutil.ReadFile(configFile)
+		if err != nil {
+			log.Fatal("Failed to read file", abstractlogger.String("filePath", configFile), abstractlogger.Error(err))
```
YuriBuerov, line 85: fatal calls `os.Exit`

StarpTech: I just copied it over. Good point will fix it in the other places too.

cli/commands/start.go
```diff
@@ -82,28 +80,39 @@ If used without --exclude-server, make sure the server is available in this dire
...
 				log.Fatal("startBlocking", abstractlogger.Error(err))
 			}
 		}()
 
-		// trigger server reload after initial config build
-		// because no fs event is fired as build is already done
-		configFileChangeChan <- struct{}{}
-
-		n.HandleGracefulShutdown(gracefulTimeout)
+		gracefulTimeoutSeconds := gracefulTimeout
```
YuriBuerov, line 114: why do you need to allocate a new variable?

cli/commands/start.go
```diff
@@ -82,28 +80,39 @@ If used without --exclude-server, make sure the server is available in this dire
...
-				node.WithFileSystemConfig(configFile),
-				node.WithDebugMode(enableDebugMode),
+				node.WithStaticWunderNodeConfig(wunderNodeConfig),
 				node.WithForceHttpsRedirects(!disableForceHttpsRedirects),
 				node.WithIntrospection(enableIntrospection),
-				node.WithGitHubAuthDemo(GitHubAuthDemo),
+				node.WithDebugMode(enableDebugMode),
 			)
 			if err != nil {
 				log.Fatal("startBlocking", abstractlogger.Error(err))
```
StarpTech, line 110: This will kill the process instead graceful shutdown

cli/commands/start.go
StarpTech: done

cli/commands/start.go
```diff
@@ -26,97 +20,57 @@ var (
...
-					log.Error("Stopping hooks-server-runner failed",
-						abstractlogger.String("runnerName", "hooks-server-runner"),
-						abstractlogger.Error(err),
-					)
-				}
-			}()
+		var (
+			errChan    = make(chan error)
+			doneChan   = make(chan struct{})
+			shutdownWg sync.WaitGroup
```
StarpTech, line 32: You can use https://pkg.go.dev/golang.org/x/sync/errgroup to simplify it.

StarpTech:
@spetrunin great improvement :smile:

cli/commands/server.go
```diff
@@ -65,11 +65,12 @@ func startWunderGraphServer(ctx context.Context) error {
 	hookServerRunner := helpers.NewServerRunner(log, srvCfg)
 
 	<-hookServerRunner.Run(ctx)
-	if ctx.Err() != nil {
-		log.Info("WunderGraph Server shutdown complete")
-	} else {
+
+	if ctx.Err() == nil {
 		log.Error("WunderGraph Server excited")
+		return fmt.Errorf("WunderGraph Server exited")
```
StarpTech, line 71: Why return an error when context err is nil?

spetrunin: it works opposite:
if context was cancelled by timeout or signal - there will be error - this means we shutdown gracefully via context

But if context is ok - there will be no error - which means script runner exited on its own

spetrunin (approved):
👍

cli/commands/server.go
StarpTech: Isn't that a synchronization issue that can be solved?

spetrunin: it could happen when wundergraph server throws an error and exits

StarpTech: Do you mean panic? which is very unliekly.

spetrunin: couldn't resolve variable or maybe placeholder somewhere

any situation when we call exit 
https://github.com/wundergraph/wundergraph/blob/main/packages/sdk/src/middleware/server.ts#L121

StarpTech: But we get this event. We only need to change the way how we communicate it.

spetrunin: open to any suggestions how to get from script runner

but it already will be in logs from server

cli/commands/node.go
```diff
@@ -31,55 +32,63 @@ var nodeStartCmd = &cobra.Command{
...
+	err = json.Unmarshal(data, &graphConfig)
+	if err != nil {
+		log.Error("Failed to unmarshal", abstractlogger.String("filePath", configFile), abstractlogger.Error(err))
+		return errors.New("failed to unmarshal config file")
+	}
+
+	wunderNodeConfig := node.CreateConfig(&graphConfig)
+	n := node.New(ctx, BuildInfo, wunderGraphDir, log)
+
+	errChan := make(chan error, 1)
```
StarpTech, line 77: Even here, I'd use errgroup. Less code at the end.

cli/commands/server.go
StarpTech: Let me check

pkg/node/node.go
```diff
@@ -154,6 +155,8 @@ func WithForceHttpsRedirects(forceHttpsRedirects bool) Option {
 }
 
 func (n *Node) StartBlocking(opts ...Option) error {
+	defer close(n.startExited)
```
StarpTech, line 158: I run into a panic

```
panic: send on closed channel

goroutine 114 [running]:
github.com/wundergraph/wundergraph/pkg/node.(*Node).StartBlocking.func1()
        /home/starptech/p/wundergraph/wundergraph/pkg/node/node.go:173 +0x7f
created by github.com/wundergraph/wundergraph/pkg/node.(*Node).StartBlocking
        /home/starptech/p/wundergraph/wundergraph/pkg/node/node.go:170 +0x2f7

```

spetrunin: could u give me reproduce steps?

spetrunin: it is not about this channel, we never send to it
it is about err channel

it means that error occured on a shutdown
I will check when it could happen

StarpTech:
@spetrunin please validate my latest changes

PR #240, opened 2022-10-05
feat: align logging format hooks server and

spetrunin:
**Pretty**

Screenshot 2022-10-06 at 00 50 05

**Json**
Screenshot 2022-10-06 at 00 50 47

**Open question:**
- zap produces time format with floating point, nevertheless, it should be int64
- for zap pretty mode there is no way to format custom props away from JSON at the end of log message

**TODO:**
- WunderNode should favor options.logger.level - zap instance should be shared, abstract logger should be separate for the node and cli


jensneuse:
results look nice

StarpTech:
You can introduce a custom time encoder https://blog.sandipb.net/2018/05/03/using-zap-creating-custom-encoders/

StarpTech:
Pretty logging does not share the same format and style.

- In node, we log as a combination of text + JSON
- In node, we don't work with colors.

How did you ensure that we use the same log level between both components?

As @jensneuse suggest it would be nice if we have a different log prefix for each component

> { "component": "server" }

> INFO: SERVER: ...

StarpTech (changes requested):

cli/commands/root.go
```diff
@@ -63,6 +63,11 @@ wunderctl is gathering anonymous usage data so that we can better understand how
 You can opt out of this by setting the following environment variable: WUNDERGRAPH_DISABLE_METRICS
 `,
 	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+		switch cmd.Name() {
+		case "loadoperations":
+			// skip any setup for loadoperations to avoid logging anything
```
StarpTech, line 68: I don't get it. Could you explain why we need to do this?

StarpTech (changes requested):

packages/sdk/src/env/env.ts
```diff
@@ -0,0 +1,17 @@
...
+	const varName = wgEnvironmentVariableToJSON(v);
+
+	if (varName === 'UNKNOWN') {
+		throw new Error(`Unknown environment variable name: ${v}`);
+	}
+
+	return varName;
+};
+
+export const WgEnvValue = (v: WgEnvironmentVariable): string | undefined => {
```
StarpTech, line 13: I suggest calling it `resolveWgEnv`

packages/sdk/src/env/env.ts
```diff
@@ -0,0 +1,17 @@
+import { WgEnvironmentVariable, wgEnvironmentVariableToJSON } from '@wundergraph/protobuf';
+
+export const WgEnvVariableName = (v: WgEnvironmentVariable): string => {
+	const varName = wgEnvironmentVariableToJSON(v);
+
+	if (varName === 'UNKNOWN') {
```
StarpTech, line 6: Weird that this is not type safe :smile:

types/protos/wundernode_config.proto
```diff
@@ -532,3 +532,8 @@ enum ConfigurationVariableKind {
 	ENV_CONFIGURATION_VARIABLE = 1;
 	PLACEHOLDER_CONFIGURATION_VARIABLE = 2;
 }
+
+enum WgEnvironmentVariable {
```
StarpTech, line 536: Can we introduce the new way of handling env in a separate PR? Otherwise, we will have two ways until someone has refactored it.

cli/commands/root.go
```diff
@@ -145,34 +152,13 @@ func init() {
 	viper.SetDefault("OAUTH_BASE_URL", "https://accounts.wundergraph.com/auth/realms/master")
 	viper.SetDefault("API_URL", "https://api.wundergraph.com")
 
-	rootCmd.PersistentFlags().StringVarP(&logLevel, "loglevel", "l", "info", "sets the log level")
+	rootCmd.PersistentFlags().StringVarP(&cliLogLevel, "loglevel", "l", "info", "sets the CLI log level")
 	rootCmd.PersistentFlags().StringVarP(&DotEnvFile, "env", "e", ".env", "allows you to set environment variables from an env file")
 	rootCmd.PersistentFlags().BoolVar(&enableDebugMode, "debug", false, "enables the debug mode so that all requests and responses will be logged")
 	rootCmd.PersistentFlags().BoolVar(&jsonEncodedLogging, "json-encoded-logging", false, "switches the logging to json encoded logging")
```
StarpTech, line 158: As with many other tools, JSON logging is always the default, and you have to set e.g. `--pretty-logger` to enable pretty print. I favor this because by default we should configure everything for production except for "wunderctl up".

StarpTech (changes requested):

cli/commands/root.go
spetrunin: currently, loadoperations returns data through stdout via fmt.Println
in order to not break it, we should not log anything

cli/commands/root.go
spetrunin: yeah, I agree with that

But before this it was opposite - so it will be a breaking change to a cli - which should be ok I belive

types/protos/wundernode_config.proto
spetrunin: sure, I will extract this into separate PR - so someone could also pick up it

packages/sdk/src/env/env.ts
spetrunin: ok

spetrunin:
> Pretty logging does not share the same format and style.

Yeah, with the current customization options of Pino and Zap
It is not possible to fully align this

> * In node, we log as a combination of text + JSON

that is one of the problems, as zap does not support logging additional keys in a different way at least with a standard console encoder
I believe we could solve it - by writing a custom log encoder - but this was out of the scope of this task

> * In node, we don't work with colors.

Partially do, the log level is colored
We could colorize output also, but I haven't added that

> How did you ensure that we use the same log level between both components?

We don't have to, it is the responsibility of the user to specify log level of WunderNode and WunderGraph Server via the options

But another breaking change - is that loglevel flag of CLI - should be responsible only for CLI log level, neither node nor server, as they use their own config


> As @jensneuse suggest it would be nice if we have a different log prefix for each component
> > { "component": "server" }

Yeah, agree, this will involve more work, I propose extracting it into a separate issue as well as environment variables refactoring

cli/commands/root.go
StarpTech: It's fine :+1:

cli/commands/root.go
jensneuse: Wouldn't it be better if we migrate loadoperation to output to a file?
Otherwise, as Dustin explained, add an explicit comment that we MUST NOT do any logging to not break this

spetrunin: +1 to migrate to file

StarpTech: or using "PersistentPreRunE" only for specific commands. I like both approaches.

spetrunin: added a comment and todo, better to do optimization as a separate issue

jensneuse (approved):

jensneuse (approved):

jensneuse:
LGTM

StarpTech:
What's the status here? The PR has been open for too long.

cli/helpers/flags.go
```diff
@@ -0,0 +1,7 @@
+package helpers
+
+type RootFlags struct {
```
StarpTech, line 3: Why did you put this here instead of in the root command?

packages/golang-client/src/index.ts
```diff
@@ -53,7 +53,7 @@ const gofmt = (code: string) => {
 		// on purpose. Otherwise we throw an error with both the code and
 		// the error message returned by gofmt.
 		if (e instanceof Error && e.message.indexOf('ENOENT') >= 0) {
-			console.error('gofmt is not installed. If you want to prettify the generated code, please install gofmt');
+			console.log('gofmt is not installed. If you want to prettify the generated code, please install gofmt');
```
StarpTech, line 56: This should use the same logger as in the SDK, right? You can expose it in the SDK and use it here too.

packages/sdk/src/codegen/index.ts
```diff
@@ -1,6 +1,7 @@
 import { ResolvedWunderGraphConfig } from '../configure';
 import path from 'path';
 import * as fs from 'fs';
+import { SdkLogger } from '../logger/logger';
```
StarpTech, line 4: Just naming but why did you prefix it with SDK? We are already in the SDK :smile:

packages/sdk/src/configure/options.ts
```diff
@@ -14,7 +14,20 @@ export enum WgEnv {
 	ServerPort = 'WG_SERVER_PORT',
 }
 
-export type LoggerLevel = 'FATAL' | 'PANIC' | 'WARNING' | 'ERROR' | 'INFO' | 'DEBUG';
+export type LoggerLevel =
```
StarpTech, line 17: This is confusing. The user will see "FATAL" and "fatal" in the IDE. I'd go with one way lowercase.

pkg/node/node.go
```diff
@@ -15,13 +15,15 @@ import (
 	"github.com/jensneuse/abstractlogger"
```
StarpTech, line 15: Can we only use one logger for the node? I see no benefit in using abstactlogger and zap at the same time.

StarpTech (changes requested):

packages/golang-client/src/index.ts
StarpTech: Ideally, the logs are prefixed with the plugin name :smiley_cat:

cli/helpers/flags.go
spetrunin: flag values used in CliEnv helper, so we will have to move both flags and env helper to commands pkg, to avoid circular references,
alternatively, we could add `cli/flags` pkg, to manage all flags of commands

StarpTech: I see, fine for me.

packages/golang-client/src/index.ts
spetrunin: it is confused me a bit, as seems it is part of SDK but sits outside of it

so SDK logger could be used here

yep, it could be smth like this
```
Logger.child("plugin", "golang-client")
```

But do we have any other plugins?

pkg/node/node.go
spetrunin: we use a single logger already - it is zap, but it should be decorated with an abstract logger on a specific log level coming from the config

abstractlogger is not a logging backend - it is decorator for zap or logrus

packages/golang-client/src/index.ts
StarpTech: Yes, but not all logging something.

packages/golang-client/src/index.ts
```diff
@@ -53,7 +54,9 @@ const gofmt = (code: string) => {
 		// on purpose. Otherwise we throw an error with both the code and
 		// the error message returned by gofmt.
 		if (e instanceof Error && e.message.indexOf('ENOENT') >= 0) {
-			console.log('gofmt is not installed. If you want to prettify the generated code, please install gofmt');
+			Logger.child({ plugin: 'golang-client' }).error(
```
StarpTech, line 57: Could you define the logger at the top? In that case the next one doesn't make the mistake to log without the child.

```ts
const logger = Logger.child({ plugin: 'golang-client' })
```

StarpTech (changes requested):

pkg/node/node.go
spetrunin: It is used in graphql-go-tools as logger abstraction, so it should be wrapped before passing to engine

we could handle it afterwords, but also we should be careful - currently zap starts with debug level and we filter reaching to zap via abstract logger

StarpTech: Fine for me to use it as a wrapper for graphql-go-tools, but I'd not use it as a general abstraction for logs.

StarpTech (approved):
LGTM

PR #242, opened 2022-10-06
chore: enable linting for Go files using golangci-lint

fiam:
Only enabled as  CI check for now, no code changes yet. Also, not
enabling all the intended linters yet, let's do it piece by piece

#### Checklist

- [x] run `make test`


.golangci.yml
```diff
@@ -0,0 +1,21 @@
...
+    # - unused
+    # - gosec
+    # - prealloc
+    - gci
+
+linters-settings:
+  gci:
+    sections:
+      - standard
+      - default
```
spetrunin, line 20: we could add also

```
- prefix(github.com/wundergraph/graphql-go-tools)
```

pkg/wgpb/wundernode_config.pb.go
```diff
@@ -7,10 +7,11 @@
 package wgpb
```
spetrunin, line 8: this file should be excluded as generated

types/go/wgpb/wundernode_config.pb.go
```diff
@@ -7,10 +7,11 @@
 package wgpb
 
 import (
-	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
-	protoimpl "google.golang.org/protobuf/runtime/protoimpl"
 	reflect "reflect"
 	sync "sync"
+
```
spetrunin, line 12: and this one will be removed, as it is accidentally commited old version of pkg/wgbp

StarpTech:
> In progress, don't review yet

@fiam put the PR in draft.

fiam:
> > In progress, don't review yet
> 
> @fiam put the PR in draft.

I want to test the CI, that's why it's not in draft.

types/go/wgpb/wundernode_config.pb.go
fiam: Thanks for the heads up. Rebased without it.

.golangci.yml
fiam: ` - prefix(github.com/wundergraph)` should handle all repos from the organization

fiam:
> > I want to test the CI, that's why it's not in draft.
> 
> CI is triggered in draft state.

It used to not be the case. Thanks for the heads up. I've changed it to draft for now.

.golangci.yml
spetrunin: I know

IMHO current repo should be a separate section

and org repos another one section

fiam: I usually have 3 sections: stdlib, 3rdparty and 1stparty, but if you want a more granular separation we can do so.

spetrunin: I believe  `stdlib, 3rdparty, 1stparty, current repo` is the best

pkg/wgpb/wundernode_config.pb.go
fiam: Thanks for the heads up. I've excluded it.

.golangci.yml
fiam: I'll separate the current repo to a different section. That way we don't need to keep adding specific prefixes if we create more 1st party packages.

StarpTech (approved):
LGTM

PR #243, opened 2022-10-06
feat: subscriptions, ws connection init hook

YuriBuerov:
#### Checklist

- [ ] run `make test`


packages/sdk/src/configure/index.ts
```diff
@@ -198,6 +208,8 @@ export interface HooksConfiguration<
 	};
 	[HooksConfigurationOperationType.Queries]?: Queries;
 	[HooksConfigurationOperationType.Mutations]?: Mutations;
+	// TODO: Subscriptions
+	[HooksConfigurationOperationType.Subscriptions]?: Subscriptions;
```
jensneuse, line 212: I think it doesn't make sense to put it here, because this hook is not about "downstream" subscriptions, but "upstream". Queries, Mutations and Subscriptions are client facing (downstream). The "connectionInit" hook is an upstream problem, not downstream. So I think these hooks belong into the global section.

Also, note that Queries, Mutations and Subscriptions are about a single operation, not the origin as a whole. The "connectionInit" hook is not about a single Operation, but the origin in general.

jensneuse:
I think you're running into solving the issue too early. You should start with the design of the user facing API and propose how origins (introspect.graphql) can be connected to the connectionInit hook instead of going into the internal solution (protobuf) already.

packages/sdk/src/configure/index.ts
YuriBuerov: Yes, but APIs don't have identification, we are just adding them to a list. And there is no direct connection between APIs and hooks, they are connected through the operations. That's why I was thinking it was a good approach.

1) Subscription operation hook approach: I have a data source, and I have an operation - subscription, I need this webhook to be able to propagate payload to the `connection_init` message. Not for queries, not for mutations and etc.  Drawback - in case I'll have multiple subscription operations - I'll need to add this hook multiple times for each op. But this way feels "natural".  This is what I started implementing here.

2) Global hooks.  Connect by origin URL.  When we call the hook from the subscription client we can pass the URL. Drawback: I need to duplicate the origin URL (when defining a data source in `wundergraph.config.ts` and hook in `wundergraf.server.ts`)
```
// wundergraph.server.ts
global: {
  httpTransort: {
     ...
  },
  wsTransport: {
    onWSInitHook: {
      enableForOrigin: "http://localhost:3000/"
      hook: () => {}
    }
}
```

3) Global hooks. Connected by origins. How origins (introspect.graphql) could be connected to the hook? Currently,  we have `apiNamespace` which is an optional parameter. To be able to identify the origin, we will need to add some uniq ID or tag.  Not sure what it should look like. Should the user be able to assign an id/tag? (need to validate uniqueness..) Or it should be done behind the scenes?
```
// wundergraph.server.ts
global: {
  httpTransort: {
     ...
  },
  wsTransport: {
    onWSInitHook: {
      enableForOrigins: [ list of API IDs/tags ? ]
      hook: () => {}
    }
}
```

4)  Not hooks, but serializable functions. Just an idea:
```
.wundergraph/functions/wsInitFunc.ts
```

```
introspect.Graphl(
  subscriptionsWsInitFuncName: 'wsInitFunc'
```
@StarpTech @jensneuse  @spetrunin  I need your help, thoughts, and opinions on that ^

YuriBuerov:
> I think you're running into solving the issue too early. You should start with the design of the user facing API and propose how origins (introspect.graphql) can be connected to the connectionInit hook instead of going into the internal solution (protobuf) already.

Yes, but for me, it was an opportunity to become more familiar with an SDK / configuration. Making changes will be faster now, than doing it for the first time.

packages/sdk/src/configure/index.ts
spetrunin: 2 probably could be the simplest option to implement

but 2, 3 is not declarative one - even if we will have an ID or match by origin, you still should be sure that these settings are in sync

4. Feels rather natural and we could validate at least that such a function exists on a configuration stage

and it could be a first step to this generic functions

jensneuse: What do you think about this approach:

```typescript
// wundergraph.config.ts
const weather = introspect.graphql({
	apiNamespace: 'weather',
	url: 'https://graphql-weather-api.herokuapp.com/',
	introspection: {
                 tag: "weather",
		pollingIntervalSeconds: 5,
	},
});
```

```typescript
// wundergraph.server.ts

import type {Origins} from "../generated/origins";

global: {
  httpTransort: {
     ...
  },
  wsTransport: {
    onWSInitHook: {
      enableForOrigins: [Origins.weather] // weather is a typesafe, generated from the tags from introspection into generated directory 
      hook: () => {}
    }
}
```

YuriBuerov: solves the problem with approach 3.

spetrunin: looks cool, I even wasn't aware that we have already referenced generated types in server.js

do u think it should be under introspection? it feels like it could be a global, cause in the future we will be able to use it to extend some particular API behavior

jensneuse: can be global, yes, updated

StarpTech: I  like the general approach with some modifications. The user doesn't need to tag the datasource. We can use the namespace as an identifier or allow it to pass an `ID`. We need this for logging as well. Also, the user doesn't need to import something. Similar to `enableForOperations` in hooks.

```ts
// wundergraph.server.ts

global: {
  httpTransport: {
     ...
  },
  wsTransport: {
    onWSInit: {
     enableForDatasources: ["weather"], // auto complete
     hook () => {}
   }
}
```

jensneuse: I'd like to vote against "namespace". It has a different purpose (namespacing in the resulting graph).
Using just "weather", a string as the tag is fine.
But I'd like to avoid using the namespace as it's a different concept I think.
I think in general it can help to tag things.
WDYT?

StarpTech: Tagging has a different meaning and looks too complex for this problem. What we need is an ID or unique name.

```ts
// wundergraph.config.ts
const weather = introspect.graphql({
    name: "weather",
	apiNamespace: 'weather',
	url: 'https://graphql-weather-api.herokuapp.com/',
	introspection: {
		pollingIntervalSeconds: 5,
	},
});
```

jensneuse: this looks repetitive -_-

StarpTech: How can it be repetitive if you say namespace is a different concept?

```
// wundergraph.config.ts
const weather = introspect.graphql({
    name: "myapiname",
	apiNamespace: 'weather',
	url: 'https://graphql-weather-api.herokuapp.com/',
	introspection: {
		pollingIntervalSeconds: 5,
	},
});
```

jensneuse:
field name: id

packages/sdk/src/configure/index.ts
StarpTech: Decision: Go with a new field `id`, which is optional and identifies the API in the SDK and node.

jensneuse:
@YuriBuerov please review, then we can merge it.

jensneuse (approved):

PR #247, opened 2022-10-07
chore: add formatting and ci tools for go

fiam:
- Install golangci-lint and gci via check-setup
- Add Makefile targets for format-go and golang-ci

#### Checklist

- [x] run `make test`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,6 +31,12 @@ test: test-go test-ts
 format-templates:
 	pnpx prettier --write pkg/templates/assets/templates --ignore-unknown
 
+golang-ci:
+	 golangci-lint run
+
+golang-ci-fix:
+	 golangci-lint run --fix
+
 install-proto:
 	go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.28.1
 
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,9 @@ engine-dev: codegen
 check-setup:
 	$(shell ./scripts/check-setup.sh)
 
+setup-dev:
+	./scripts/setup-dev.sh
+
 bootstrap-minio:
 	./scripts/minio-setup.sh
 
--- a/scripts/check-setup.sh
+++ b/scripts/check-setup.sh
@@ -8,3 +8,8 @@ if ! [ -x "$(command -v protoc)" ]; then
 	echo 'curl -LO "https://github.com/protocolbuffers/protobuf/releases/download/v21.5/protoc-21.5-linux-x86_64.zip"'
 	exit 1
 fi
+
+if ! [ -x "$(command -v golangci-lint)" ]; then
+	echo "Installing golangci-lint"
+	go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest
+fi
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -28,9 +28,15 @@ test-ts:
 
 test: test-go test-ts
 
+format-go:
+	gci write --skip-generated -s "standard,default,prefix(github.com/wundergraph),prefix(github.com/wundergraph/wundergraph)" cli cmd internal pkg
```
spetrunin, line 32: Could gci pickup config from .golangci.yml?

or do we have an option for golangcilint to write this?

fiam: Unfortunately, not that I know of. `golangci-lint` doesn't call other binaries, bundling them within a single binary instead. I know having to maintain the import list in two places is not ideal, but this approach lets us avoid managing each linter separately (and having a config file per linter, duplicating the exclusions, etc...)

spetrunin: https://github.com/golangci/golangci-lint/pull/2892

there is some autofix option

scripts/check-setup.sh
```diff
@@ -8,3 +8,13 @@ if ! [ -x "$(command -v protoc)" ]; then
 	echo 'curl -LO "https://github.com/protocolbuffers/protobuf/releases/download/v21.5/protoc-21.5-linux-x86_64.zip"'
 	exit 1
 fi
+
+if ! [ -x "$(command -v golangci-lint)" ]; then
+	echo "Installing golangci-lint"
+	go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest
```
spetrunin, line 14: https://golangci-lint.run/usage/install/


linter docs do not recommend to setup it via goinstall

fiam: Expand on the `Why` and you'll see the reasons are mostly either outdated (go get) or harmless (you can use master and it's slower than binary installation). The reason I prefer to go with `go install` is that we can rely on it working on Linux, macOS and Windows. If we use an HTTP based download we need to specialise per platform.

Makefile
spetrunin: seems smth like this

```
golangci-lint run --fix
```

fiam: Ahh, thanks a lot! I've been using it for a lot of time, and that's option is relatively new. That's super helpful in this case because we can even skip installing gci manually. I'll update the PR.

fiam: I've tested it and it does indeed apply the formatting fixes (docs said not all linters supported it), awesome!

spetrunin: :+1:

spetrunin (approved):

StarpTech (changes requested):

scripts/check-setup.sh
```diff
@@ -8,3 +8,8 @@ if ! [ -x "$(command -v protoc)" ]; then
...
```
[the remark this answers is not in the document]
fiam: Renamed and Makefile target added

StarpTech (approved):
LGTM

fiam:
@StarpTech Thanks for the review! Could you please override the check and merge? This is hitting the `build_test` filter problem that we talked about. We should look into it next week, but it would be nice to close PRs in the meantime.

PR #248, opened 2022-10-10
fix: live query support in generated web client

xzyfer:
This regression was introduced in #78. The client SDK expects this properted to be `liveQuery`.

See https://github.com/wundergraph/wundergraph/blob/b0bd559d8f2e2785f5f27029bf113da19e13dd65/packages/sdk/src/client/client.ts#L253

#### Checklist

- [ ] run `make test`


jensneuse:
@thisisnithin @JivusAyrus can you verify this?

thisisnithin:
> @thisisnithin @JivusAyrus can you verify this?

Yup, the interface has `liveQuery`

StarpTech:
It also looks good to me. @thisisnithin did you verify the behaviour?

StarpTech:
@xzyfer thank you for the fix.

PR #249, opened 2022-10-10
fix: incorrect go code generated from JSON models

fiam:
- fix: don't swallow gofmt errors, except gofmt being not present
- fix: go code generated from JSON mdel was missing a \n after slices

#### Checklist

- [x] run `make test`

### DIFF
--- a/packages/golang-client/src/index.ts
+++ b/packages/golang-client/src/index.ts
@@ -21,6 +21,23 @@ const defaultTemplateConfig: GolangClientTemplateConfig = {
 	packageName: 'client',
 };
 
+/**
+ * Formats a given string with each line prepended by its line number
+ * (starting at 1).
+ *
+ * @remarks
+ * Lines are split by \r?\n, then joined by \n
+ *
+ * @param code - The code to split and prepend line numbers to
+ * @returns A string with the line annotated code
+ */
+function linefy(code: string): string {
+	return code
+		.split(/\r?\n/)
+		.map((line, idx) => `${idx + 1}: ${line}`)
+		.join('\n');
+}
+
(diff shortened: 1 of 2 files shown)


packages/golang-client/src/index.ts
```diff
@@ -34,7 +34,11 @@ const gofmt = (code: string) => {
 	} catch (e: any) {
 		// we silently ignore the error on purpose
 		// It's not a must to prettify the code
-		console.error('gofmt is not installed. If you want to prettify the generated code, please install gofmt');
+		if (e instanceof Error && e.message.indexOf('ENOENT') >= 0) {
```
StarpTech, line 37: Swallowing was intentional because we can't expect that everyone has go installed.

StarpTech (changes requested):

packages/golang-client/src/index.ts
fiam: The problem was that it was also swallowing  other errors, like feeding gofmt invalid code. With this change, it will still swallow the exception when gofmt is not present (printing a warning), but it will fail when gofmt is installed but returns an error.

jensneuse (approved):

packages/golang-client/src/index.ts
StarpTech: Thanks for the explanation. Makes sense.

StarpTech (approved):
LGTM

fiam:
I've changed the `packages-ci` workflow to also install Go, so we get a `gofmt` recent enough to support generics. Otherwise, CI would fail because our Go code generator does emit generics.

PR #256, opened 2022-10-12
feat: allows node to sleep after an idle timeout

fiam:
- feat: add httpidletimeout package
- feat: implement idle timeout in wunderctl start

#### Checklist

- [x] run `make test`

### DIFF
--- a/cli/commands/node.go
+++ b/cli/commands/node.go
@@ -10,6 +10,7 @@ import (
 	"os/signal"
 	"path"
 	"syscall"
+	"time"
 
 	"github.com/jensneuse/abstractlogger"
 	"github.com/spf13/cobra"
@@ -44,7 +45,7 @@ var nodeStartCmd = &cobra.Command{
 		}
 
 		g.Go(func() error {
-			err := StartWunderGraphNode(n)
+			err := StartWunderGraphNode(n, stop)
 			if err != nil {
 				log.Error("Start node", abstractlogger.Error(err))
 			}
@@ -64,6 +65,8 @@ var nodeStartCmd = &cobra.Command{
 func init() {
 	nodeCmd.AddCommand(nodeStartCmd)
 	rootCmd.AddCommand(nodeCmd)
+
+	nodeStartCmd.Flags().IntVar(&exitAfterIdle, "exit-after-idle", 0, "exits after the given timeout has elapsed without any requests, in seconds")
 }
 
 func NewWunderGraphNode(ctx context.Context) (*node.Node, error) {
@@ -75,7 +78,7 @@ func NewWunderGraphNode(ctx context.Context) (*node.Node, error) {
 	return node.New(ctx, BuildInfo, wunderGraphDir, log), nil
 }
 
-func StartWunderGraphNode(n *node.Node, opts ...node.Option) error {
+func StartWunderGraphNode(n *node.Node, stop func()) error {
 	configFile := path.Join(n.WundergraphDir, "generated", configJsonFilename)
 	if !files.FileExists(configFile) {
 		return fmt.Errorf("could not find configuration file: %s", configFile)
@@ -105,13 +108,19 @@ func StartWunderGraphNode(n *node.Node, opts ...node.Option) error {
 		return err
 	}
 
-	allOpts := []node.Option{
+	opts := []node.Option{
 		node.WithStaticWunderNodeConfig(wunderNodeConfig),
 		node.WithDebugMode(rootFlags.DebugMode),
 	}
-	allOpts = append(allOpts, opts...)
 
-	err = n.StartBlocking(allOpts...)
+	if exitAfterIdle > 0 {
+		opts = append(opts, node.WithIdleTimeout(time.Duration(exitAfterIdle)*time.Second, func() {
+			log.Info("exiting due to idle timeout")
+			stop()
+		}))
+	}
+
+	err = n.StartBlocking(opts...)
 	if err != nil {
 		return err
 	}
--- a/cli/commands/node.go
+++ b/cli/commands/node.go
@@ -66,7 +66,7 @@ func init() {
 	nodeCmd.AddCommand(nodeStartCmd)
 	rootCmd.AddCommand(nodeCmd)
 
-	nodeStartCmd.Flags().IntVar(&exitAfterIdle, "exit-after-idle", 0, "exits after the given timeout has elapsed without any requests, in seconds")
+	nodeStartCmd.Flags().IntVar(&shutdownAfterIdle, "shutdown-after-idle", 0, "shuts down the server after given seconds in idle when no requests have been served")
 }
 
 func NewWunderGraphNode(ctx context.Context) (*node.Node, error) {
@@ -113,9 +113,9 @@ func StartWunderGraphNode(n *node.Node, stop func()) error {
 		node.WithDebugMode(rootFlags.DebugMode),
 	}
 
-	if exitAfterIdle > 0 {
-		opts = append(opts, node.WithIdleTimeout(time.Duration(exitAfterIdle)*time.Second, func() {
-			log.Info("exiting due to idle timeout")
+	if shutdownAfterIdle > 0 {
+		opts = append(opts, node.WithIdleTimeout(time.Duration(shutdownAfterIdle)*time.Second, func() {
+			log.Info("shutting down due to idle timeout")
 			stop()
 		}))
 	}
--- a/cli/commands/node.go
+++ b/cli/commands/node.go
@@ -75,7 +75,7 @@ func NewWunderGraphNode(ctx context.Context) (*node.Node, error) {
 	return node.New(ctx, BuildInfo, wunderGraphDir, log), nil
 }
 
-func StartWunderGraphNode(n *node.Node) error {
+func StartWunderGraphNode(n *node.Node, opts ...node.Option) error {
 	configFile := path.Join(n.WundergraphDir, "generated", configJsonFilename)
 	if !files.FileExists(configFile) {
 		return fmt.Errorf("could not find configuration file: %s", configFile)
@@ -105,10 +105,13 @@ func StartWunderGraphNode(n *node.Node) error {
 		return err
 	}
 
-	err = n.StartBlocking(
+	allOpts := []node.Option{
 		node.WithStaticWunderNodeConfig(wunderNodeConfig),
 		node.WithDebugMode(rootFlags.DebugMode),
-	)
+	}
+	allOpts = append(allOpts, opts...)
+
+	err = n.StartBlocking(allOpts...)
 	if err != nil {
 		return err
 	}
--- a/cli/commands/start.go
+++ b/cli/commands/start.go
@@ -17,7 +17,7 @@ var (
 	disableForceHttpsRedirects bool
 	enableIntrospection        bool
 	gracefulTimeout            int
-	exitAfterIdle              int
+	shutdownAfterIdle          int
 )
 
 // startCmd represents the start command
@@ -70,5 +70,5 @@ func init() {
 	startCmd.Flags().BoolVar(&excludeServer, "exclude-server", false, "starts the engine without the server")
 	startCmd.Flags().BoolVar(&enableIntrospection, "enable-introspection", false, "enables GraphQL introspection on /%api%/%main%/graphql")
 	startCmd.Flags().BoolVar(&disableForceHttpsRedirects, "disable-force-https-redirects", false, "disables authentication to enforce https redirects")
-	startCmd.Flags().IntVar(&exitAfterIdle, "exit-after-idle", 0, "exits after the given timeout has elapsed without any requests, in seconds")
+	startCmd.Flags().IntVar(&shutdownAfterIdle, "shutdown-after-idle", 0, "shuts down the server after given seconds in idle when no requests have been served")
 }
--- a/cli/commands/start.go
+++ b/cli/commands/start.go
@@ -6,13 +6,10 @@ import (
 	"os"
 	"os/signal"
 	"syscall"
-	"time"
 
 	"github.com/jensneuse/abstractlogger"
 	"github.com/spf13/cobra"
 	"golang.org/x/sync/errgroup"
-
-	"github.com/wundergraph/wundergraph/pkg/node"
 )
 
 var (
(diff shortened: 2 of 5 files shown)


pkg/httpidletimeout/httpidletimeout.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+// HandlerFunc is a shorthand for wrapping http.HandlerFunc handlers
+func (m *Middleware) HandlerFunc(next http.HandlerFunc) http.Handler {
+	return m.Handler(next)
+}
+
+// C returns a read-only channel which will receive an empty struct every
+// time Middleware triggers
+func (m *Middleware) C() <-chan struct{} {
```
spetrunin, line 107: maybe give it a longer name?

pkg/httpidletimeout/httpidletimeout.go
```diff
@@ -0,0 +1,115 @@
...
+	eventCh  chan event
+	notifyCh chan struct{}
+}
+
+// New returns a new Middleware which will timeout after the given duration.
+// See Middleware for more information.
+func New(timeout time.Duration) *Middleware {
+	eventCh := make(chan event, 1)
+	notifyCh := make(chan struct{}, 1)
+	go func() {
```
spetrunin, line 41: new has a side effects and starts go routine immediately

pkg/httpidletimeout/httpidletimeout.go
```diff
@@ -0,0 +1,115 @@
...
+		parent := context.Background()
+		ctx, cancel := context.WithTimeout(parent, timeout)
+	LOOP:
```
spetrunin, line 44: it is possible to implement it without label?

cli/commands/start.go
```diff
@@ -46,7 +63,7 @@ var startCmd = &cobra.Command{
 		}
 
 		g.Go(func() error {
-			err := StartWunderGraphNode(n)
+			err := StartWunderGraphNode(n, opts...)
```
spetrunin, line 66: I believe it should be implemented within StartWunderGraphNode

With the current approach we have this behavior only for `wunderctl start`

The goal of `start` is to launch node and server in exactly the same way how separate commands behaves

pkg/httpidletimeout/httpidletimeout.go
```diff
@@ -0,0 +1,115 @@
+// Package httpidletimeout implements an HTTP middleware that triggers
```
spetrunin, line 1: does it make sense to put it under `node` pkg? as it fully belongs to node?

cli/commands/start.go
```diff
@@ -35,6 +40,18 @@ var startCmd = &cobra.Command{
 			return err
 		}
 
+		var opts []node.Option
+		if idleTimeout > 0 {
+			middleware := httpidletimeout.New(time.Duration(idleTimeout) * time.Second)
+			defer middleware.Cancel()
```
spetrunin, line 46: Node could have an option `WithCleanupFunc()` - to run smth before node exits

we could have a more specific option `WithIdleTimeoutMiddleware`.

wdt?

spetrunin (changes requested):

pkg/httpidletimeout/httpidletimeout.go
fiam: I'd rather have it at the top level. As it stands, the package doesn't have any dependencies besides stdlib and we could use it to implement a separate proxy if we need it at some point in the future.

pkg/httpidletimeout/httpidletimeout.go
fiam: I think `C` is fine, it's what the `time` package uses (although it's a field instead of a function), but we can change it to something longer. What name would you prefer?

pkg/httpidletimeout/httpidletimeout.go
fiam: Would you prefer an explicit `Start()` method?

cli/commands/start.go
fiam: I like the more specific option, something like `WithIdleTimeout` (if we make it opaque I see no reason to specify it's implemented as a middleware, that would become an implementation detail). In that case, at least for the time being, I wouldn't expose `WithCleanupFunc()` either, handling both the middleware and the cleanup internally.

cli/commands/start.go
fiam: The reason I went with that is because I didn't want to make the flag global. Let me check how cobra handles having multiple flags pointing to the same variable.

pkg/httpidletimeout/httpidletimeout.go
spetrunin: idk, it just unclear from name what it is

for example Trigger

pkg/httpidletimeout/httpidletimeout.go
spetrunin: yep, smth like that

cli/commands/start.go
spetrunin: you could add same flag to both `start` and `node start`

pkg/httpidletimeout/httpidletimeout.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+func (m *Middleware) startRequest() {
+	if atomic.AddUint64(&m.counter, 1) == 1 {
+		m.eventCh <- eventResetTimer
+	}
+}
+
+func (m *Middleware) endRequest() {
+	if atomic.AddUint64(&m.counter, ^uint64(0)) == 0 {
```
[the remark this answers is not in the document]
fiam: There's no method to decrement it. I know it feels a bit weird the first time you see it, but it's explicitly documented in the atomic package.

cli/commands/start.go
```diff
@@ -69,4 +86,5 @@ func init() {
 	startCmd.Flags().BoolVar(&excludeServer, "exclude-server", false, "starts the engine without the server")
 	startCmd.Flags().BoolVar(&enableIntrospection, "enable-introspection", false, "enables GraphQL introspection on /%api%/%main%/graphql")
 	startCmd.Flags().BoolVar(&disableForceHttpsRedirects, "disable-force-https-redirects", false, "disables authentication to enforce https redirects")
+	startCmd.Flags().IntVar(&idleTimeout, "idle-timeout", 0, "exits after the given timeout has elapsed without any requests, in seconds")
```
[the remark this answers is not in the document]
fiam: Sure, I'll change it.

cli/commands/start.go
```diff
@@ -6,17 +6,22 @@ import (
...
+	"github.com/wundergraph/wundergraph/pkg/httpidletimeout"
+	"github.com/wundergraph/wundergraph/pkg/node"
 )
 
 var (
 	excludeServer              bool
 	disableForceHttpsRedirects bool
 	enableIntrospection        bool
 	gracefulTimeout            int
+	idleTimeout                int
```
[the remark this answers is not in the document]
fiam: I'll rename it as the flag.

pkg/httpidletimeout/httpidletimeout.go
```diff
@@ -0,0 +1,115 @@
...
+	LOOP:
+		for {
+			select {
+			case <-ctx.Done():
+				// Deadline triggered
+				cancel()
+				// Avoid blocking here if there are no listeners, otherwise
+				// we could block here and not be able to service an event
+				select {
+				case notifyCh <- struct{}{}:
```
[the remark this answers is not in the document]
fiam: I know, but if you close the channel then you can't send anything else once the timeout triggers once. I think at some point we might want to implement a more sophisticated logic, so I'd preferred a design that sent multiple notifications.

pkg/node/node.go
```diff
@@ -144,6 +145,14 @@ func WithForceHttpsRedirects(forceHttpsRedirects bool) Option {
 	}
 }
 
+// WithMiddleware appends an HTTP middleware to the node's HTTP server.
+// Middlewares run in the same order they were added.
+func WithMiddleware(middleware func(http.Handler) http.Handler) Option {
```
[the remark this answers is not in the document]
fiam: I'll change it

cli/commands/start.go
```diff
@@ -35,6 +40,18 @@ var startCmd = &cobra.Command{
 			return err
 		}
 
+		var opts []node.Option
+		if idleTimeout > 0 {
+			middleware := httpidletimeout.New(time.Duration(idleTimeout) * time.Second)
+			defer middleware.Cancel()
+			opts = append(opts, node.WithMiddleware(middleware.Handler))
+			go func() {
+				<-middleware.C()
```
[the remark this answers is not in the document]
fiam: The channel gives you the most flexibility and it's more lightweight. You can receive or select on it, as you please. You can also chain it with a Context if you want. But we can  expose a blocking function if you prefer.

pkg/httpidletimeout/httpidletimeout.go
```diff
@@ -0,0 +1,115 @@
...
+				default:
+				}
+
+				// Continue looping
+				ctx, cancel = context.WithTimeout(parent, timeout)
+			case ev := <-eventCh:
+				cancel()
+				switch ev {
+				case eventResetTimer:
+					ctx, cancel = context.WithTimeout(parent, infiniteDuration)
```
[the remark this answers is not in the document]
fiam: Timer.Reset() is such a problematic API that I rarely reach for it for anything that requires a reset. I'll take a look and check how it would work with timers.

pkg/httpidletimeout/httpidletimeout.go
StarpTech: or `timedoutChan`

fiam: I've removed it in favour of a function/Context based API, as Jens requested

cli/commands/start.go
fiam: I've tested it and you're right. Having two flags with the same flag name and pointing the same variable is complete fine. I've moved it to StartWunderGraphNode().

StarpTech:
Sorry, I didn't see the tests. Looks pretty solid.

cli/commands/node.go
```diff
@@ -44,7 +45,7 @@ var nodeStartCmd = &cobra.Command{
 		}
 
 		g.Go(func() error {
-			err := StartWunderGraphNode(n)
+			err := StartWunderGraphNode(n, stop)
```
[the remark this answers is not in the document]
fiam: It's the cancellation function for the parent `Context`. We need to pass it to the `Node` because now it handles the idle timeouts itself rather than exposing a middleware.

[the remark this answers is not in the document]
StarpTech: ah its the cancel function.

cli/commands/node.go
```diff
@@ -75,7 +78,7 @@ func NewWunderGraphNode(ctx context.Context) (*node.Node, error) {
 	return node.New(ctx, BuildInfo, wunderGraphDir, log), nil
 }
 
-func StartWunderGraphNode(n *node.Node) error {
+func StartWunderGraphNode(n *node.Node, stop func()) error {
```
StarpTech, line 81: I'd expect that cancellation is communicated through the node context and that the registration of the middleware is part of the Node struct.

StarpTech (changes requested):

cli/commands/node.go
fiam: That's what I went for initially, but that requires quite a big refactor because `StartWunderGraphNode` blocks waiting for `Node.StartBlocking`, which only returns once the node has stopped.

fiam: Unless I'm mistaken, at that point the Context that Node receives is not guaranteed to be the one that sits at the top of hierarchy. Since we want to cancel everything and exit cleanly, I think it would be safer to handle it at the top level. That's why I'm passing the cancelation function from the top.

StarpTech: Could you give it another shoot? I can help you.

fiam: Sure! Maybe I'm misunderstanding what you have in mind. Let's have a chat to ensure we're both on the same page.

jensneuse (approved):
LGTM
@StarpTech please verify as well

pkg/node/node.go
```diff
@@ -400,6 +412,17 @@ func (n *Node) startServer(nodeConfig WunderNodeConfig) error {
 		// ErrorLog: log.New(ioutil.Discard, "", log.LstdFlags),
 	}
 
+	if n.options.idleTimeout > 0 {
+		m := httpidletimeout.New(n.options.idleTimeout)
```
StarpTech, line 416: Could you use more meaningful names here?

pkg/node/node.go
```diff
@@ -144,6 +147,15 @@ func WithForceHttpsRedirects(forceHttpsRedirects bool) Option {
 	}
 }
 
+// WithIdleTimeout makes the Node call the given handler when idleTimeout
+// has elapsed without any requests.
```
StarpTech, line 151: Could you also mention that the handler is called after idleTimeout?

StarpTech (changes requested):

pkg/node/node.go
fiam: Do you mean the `m`? We can call it `middleware`

pkg/node/node.go
fiam: Sure, I'll add that.

pkg/node/node.go
fiam: Thanks for the clarification! I'll get it done.

cli/commands/start.go
```diff
@@ -69,4 +70,5 @@ func init() {
 	startCmd.Flags().BoolVar(&excludeServer, "exclude-server", false, "starts the engine without the server")
 	startCmd.Flags().BoolVar(&enableIntrospection, "enable-introspection", false, "enables GraphQL introspection on /%api%/%main%/graphql")
 	startCmd.Flags().BoolVar(&disableForceHttpsRedirects, "disable-force-https-redirects", false, "disables authentication to enforce https redirects")
+	startCmd.Flags().IntVar(&exitAfterIdle, "exit-after-idle", 0, "exits after the given timeout has elapsed without any requests, in seconds")
```
StarpTech, line 73: "Shutdown the server after given seconds in idle when no requests have been served"

"exit" sounds hard, but we gracefully shut down the server. Maybe "idle-shutdown"?

StarpTech (changes requested):

cli/commands/start.go
fiam: I went with all lowercase and 3rd person form, for consistency with the rest of the flags, but pretty much copied your suggestion verbatim otherwise

StarpTech: thanks

StarpTech (approved):
LGTM

spetrunin (approved):

PR #259, opened 2022-10-13
fix: avoid overwriting the shared graphql_datasource.Factory instance

fiam:
As pointed by Sergey, this is shared among the different data sources so
we can't overwrite it. Make a copy and overwrite its HTTPClient field in
it.

#### Checklist

- [x] run `make test`

### DIFF
--- a/pkg/engineconfigloader/engineconfigloader.go
+++ b/pkg/engineconfigloader/engineconfigloader.go
@@ -169,7 +169,7 @@ func (d *DefaultFactoryResolver) Resolve(ds *wgpb.DataSourceConfiguration) (plan
 			if err != nil {
 				return nil, err
 			}
-			return d.rest.WithHTTPClient(client), nil
+			return d.rest.CopyWithHTTPClient(client), nil
 		}
 		return d.rest, nil
 	case wgpb.DataSourceKind_STATIC:
@@ -185,7 +185,7 @@ func (d *DefaultFactoryResolver) Resolve(ds *wgpb.DataSourceConfiguration) (plan
 			if err != nil {
 				return nil, err
 			}
-			return d.database.WithHTTPClient(client), nil
+			return d.database.CopyWithHTTPClient(client), nil
 		}
 		return d.database, nil
 	default:
--- a/pkg/engineconfigloader/engineconfigloader.go
+++ b/pkg/engineconfigloader/engineconfigloader.go
@@ -164,7 +164,9 @@ func (d *DefaultFactoryResolver) Resolve(ds *wgpb.DataSourceConfiguration) (plan
 			if err != nil {
 				return nil, err
 			}
-			d.graphql.HTTPClient = client
+			graphql := *d.graphql
+			graphql.HTTPClient = client
+			return &graphql, nil
 		}
 		return d.graphql, nil
 	case wgpb.DataSourceKind_REST:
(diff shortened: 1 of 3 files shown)


pkg/engineconfigloader/engineconfigloader.go
```diff
@@ -164,7 +164,9 @@ func (d *DefaultFactoryResolver) Resolve(ds *wgpb.DataSourceConfiguration) (plan
 			if err != nil {
 				return nil, err
 			}
-			d.graphql.HTTPClient = client
+			graphql := *d.graphql
+			graphql.HTTPClient = client
+			return &graphql, nil
```
jensneuse, line 169: Is this necessary for the streaming client as well? I'm not familiar with the implementation of per-datasource timeouts, but maybe we need it for both? But I'm not sure. Please investigate if required.

fiam: I think we need to adjust the streaming client here, because the previous implementation of `Resolve()` didn't touch that field, even when mTLS was configured.

fiam: Updated my comment because it was missing a "don't"

pkg/engineconfigloader/engineconfigloader.go
```diff
@@ -164,7 +164,9 @@ func (d *DefaultFactoryResolver) Resolve(ds *wgpb.DataSourceConfiguration) (plan
 			if err != nil {
 				return nil, err
 			}
-			d.graphql.HTTPClient = client
+			graphql := *d.graphql
```
StarpTech, line 167: I had to read this twice. This will only create a shallow copy of the struct. Nested pointers or array aren't copied. This solution is a great source of issues. I'd implement a copy method on the datasource.

StarpTech (changes requested):

pkg/engineconfigloader/engineconfigloader.go
fiam: That's what I suggested initially, but I was under the impression that no one else agreed with me (there was a comment with an XXX in the original PR). From my POV, these are the potential solutions in decreasing order of preference:

- Method on `graphql_datasource.Factory` (a bit more verbose, but solid moving forward because it lives next to the type definition)
- Shallow copy via pointer dereference (prevents missing fields, but we might introduce a field that we can't shallow copy and then we'll run into problems)
- Ad-hoc copy (potential problem is missing fields as they're added - this is what we had, and I was asked to switch to to a field overwrite on the existing instance)

I implemented `WithHTTPClient` in the other two factories that live in this same repo, but I didn't go ahead with `graphql_datasource.Factory` because it's in a different repo and we'll need some coordination.

So long story short, I would propose we move forward with:

1. Merge the shallow copy, which fixes the bug for the time being
2. Decide how we call the method (Sergey pointed `WithHTTPClient` was not ideal because it's not obvious that it makes a copy)
3. Implement this new method in `graphql-go-tools`
4. Bump `graphql-go-tools`
5. Use this new method to generate a copy with a new `http.Client`

What do you think?

StarpTech: For now, I'd ensure all properties are copied and move that login into a `newGraphqlFactory` and `newHttpDatasourceFactory` function.

fiam: What would be `newHttpDatasourceFactory` be used for? Besides `graphql_datasource.Factory` we also have `oas_datasource.Factory` and `database.Factory`. Note that all 3 rely on an `http.Client`.

jensneuse:
@fiam please resolve conflicts, then we can merge it.

StarpTech (approved):
LGTM

fiam:
This was already fixed in https://github.com/wundergraph/wundergraph/commit/2e4ae8506558165a9bf3ada4b8f45cee55a9f18d

PR #261, opened 2022-10-15
feat: improvements to "getting started" docs page

fiam:
- fix: default output directory in wunderctl init is . - not .wundergraph
- feat: add wunderctl templates
- fix(devdocs): port for docs-website in development is 3005
- feat: minor improvements in getting-started page


docs-website/src/pages/getting-started.md
```diff
@@ -8,10 +8,21 @@ description: Initialize your WunderGraph development environment and get started
 
 This quick start guide will help you to start a new WunderGraph project from scratch.
 
+## Install Node.js
+
+Wundergraph uses Node.js to manage and run your projects. Install it following their
```
StarpTech, line 13: > Wundergraph uses Node.js to manage and run your projects.

This is not accurate. WunderGraph uses Node.js to build the configuration and run the middleware server. The gateway is running with go.

docs-website/src/pages/getting-started.md
```diff
@@ -8,10 +8,21 @@ description: Initialize your WunderGraph development environment and get started
...
+[official instructions](https://nodejs.org).
```
StarpTech, line 14: Point to LTS. This is the only version we actively supported.

cli/commands/templates.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"github.com/wundergraph/wundergraph/pkg/templates"
+)
+
+// templatesCmd represents the templates command
+var templatesCmd = &cobra.Command{
```
StarpTech, line 12: We should not introduce more commands that will be removed soon. We would like to change how templates are managed. We no longer want to manage them with wunderctl. This has several reasons. There is no easy way to download wunderctl binary. The flow starts with npm. Therefore we should provide an npm package that manage the templates e.g `npx create-wundergraph-app@latest`

 A `npx create-wundergraph-app@latest -e nextjs` will checkout the example from a git repository. In that way, we don't need to prepare a release for every template change.

I'd also change templates to examples which I think is cleaner:

```
npx create-wundergraph-app list
npx create-wundergraph-app ---example nextjs
```

This is the same approach of https://nextjs.org/docs

StarpTech (changes requested):

docs-website/src/pages/getting-started.md
fiam: The motivation for that sentence is to explain why Node is needed. Maybe we can leave the "manage" part and remove the "run"?

cli/commands/templates.go
fiam: I wasn't aware of that. In that case, updating the project creation first and then updating the docs might be a better approach.

StarpTech: My fault, we didn't document it.

docs-website/src/pages/getting-started.md
StarpTech: Yes, sounds good.

fiam:
Since we're now working on the tutorials, maybe changing this to link to the tutorials instead of the examples would be a good idea.

StarpTech (approved):
LGTM

jensneuse (approved):

Issue #263, opened 2022-10-17
wunderctl generate hangs when run inside a docker build


rupurt:
### Bug description

`RUN wunderctl generate` inside a docker image hangs until a timeout when trying to generate the prisma engines

### How to reproduce

Cross posted from https://github.com/wundergraph/docker/issues/2

### Expected behavior

`wunderctl generate` completes in a reasonable amount of time

### Environment & setup

- OS: Mac OS, Windows, Debian, CentOS
- Go version: go version go1.18 darwin/arm64
- Database: PostgreSQL
- Node.js version: v18.11.0

### WunderCtl Version

```
Version: 0.111.4
Commit: 4e9eb8f7500fa37afa524890cae122cf801033b9
Date: 2022-10-10T17:58:54Z
BuiltBy: ci
```

Pagebakers:
Closing this in favor of https://github.com/wundergraph/docker/issues/2

PR #267, opened 2022-10-17
fix: make error handling in IntrospectPrismaDatabaseSchema more robust

fiam:
- Use a buffer protected by a mutex instead of bytes.Buffer
- Pass the buffer only as stdout, if anything is printed to stderr, exit
- Check if the command exits while we're waiting for its output and
return an error

refs: https://github.com/wundergraph/docker/issues/2
refs: https://github.com/wundergraph/wundergraph/issues/263

#### Checklist

- [x] run `make test`


pkg/datasources/database/engine.go
```diff
@@ -24,6 +25,41 @@ import (
 	"github.com/wundergraph/graphql-go-tools/pkg/repair"
 )
 
+// lockedBuffer implements a multithread-safe bytes.Buffer
+type lockedBuffer struct {
```
StarpTech, line 29: In the core, we use https://github.com/go-cmd/cmd I'd use it here as well.

fiam: I've changed it to use go-cmd

StarpTech: :+1:

PR #270, opened 2022-10-18
feat: make WunderGraphConfigApplicationConfig.cors optional

fiam:
If nothing is provided, default to allowing all origins

#### Checklist

- [x] run `make test`

### DIFF
--- a/packages/sdk/src/configure/index.ts
+++ b/packages/sdk/src/configure/index.ts
@@ -86,7 +86,7 @@ export interface WunderGraphConfigApplicationConfig {
 	codeGenerators?: CodeGen[];
 	options?: NodeOptions;
 	server?: WunderGraphHooksAndServerConfig;
-	cors: WunderGraphCorsConfiguration;
+	cors?: WunderGraphCorsConfiguration;
 	s3UploadProvider?: S3Provider;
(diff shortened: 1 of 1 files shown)


StarpTech (approved):
LGTM

fiam:
@jensneuse No, with this PR leaving `WunderGraphConfigApplicationConfig.cors` will always omit the CORS headers from the responses. Would you prefer if we allowed all hosts by default? It's a 1-line change.

jensneuse:
> @jensneuse No, with this PR leaving `WunderGraphConfigApplicationConfig.cors` will always omit the CORS headers from the responses. Would you prefer if we allowed all hosts by default? It's a 1-line change.

Yes. Allow all origins by default.

fiam:
> > @jensneuse No, with this PR leaving `WunderGraphConfigApplicationConfig.cors` will always omit the CORS headers from the responses. Would you prefer if we allowed all hosts by default? It's a 1-line change.
> 
> Yes. Allow all origins by default.

Thanks for the clarification. I’ll change it.

fiam:
@jensneuse There are many ways to set the default to allow all origins, so please let me know what you think of this way and I'll update the PR with the appropriate documentation changes.

PR #274, opened 2022-10-19
feat: healthcheck, node and hooks server

YuriBuerov:
#### Checklist

- [ ] run `make test`

### DIFF
--- a/cli/commands/node.go
+++ b/cli/commands/node.go
@@ -138,11 +138,7 @@ func StartWunderGraphNode(n *node.Node, stop func(), opts ...Option) error {
 	}
 
 	if options.hooksServerHealthCheck {
-		nodeOpts = append(nodeOpts, node.WithHooksServerHealthCheck())
-	}
-
-	if healthCheckTimeout > 0 {
-		nodeOpts = append(nodeOpts, node.WithHealthCheckTimeout(time.Duration(healthCheckTimeout)*time.Second))
+		nodeOpts = append(nodeOpts, node.WithHooksServerHealthCheck(time.Duration(healthCheckTimeout)*time.Second))
 	}
 
 	err = n.StartBlocking(nodeOpts...)
--- a/cli/commands/node.go
+++ b/cli/commands/node.go
@@ -45,7 +45,7 @@ var nodeStartCmd = &cobra.Command{
 		}
 
 		g.Go(func() error {
-			err := StartWunderGraphNode(n, stop)
+			err := StartWunderGraphNode(n, WithIdleHandler(stop))
 			if err != nil {
 				log.Error("Start node", abstractlogger.Error(err))
 			}
@@ -80,6 +80,7 @@ func NewWunderGraphNode(ctx context.Context) (*node.Node, error) {
 
 type options struct {
 	hooksServerHealthCheck bool
+	idleHandler            func()
 }
 
 type Option func(options *options)
@@ -90,7 +91,13 @@ func WithHooksServerHealthCheck() Option {
 	}
 }
 
-func StartWunderGraphNode(n *node.Node, stop func(), opts ...Option) error {
+func WithIdleHandler(idleHander func()) Option {
+	return func(options *options) {
+		options.idleHandler = idleHander
+	}
+}
+
+func StartWunderGraphNode(n *node.Node, opts ...Option) error {
 	var options options
 	for i := range opts {
 		opts[i](&options)
@@ -133,7 +140,7 @@ func StartWunderGraphNode(n *node.Node, stop func(), opts ...Option) error {
 	if shutdownAfterIdle > 0 {
 		nodeOpts = append(nodeOpts, node.WithIdleTimeout(time.Duration(shutdownAfterIdle)*time.Second, func() {
 			log.Info("shutting down due to idle timeout")
-			stop()
+			options.idleHandler()
 		}))
 	}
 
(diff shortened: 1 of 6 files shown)


packages/sdk/src/middleware/server.ts
```diff
@@ -165,6 +165,14 @@ export const createServer = async ({
 
 	fastify.decorateRequest('ctx', null);
 
+	fastify.route({
+		method: 'GET',
+		url: '/',
```
StarpTech, line 170: Let's call it "/health" to make it clear.

pkg/node/node.go
```diff
@@ -411,8 +411,18 @@ func (n *Node) startServer(nodeConfig WunderNodeConfig) error {
 	}()
 
 	router.Handle("/", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
StarpTech, line 413: Here too call it "/health"

pkg/node/node.go
```diff
@@ -411,8 +411,18 @@ func (n *Node) startServer(nodeConfig WunderNodeConfig) error {
...
-		_, _ = w.Write([]byte("WunderNode Status: OK\n"))
-		_, _ = w.Write([]byte(fmt.Sprintf("BuildInfo: %+v\n", n.info)))
+		resp, err := http.DefaultClient.Get(serverUrl + "/")
```
StarpTech, line 414: Please ensure proper max timeouts.

pkg/node/node.go
```diff
@@ -411,8 +411,18 @@ func (n *Node) startServer(nodeConfig WunderNodeConfig) error {
...
+		if err != nil || resp.StatusCode != 200 {
+			serverStatus = "DEAD"
+		}
+
+		health := HealthCheck{
+			ServerStatus: serverStatus,
+			NodeStatus:   "OK",
+			BuildInfo:    n.info,
+		}
+		_ = json.NewEncoder(w).Encode(health)
```
StarpTech, line 425: The endpoint has to return with a non-200 status code when the server is unhealthy.

pkg/node/node.go
StarpTech, line 414: This check should only happen when wunderctl manage the server process "wunderctl start"

StarpTech (changes requested):

cli/commands/node.go
```diff
@@ -78,7 +78,24 @@ func NewWunderGraphNode(ctx context.Context) (*node.Node, error) {
...
+
+type Option func(options *options)
+
+func WithHooksServerHealthCheck() Option {
+	return func(options *options) {
+		options.hooksServerHealthCheck = true
+	}
+}
+
+func StartWunderGraphNode(n *node.Node, stop func(), opts ...Option) error {
```
StarpTech, line 93: Could you convert "stop" also to an option?

pkg/node/node.go
```diff
@@ -410,11 +424,21 @@ func (n *Node) startServer(nodeConfig WunderNodeConfig) error {
...
-	router.Handle("/", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		resp, err := http.DefaultClient.Get(serverUrl + "/")
-		serverStatus := "OK"
-		if err != nil || resp.StatusCode != 200 {
-			serverStatus = "DEAD"
+	router.Handle("/health", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		serverStatus := "SKIP"
+		if n.options.hooksServerHealthCheck {
+			c := http.DefaultClient
+			c.Timeout = 30 * time.Second
```
StarpTech: I'd suggest creating a custom HTTP client outside of the handler.

pkg/node/node.go
```diff
@@ -410,11 +424,21 @@ func (n *Node) startServer(nodeConfig WunderNodeConfig) error {
...
+			if n.options.healthCheckTimeout != nil {
+				c.Timeout = *n.options.healthCheckTimeout
+			}
+
+			resp, err := c.Get(serverUrl + "/health")
```
StarpTech, line 431: Use

```
url.JoinPath("http://host/foo", "bar/")
```

pkg/node/node.go
StarpTech: The default should come from the cli flag default value.

cli/commands/start.go
```diff
@@ -71,4 +72,5 @@ func init() {
 	startCmd.Flags().BoolVar(&enableIntrospection, "enable-introspection", false, "enables GraphQL introspection on /%api%/%main%/graphql")
 	startCmd.Flags().BoolVar(&disableForceHttpsRedirects, "disable-force-https-redirects", false, "disables authentication to enforce https redirects")
 	startCmd.Flags().IntVar(&shutdownAfterIdle, "shutdown-after-idle", 0, "shuts down the server after given seconds in idle when no requests have been served")
+	startCmd.Flags().IntVar(&healthCheckTimeout, "healthcheck-timeout", 0, "healthcheck timeout in seconds")
```
StarpTech, line 75: You can define a default here.

StarpTech (changes requested):

pkg/node/node.go
spetrunin: It will require go 1.19

StarpTech: :cry:

StarpTech (approved):
LGTM

PR #277, opened 2022-10-19
Changed missing end_session_endpoint error to an end_session_endpoint…

John-Appleseed:
End session endpoint is optional.


jensneuse (approved):

jensneuse:
@John-Appleseed can you also set the Error to Warn when end session endpoint is missing?

PR #285, opened 2022-10-25
fix: crash when message.Location is nil

fiam:
Triggered when esbuild encounters an error which is not associated with
a source code location (e.g. file referenced from command line is
missing)


pkg/bundler/bundler.go
```diff
@@ -240,11 +240,17 @@ func (b *Bundler) watch(ctx context.Context, rebuild func() api.BuildResult) {
 				}
 			} else {
 				for _, message := range result.Errors {
+					location := message.Location
+					if location == nil {
+						location = &api.Location{
+							File: "????",
```
jensneuse, line 246: how about replacing `????` with `unknown`

fiam: I was afraid people would start looking for a file named `unknown` in their `.wundergraph` so I used `????` because `?` is not allowed in filenames in Windows, and I guess on Unix people assume it's not meant to be taken as a literal filename. I've seen some programs use `<stdin>` to refer to the standard input, though. Maybe we can use `<unknown>`?

jensneuse (approved):

PR #286, opened 2022-10-25
chore: increase go linter timeout in CI to 2 minutes

fiam:
1 minute seems a bit tight and is timing out from time to time.
Also, pin the golangci version to avoid having a new release trigger new
errors.

### DIFF
--- a/.github/workflows/engine-ci.yaml
+++ b/.github/workflows/engine-ci.yaml
@@ -9,6 +9,7 @@ on:
       - 'types/**/*'
       - 'go.mod'
       - 'go.sum'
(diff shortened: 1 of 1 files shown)


StarpTech:
I'm fighting for fast CI builds. If the linter needs more than 1min to check our code, we should evaluate alternatives.

fiam:
Fast CI builds are nice, but I think waiting 1-2 minutes (right now it takes around 65 seconds) is a totally worthy tradeoff to ensure we catch preventable errors.

jensneuse (approved):

PR #291, opened 2022-10-26
feat: add flag for disabling environment loading from file

fiam:
- Add --no-env, which disables the .env file
- Refactor the checks for environment files that do not exist and
report a failure if the environment file is set to something else than
the default, but does not exist.

### DIFF
--- a/cli/commands/root.go
+++ b/cli/commands/root.go
@@ -3,7 +3,6 @@ package commands
 import (
 	"errors"
 	"fmt"
-	"io/fs"
 	"net/http"
 	"os"
 	"path/filepath"
@@ -32,6 +31,7 @@ const (
 	serverEntryPointFilename = "wundergraph.server.ts"
 
 	defaultNodeGracefulTimeoutSeconds = 10
+	defaultDotEnvFilename             = ".env"
 )
 
 var (
@@ -43,6 +43,7 @@ var (
 	_wunderGraphDirConfig string
 	disableCache          bool
 	clearCache            bool
+	disableEnvFile        bool
 
 	rootFlags helpers.RootFlags
 
@@ -92,18 +93,22 @@ You can opt out of this by setting the following environment variable: WUNDERGRA
 		zapLog := logging.Zap().With(zap.String("component", "@wundergraph/wunderctl"))
 		log = abstractlogger.NewZapLogger(zapLog, logLevel)
 
-		err = godotenv.Load(DotEnvFile)
-		if err != nil {
-			if _, ok := err.(*fs.PathError); ok {
-				log.Debug("starting without env file")
+		if !disableEnvFile {
+			err = godotenv.Load(DotEnvFile)
+			if err != nil {
+				// Only swallow the error if the env file has its default value,
+				// which likely means it was not specified.
+				if DotEnvFile == defaultDotEnvFilename && errors.Is(err, os.ErrNotExist) {
+					log.Debug("starting without env file")
+				} else {
+					log.Fatal("error loading env file",
+						abstractlogger.Error(err))
+				}
 			} else {
-				log.Fatal("error loading env file",
-					abstractlogger.Error(err))
+				log.Debug("env file successfully loaded",
+					abstractlogger.String("file", DotEnvFile),
+				)
 			}
-		} else {
-			log.Debug("env file successfully loaded",
-				abstractlogger.String("file", DotEnvFile),
-			)
 		}
 
 		if clearCache {
@@ -177,7 +182,8 @@ func init() {
 	viper.SetDefault("API_URL", "https://api.wundergraph.com")
 
 	rootCmd.PersistentFlags().StringVarP(&rootFlags.CliLogLevel, "cli-log-level", "l", "info", "sets the CLI log level")
-	rootCmd.PersistentFlags().StringVarP(&DotEnvFile, "env", "e", ".env", "allows you to set environment variables from an env file")
+	rootCmd.PersistentFlags().BoolVar(&disableEnvFile, "no-env", false, "disables loading environment variables from a file")
+	rootCmd.PersistentFlags().StringVarP(&DotEnvFile, "env", "e", defaultDotEnvFilename, "allows you to set environment variables from an env file")
 	rootCmd.PersistentFlags().BoolVar(&rootFlags.DebugMode, "debug", false, "enables the debug mode so that all requests and responses will be logged")
 	rootCmd.PersistentFlags().BoolVar(&rootFlags.PrettyLogs, "pretty-logging", false, "switches to human readable format")
 	rootCmd.PersistentFlags().StringVar(&_wunderGraphDirConfig, "wundergraph-dir", files.WunderGraphDirName, "path to your .wundergraph directory")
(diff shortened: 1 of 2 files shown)


cli/commands/root.go
```diff
@@ -173,7 +178,8 @@ func init() {
 	viper.SetDefault("API_URL", "https://api.wundergraph.com")
 
 	rootCmd.PersistentFlags().StringVarP(&rootFlags.CliLogLevel, "cli-log-level", "l", "info", "sets the CLI log level")
-	rootCmd.PersistentFlags().StringVarP(&DotEnvFile, "env", "e", ".env", "allows you to set environment variables from an env file")
+	rootCmd.PersistentFlags().BoolVar(&disableEnvFile, "no-env", false, "disables loading environment variables from a file")
```
StarpTech, line 181: Hi, did you check if existing env variables take precedence over variables with the same name in .env?

StarpTech: I think we don't need this flag if we can ensure that environment variables passed by the user take precedence over variables in `.env`. In https://github.com/joho/godotenv you can read them to a map and implement this behavior.

StarpTech (changes requested):

cli/commands/root.go
fiam: I've verified that environment variables take precedence over the `.env` file. `godotenv` handles it here https://github.com/joho/godotenv/blob/5c76d3e02c63f90a1bfae60738bbc24ba16f05c4/godotenv.go#L190

StarpTech: Great, do you think it still makes sense to add this flag?

fiam: While the flag is not entirely mandatory, it might make some sense if you want to be absolutely sure that nothing from an external file pollutes the environment. On the other hand, naming it properly is a bit annoying as we discussed: `--no-env` might seem like it ignores the  previous environment while naming it `--no-env-file` does not make it as obvious that's related to `--env`. I you think that's useful for cloud, I'd keep it. Otherwise we can drop it and if someone asks for it in the future we can revive the PR.

fiam: I agree with renaming the flags, but only loading `.env` from `wunderctl up` seems like it could lead to some unpleasant surprises for users. I think we either load it by default in all cases or either we default `--env-file` to empty and only load it when the user provides a value for the flag.

StarpTech: We should prevent users from making such mistakes. Using `.env` by default could encourage people to commit their secrets to git or disk. This is only reasonable for development. We can expose this flag, but it shouldn't be the default behavior for production commands like `wunderctl start` or `wunderctl node start` ...

fiam: Those are orthogonal concepts. People that think committing secrets is not a problem will still find ways to do so, like adding a script to launch `wunderctl` with the secrets in it.

fiam:
I've changed it to define two flags, one for the root command and another one for `wunderctl up`, in the same way that it was done for `--pretty-logging`. First has an empty value by default, while the latter defaults to `.env`. Problem is cobra doesn't seem to support this. Notice the output of `wunderctl up -h`:

```
Start the WunderGraph application in development mode and watch for changes
 
Usage:
  wunderctl up [flags]

Flags:
  -h, --help   help for up

Global Flags:
      --clear-cache              clears local caches during startup
  -l, --cli-log-level string     sets the CLI log level (default "info")
      --debug                    enables the debug mode so that all requests and responses will be logged
  -e, --env-file string          allows you to set environment variables from an env file
      --no-cache                 disables local caches
      --no-env-file              disables loading environment variables from a file
      --pretty-logging           switches to human readable format
      --wundergraph-dir string   path to your .wundergraph directory (default ".wundergraph")
```
The root command definition is the one that prevails (notice it's shown only as global and with no default). Basically, what this ends up doing is overriding the default value if the command is `up` with no way for the user to discover that (other than trial and error).

cli/commands/root.go
StarpTech: Let me describe it differently. Let's say you accidentally copy your .env file to your docker context. Your app will likely work and you're going crazy about why the wrong variables are used.

cli/commands/root.go
```diff
@@ -177,7 +176,8 @@ func init() {
 	viper.SetDefault("API_URL", "https://api.wundergraph.com")
 
 	rootCmd.PersistentFlags().StringVarP(&rootFlags.CliLogLevel, "cli-log-level", "l", "info", "sets the CLI log level")
-	rootCmd.PersistentFlags().StringVarP(&DotEnvFile, "env", "e", ".env", "allows you to set environment variables from an env file")
+	rootCmd.PersistentFlags().BoolVar(&disableEnvFile, "no-env-file", false, "disables loading environment variables from a file")
+	rootCmd.PersistentFlags().StringVarP(&dotEnvFile, "env-file", "e", "", "allows you to set environment variables from an env file")
```
StarpTech, line 180: `load` instead `set`?

cli/commands/root.go
```diff
@@ -77,6 +77,7 @@ You can opt out of this by setting the following environment variable: WUNDERGRA
 			// it would overwrite the default value for all other commands
 		case UpCmdName:
 			rootFlags.PrettyLogs = upCmdPrettyLogging
+			dotEnvFile = upCmdEnvFile
```
StarpTech, line 80: Now, you're running into the same ugly workaround :smile:

cli/commands/root.go
```diff
@@ -177,7 +176,8 @@ func init() {
 	viper.SetDefault("API_URL", "https://api.wundergraph.com")
 
 	rootCmd.PersistentFlags().StringVarP(&rootFlags.CliLogLevel, "cli-log-level", "l", "info", "sets the CLI log level")
-	rootCmd.PersistentFlags().StringVarP(&DotEnvFile, "env", "e", ".env", "allows you to set environment variables from an env file")
+	rootCmd.PersistentFlags().BoolVar(&disableEnvFile, "no-env-file", false, "disables loading environment variables from a file")
```
StarpTech, line 179: Is it required to have `--no-env-file` and `--env-file`? Who not just `--env-file` and by default `wunderctl up` uses `.env` and all others empty string? WDYT?

StarpTech (changes requested):

cli/commands/root.go
fiam: Maybe we have different workflows in mind. Won’t the environment variables that are already defined take precedence in that case? If you forgot to define a secret I would assume you won’t use the same values in production and development and it will fail during startup.

cli/commands/root.go
fiam: In practical terms, I think we're better served by checking for `dotEnvFile == ""` and command being `up`, then overriding the value. The help for the subcommand seems to be ignored, and that way we avoid repeating its help text.

cli/commands/root.go
fiam: Now that we've changed it to default to an empty string, I think we can remove `--no-env-file`. What bothers me a bit is defaulting to `.env` for up. I'd be more comfortable with leaving it empty in all cases, then if it has a value and it fails to load we can report an error. Having two different defaults makes error handling a bit messy.

StarpTech: `.env` is very common for development exists since the early days of wundergraph. I'd not remove this default for up. Loading .env in development is best-effort.

PR #292, opened 2022-10-26
fix: pin node in CI to version 16

fiam:
Required by some dependencies. At some point we might want to bump
those, though.


StarpTech (approved):
LGTM

StarpTech:
@fiam We should evaluate after that what's blocking us not to support LTS.

fiam:
> @fiam We should evaluate what's blocking us not to support LTS.

I agree, but I didn't want to upgrade dependencies myself and risk breaking something else today. This is just an interim fix.

The dependency that fails seem to be `@apollo/composition` - Check this CI run for reference https://github.com/wundergraph/wundergraph/actions/runs/3329470141/jobs/5506725015

StarpTech:
ok, let's merge it after CI is green.

fiam:
@StarpTech We need to take care of the timeout in Engine CI first https://github.com/wundergraph/wundergraph/pull/286

jensneuse:
@fiam please resolve merge conflicts, update branch, then we can merge I think

PR #296, opened 2022-10-27
feat: make idle timer ignore healtcheck endpoint

fiam:
- feat: add skipping support to HTTP idle timer middleware
- feat: make idle timer ignore healthcheck endpoint


jensneuse (approved):

jensneuse:
It's good, but failing. -_-

fiam:
Thanks for the review. CI had failed because the Go linter had timed out. Now that we've increased the timeout it should be good. I've restarted it and enabled auto-merging.

PR #298, opened 2022-10-28
chore: rename X-Cache header to X-Wg-Cache

fiam:
refactor the cache key to a constant to avoid repeating it

### DIFF
--- a/pkg/apihandler/apihandler.go
+++ b/pkg/apihandler/apihandler.go
@@ -57,9 +57,10 @@ import (
 )
 
 const (
-	WG_PREFIX    = "wg_"
-	WG_LIVE      = WG_PREFIX + "live"
-	WG_VARIABLES = WG_PREFIX + "variables"
+	WG_PREFIX       = "wg_"
+	WG_LIVE         = WG_PREFIX + "live"
+	WG_VARIABLES    = WG_PREFIX + "variables"
+	WG_CACHE_HEADER = "X-Wg-Cache"
 )
 
 type Builder struct {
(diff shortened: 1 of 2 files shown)


pkg/apihandler/apihandler.go
```diff
@@ -57,9 +57,10 @@ import (
 )
 
 const (
-	WG_PREFIX    = "wg_"
-	WG_LIVE      = WG_PREFIX + "live"
-	WG_VARIABLES = WG_PREFIX + "variables"
+	WG_PREFIX       = "wg_"
+	WG_LIVE         = WG_PREFIX + "live"
+	WG_VARIABLES    = WG_PREFIX + "variables"
+	WG_CACHE_HEADER = "X-Wg-Cache"
```
bmcilw1, line 63: Should this be 
```suggestion
	WG_CACHE_HEADER = `X-${WG_PREFIX}-Cache`
```

jensneuse: no, but I'd suggest we make the key "X-WG-Cache" instead of "X-Wg-Cache" @fiam

jensneuse (changes requested):
x

pkg/apihandler/apihandler.go
fiam: `X-WG-Cache` is not in canonical form. While HTTP headers are case insensitive, some APIs/libraries delegate the canonicalisation to the user while others handle it internally (Go falls into the latter, see [CanonicalMimeHeaderKey](https://cs.opensource.google/go/go/+/refs/tags/go1.19.2:src/net/textproto/reader.go;l=579;drc=e09bbaec69a8ff960110e13eabb3bef5331ecb0c;bpv=1;bpt=1)). To avoid having to handle this directly (e.g. writing to `resp.Header` instead of using `resp.Header.Set` and potentially causing issues for consumers, I'd rather go with `X-Wg-Cache`.

bmcilw1: Just for the heck of saying something, since that's true it could also be `X-${WG_PREFIX[0].toUpper() + WG_PREFIX.restOfStringAfter0().toLower()}-Cache`

That is if fully respecting the given prefix was a desired feature.

jensneuse: Ok @fiam

jensneuse (approved):

PR #299, opened 2022-10-28
fix: don't swallow errors in wunderctl loadoperations

fiam:
Always return the Output to the caller, then convert it to string
before printing. Return zero in all cases, since the TS side uses the
exit code to determine wether to parse the outout or print it as a raw
error.

### DIFF
--- a/cli/commands/loadoperations.go
+++ b/cli/commands/loadoperations.go
@@ -19,7 +19,7 @@ var loadoperationsCmd = &cobra.Command{
 		out := loader.Load(args[0], args[1], args[2])
 		// TODO: migrate to writing it to file because it can infer with logs
 		// or log and print to two different streams and respect that on the consumer side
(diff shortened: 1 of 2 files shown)


jensneuse:
Instead of this PR, I'd rather abandon the idea of trying to fix this approach here, I'd go with ENG-649.

fiam:
> Instead of this PR, I'd rather abandon the idea of trying to fix this approach here, I'd go with ENG-649.

This would be a subset of ENG-649

StarpTech (changes requested):
Can we provide a basic test for this?

fiam:
> Can we provide a basic test for this?

Not sure if it's worth the hassle at this point, because however takes ENG-649 might decide to redesign the API.

fiam:
Closed in favour of ENG-649

PR #300, opened 2022-10-28
feat: ensure wunderctl always call the same copy when calling itself

fiam:
Introduce a WUNDERCTL_BINARY_PATH environment variable. This can
be set either by the user or either by wunderctl itself. On startup,
wunderctl checks if the variable is unset and if it is, then it
sets it to its own path. The TS side now checks for this new variable
and when it's non-empty, it overrides the path to the wunderctl binary.

### DIFF
--- a/cli/commands/root.go
+++ b/cli/commands/root.go
@@ -45,7 +45,6 @@ var (
 	_wunderGraphDirConfig string
 	disableCache          bool
 	clearCache            bool
-	wunderctlBinaryPath   string
 
 	rootFlags helpers.RootFlags
 
@@ -69,7 +68,6 @@ wunderctl is gathering anonymous usage data so that we can better understand how
 You can opt out of this by setting the following environment variable: WUNDERGRAPH_DISABLE_METRICS
 `,
 	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
-		setupWunderctlBinaryPath()
 
 		switch cmd.Name() {
 		// skip any setup to avoid logging anything
@@ -173,18 +171,20 @@ func Execute(buildInfo node.BuildInfo, githubAuthDemo node.GitHubAuthDemo) {
 	}
 }
 
-func setupWunderctlBinaryPath() {
-	// Variable might be set, maybe by us or maybe by the
-	// user, but we never override it
-	value, isSet := os.LookupEnv(wunderctlBinaryPathEnvKey)
+// wunderctlBinaryPath() returns the path to the currently executing parent wunderctl
+// command which is then passed via wunderctlBinaryPathEnvKey to subprocesses. This
+// ensures than when the SDK calls back into wunderctl, the same copy is always used.
+func wunderctlBinaryPath() string {
+	// Check if a parent wunderctl set this for us
+	path, isSet := os.LookupEnv(wunderctlBinaryPathEnvKey)
 	if !isSet {
 		// Variable is not set, find out our path and set it
 		exe, err := os.Executable()
 		if err == nil {
-			value = exe
+			path = exe
 		}
 	}
-	wunderctlBinaryPath = value
+	return path
 }
 
 func init() {
--- a/cli/commands/root.go
+++ b/cli/commands/root.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/fatih/color"
@@ -30,6 +31,8 @@ const (
 	configJsonFilename       = "wundergraph.config.json"
 	configEntryPointFilename = "wundergraph.config.ts"
 	serverEntryPointFilename = "wundergraph.server.ts"
+
+	wunderctlBinaryPathEnvKey = "WUNDERCTL_BINARY_PATH"
 
 	defaultNodeGracefulTimeoutSeconds = 10
 )
@@ -66,6 +69,7 @@ wunderctl is gathering anonymous usage data so that we can better understand how
 You can opt out of this by setting the following environment variable: WUNDERGRAPH_DISABLE_METRICS
 `,
 	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+		setupWunderctlBinaryPath()
 
 		switch cmd.Name() {
 		// skip any setup to avoid logging anything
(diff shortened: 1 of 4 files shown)


cli/commands/root.go
```diff
@@ -165,6 +170,21 @@ func Execute(buildInfo node.BuildInfo, githubAuthDemo node.GitHubAuthDemo) {
...
+		if strings.HasPrefix(v, wunderctlBinaryPathEnvKey+"=") {
+			// Variable is set, maybe by us or maybe by the
+			// user, but we never override it
+			return
+		}
+	}
+	// Variable is not set, find out our path and set it
+	exe, err := os.Executable()
+	if err == nil {
+		os.Setenv(wunderctlBinaryPathEnvKey, exe)
```
StarpTech, line 184: I don't like it this way. The same reason why globals are evil. I prefer to set this variable explicitly in the process call of the "config runner" https://github.com/wundergraph/wundergraph/blob/main/cli/commands/up.go#L82.

StarpTech (changes requested):

cli/commands/root.go
fiam: Reading an environment variable, moving it to a Go global, then passing it as the environment for each invocation is still using a global, just with more steps and hence more chances to mess it up (like we have right now with both `WG_ABS_DIR` and `WG_DIR_ABS`. If you want to set environment variables on a per-invocation basis we need to refactor all the environment variables to a centralised place. We have some of them handled in `helpers/env.go`, but others are still handled manually at each invocation.

StarpTech: My point is: It's not clear that the runner requires this env to work. This is very error-prone.

fiam: If the script requires a specific environment variable to work, we should wrap it in a function and require an explicit argument. Adding the environment variables explicitly to each invocation doesn't mean they're required, maybe they were required at some point in the past and someone forgot to remove them, because you have no way to know if they're actually used without looking at what the child (and its potential children) does. If you want to be absolutely sure that the child always receives a given set of environment variables, we should have the child validate its environment.

StarpTech: :+1:

cli/commands/root.go
```diff
@@ -165,6 +170,21 @@ func Execute(buildInfo node.BuildInfo, githubAuthDemo node.GitHubAuthDemo) {
 	}
 }
 
+func setupWunderctlBinaryPath() {
+	for _, v := range os.Environ() {
```
jensneuse, line 174: Replace with idiomatic go:
https://golangbyexample.com/check-environment-variable-set-go/

jensneuse:
Is this passed back to go when called from the wunderctl Node wrapper?

jensneuse (changes requested):

cli/commands/root.go
fiam: Good idea. I wasn't aware of that.

cli/commands/root.go
StarpTech: This env variable is only needed when generating the config. Could you pass the env variable as with the others? https://github.com/wundergraph/wundergraph/blob/main/cli/commands/up.go#L82

fiam: What if we take the chance to refactor all that? We have variables duplicated in a couple of places and right now, with less than 10 variables, it shouldn't take long to organise them.

StarpTech: Could you make an example of how that refactoring would look like? IMO I don't see much room for valuable refactoring.

fiam: Move all variables passed to scripts to functions. e.g. `GeneratorEnvironment()`, `HooksServerEnviroment()`, etc.. and don't type any variables directly when creating the `ScriptRunner`. If we need to adjust those variables depending on something else, pass them as parameters to those functions.

StarpTech: This would introduce another abstraction. We would still pass it like this right?

```go
configRunner := scriptrunner.NewScriptRunner(&scriptrunner.Config{
			Name:          "config-runner",
			Executable:    "node",
			AbsWorkingDir: wunderGraphDir,
			Logger:        log,
			ScriptEnv: HooksServerEnviroment(disableCache, .....)
		})
```

I'd just pass the wunderctlBinaryPathEnvKey to the runner for now.

jensneuse: Agree. Just pass on the process env. Simple and clear.

fiam: I've refactored it to pass each the variable to each invocation individually

cli/commands/root.go
```diff
@@ -169,6 +173,20 @@ func Execute(buildInfo node.BuildInfo, githubAuthDemo node.GitHubAuthDemo) {
...
+	// user, but we never override it
+	value, isSet := os.LookupEnv(wunderctlBinaryPathEnvKey)
+	if !isSet {
+		// Variable is not set, find out our path and set it
+		exe, err := os.Executable()
+		if err == nil {
+			value = exe
+		}
+	}
+	wunderctlBinaryPath = value
```
StarpTech, line 187: Can we avoid working with a global here? Why not

```go
path, err := getBinaryPath()
fmt.Sprintf("%s=%s", wunderctlBinaryPathEnvKey, path),
```

StarpTech (changes requested):

cli/commands/root.go
fiam: There used to be a time when `os.Executable()` could get a bit fishy after starting up (if e.g. you `chdir()`'ed), but I checked its most recent source code and it should now work fine. I will remove the global variable.

fiam: Changed

packages/wunderctl/src/binarypath.ts
```diff
@@ -52,5 +52,8 @@ export const wunderctlName = (): string => {
 export const binaryDir = path.join(__dirname, '..', 'download');
 
 export const wunderctlBinaryPath = (): string => {
+	if (process.env.WUNDERCTL_BINARY_PATH) {
+		return process.env.WUNDERCTL_BINARY_PATH;
```
StarpTech, line 56: Should we validate for a valid path? I could imagine the error message is not very helpful when the path does not point to a binary.

StarpTech (changes requested):

packages/wunderctl/src/binarypath.ts
fiam: At least for now, we're the only ones setting that variable. If someone is setting it manually before invocation I think letting it blow up makes it more evident than swallowing the error. There's also already a check in `wunderCtlFile()` which throws if the binary doesn't exist, so I'd say there's already enough validation.

StarpTech: ok

StarpTech (approved):
LGTM

jensneuse (approved):

jensneuse:
LGTM

PR #301, opened 2022-10-28
fix: make introspection cache keys change when referenced files change

fiam:
When calculating the cache key, try reading any string found as if it
was a file. If we succeed at reading it, add it to the hash. This
ensures any change to a referenced file is picked up when calculating
the key and all data is regenerated accordingly.

### DIFF
--- a/packages/sdk/src/definition/introspection-cache.ts
+++ b/packages/sdk/src/definition/introspection-cache.ts
@@ -7,10 +7,10 @@ import {
 	WG_ENABLE_INTROSPECTION_CACHE,
 	WG_ENABLE_INTROSPECTION_OFFLINE,
 } from './index';
+import * as crypto from 'crypto';
 import path from 'path';
 import fsP from 'fs/promises';
 import { FieldConfiguration, TypeConfiguration } from '@wundergraph/protobuf';
-import objectHash from 'object-hash';
 import { Logger } from '../logger';
 import { onParentProcessExit } from '../utils/process';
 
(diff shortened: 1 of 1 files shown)


packages/sdk/src/definition/introspection-cache.ts
```diff
@@ -194,3 +194,44 @@ export const introspectWithCache = async <Introspection extends IntrospectionCon
...
+			await objectHash(hash, v);
+		} else if (object.hasOwnProperty(k)) {
+			hash.update(typeof v === 'undefined' ? 'undefined ' : v.toString());
+			if (typeof v === 'string') {
+				// The value might reference a file. Try to read
+				// and hash it.
+				await fileHash(hash, v);
+				if (process.env.WG_DIR_ABS) {
+					await fileHash(hash, path.join(process.env.WG_DIR_ABS, v));
+				}
```
jensneuse, line 227: Why hash it twice?
If line 224 succeeds, why hash again with abs dir path join?
Also, should you not check if the path is relative?
If it's an absolute file path, does it make sense to join the path?

fiam: There's no harm in hashing the file multiple times (other than a small runtime penalty), so I preferred to err on the safe side. Since at that point we cannot determine the exact file authoritatively, we need to make a guess. Under those circumstances we cannot assume that the hashing in line 224 picked the correct file, because the string in the object might reference a file in the current directory or maybe a file in `.wundergraph`. A more correct, but more tedious, approach is to go through all possible properties that might reference a file and walk them explicitly, but that takes more maintenance (and more chances to miss something).

jensneuse: Ok

jensneuse (approved):

fiam:
Thanks  👍

PR #302, opened 2022-10-28
feat: sdk, support schema extension

YuriBuerov:
#### Checklist

- [ ] run `make test`


YuriBuerov:
I would like to have feedback before I update the docs

packages/sdk/src/transformations/shema.ts
```diff
@@ -0,0 +1,104 @@
+import { ReplaceJSONTypeFieldConfiguration } from '../definition';
+import { parse, print, visit } from 'graphql/index';
+
+const replaceCustomScalars = <
+	T extends {
+		schemaExtension?: string | undefined;
+		replaceJSONTypeFields?: ReplaceJSONTypeFieldConfiguration[] | undefined;
```
jensneuse, line 7: I think we agreed to go away from naming this JSON, right?

packages/sdk/src/definition/index.ts
```diff
@@ -239,6 +239,10 @@ interface GraphQLIntrospectionOptions {
 	// so that the request context can be enriched
 	internal?: boolean;
 	skipRenameRootFields?: string[];
+	// the schemaExtension field is used to extend the generated GraphQL schema with additional types and fields
+	// this is useful for specifying type definitions for JSON objects
+	schemaExtension?: string;
+	replaceJSONTypeFields?: ReplaceJSONTypeFieldConfiguration[];
```
jensneuse, line 245: rename to "replaceCustomScalarTypeFields" (consistently across all data sources)

packages/sdk/src/v2openapi/index.test.ts
```diff
@@ -93,3 +94,35 @@ test('subscription as object field', async () => {
...
+					type: String
+					phone: String
+				}
+				`,
+		replaceJSONTypeFields: [
+			{
+				entityName: `User`,
+				fieldName: `contact`,
+				responseTypeReplacement: `Contact`,
+				inputTypeReplacement: `ContactInput`,
```
jensneuse, line 122: Is this optional? Because it should. As you can see in this example, we're not using it.
I think for GraphQL and OpenAPI, we really only set responseTypeReplacement.

jensneuse (changes requested):

jensneuse (approved):

PR #305, opened 2022-10-31
feat: improved SWR hooks api, useMutation result no longer cached and updated to SWR 2.0

Pagebakers:
#### Checklist

- [x] run `make test`

Fixes ENG-570, ENG-595, ENG-395

New and improved docs will be added in #319


packages/swr/src/hooks.ts
```diff
@@ -153,80 +203,184 @@ export const createHooks = <Operations extends OperationsDefinition>(client: Cli
...
+
+		const startedAtRef = useRef<number | null>(null);
+
+		const [state, setState] = useState({
+			isLoading: false,
+			isSubscribed: false,
+		});
+
+		useEffect(() => {
+			if (!startedAtRef.current && resetOnMount) {
```
Pagebakers, line 266: Allow  to reset the cached data when the hook in mounted, so we don't have 'old' subscription values.

@StarpTech what do you think of the `resetOnMount` prop name?

StarpTech: I like the name.

packages/sdk/src/client/client.ts
```diff
@@ -247,23 +247,31 @@ export class Client {
 		subscription: SubscriptionRequestOptions,
```
StarpTech, line 247: Don't forget to add tests for the types https://github.com/wundergraph/wundergraph/blob/main/packages/type-tests/test/swr.test-d.ts. It's a great place to understand the API of the new client.

StarpTech (changes requested):

packages/swr/src/hooks.ts
```diff
@@ -153,80 +203,184 @@ export const createHooks = <Operations extends OperationsDefinition>(client: Cli
...
-	>(
-		options: UseSubscriptionOptions<OperationName, Input>
-	): SWRResponse<Data, GraphQLResponseError> => {
-		const { enabled = true, operationName, input, subscribeOnce, ...swrConfig } = options;
-		const key = { operationName, input };
-		const _key = serialize(key);
-		const response = useSWR(enabled ? key : null, null, swrConfig);
+	// Helper hook used in useQuery and useSubscription
+	const useSubscribeTo = (props: UseSubscribeToProps) => {
+		const { mutate } = useSWRConfig();
```
StarpTech, line 254: Does that mean the user has to use `swr` context?

packages/swr/src/hooks.ts
```diff
@@ -1,38 +1,25 @@
-import useSWR, { mutate, SWRConfiguration, SWRResponse, MutatorOptions } from 'swr';
+import useSWR, { SWRConfiguration, useSWRConfig } from 'swr';
+import useSWRMutation, { SWRMutationConfiguration } from 'swr/mutation';
```
StarpTech, line 2: Can we set up a minimal test suite? We could use swr tests as an inspiration.

StarpTech (changes requested):

packages/swr/src/hooks.ts
Pagebakers: No, SWR always fallback to global context by default.

packages/swr/src/hooks.ts
Pagebakers: Yes I think it's a good idea, will look into this today.

packages/swr/src/hooks.ts
StarpTech: ok

packages/sdk/src/client/client.ts
Pagebakers: Fixed tests

packages/sdk/src/client/client.ts
```diff
@@ -247,23 +251,31 @@ export class Client {
...
+				wg_variables: this.stringifyInput(subscription.input),
+				wg_live: subscription?.liveQuery ? 'true' : 'false',
+				wg_sse: 'true',
+			});
+			const url = this.addUrlParams(this.operationUrl(subscription.operationName), params);
+			const eventSource = new EventSource(url, {
+				withCredentials: true,
+			});
+			eventSource.addEventListener('error', () => {
+				reject(new Error(`SSE connection error: ${subscription.operationName}`));
```
StarpTech, line 266: Can we log / enrich the original error? `(err)`

packages/swr/package.json
```diff
@@ -17,17 +18,27 @@
     "url": "git+https://github.com/wundergraph/wundergraph.git"
   },
   "dependencies": {
-    "swr": "^1.3.0"
+    "swr": "^2.0.0-rc.0"
   },
   "peerDependencies": {
     "@wundergraph/sdk": ">=0.91.0",
```
StarpTech, line 24: Update it?

packages/swr/src/hooks.test.tsx
```diff
@@ -0,0 +1,321 @@
...
+	};
+};
+
+describe('SWR - createHooks', () => {
+	const client = createClient();
+
+	const hooks = createHooks(client);
+
+	it('should be able to init hooks', async () => {
+		expect(hooks).toBeTruthy();
```
StarpTech, line 100: Can we validate here if `createHooks` returns the correct api?

packages/swr/src/types.ts
```diff
@@ -0,0 +1,161 @@
...
+}
+
+export interface SubscribeToOptions extends SubscriptionRequestOptions {
+	onResult(response: ClientResponse): void;
+	onSuccess?(response: ClientResponse): void;
+	onError?(error: GraphQLResponseError): void;
+	onAbort?(): void;
+}
+
+export interface UseUserOptions<User> extends FetchUserRequestOptions, SWRConfiguration<User> {
```
StarpTech, line 131: `UseUserHookOptions` ?

StarpTech (changes requested):

testapps/nextjs/.wundergraph/wundergraph.server.ts
```diff
@@ -86,6 +86,37 @@ export default configureWunderGraphServer<HooksConfig, InternalClient>(() => ({
 				},
 			},
 		},
+		subscriptions: {
+			// .wundergraph/operations/Counter.graphql
```
StarpTech, line 90: Delete?

StarpTech:
Please rename the PR title to something more meaningful for the user e.g improve mutation support.

packages/sdk/src/client/client.ts
Pagebakers: There was no valuable information available in the parameter, as it's just a generic event.

https://developer.mozilla.org/en-US/docs/Web/API/EventSource/error_event

StarpTech: ok

packages/swr/src/hooks.test.tsx
StarpTech: ok

StarpTech (approved):
LGTM

PR #306, opened 2022-10-31
feat: build hash from file sources in config

StarpTech:
#### Checklist

- [ ] run `make test`

### DIFF
--- a/packages/sdk/src/definition/index.ts
+++ b/packages/sdk/src/definition/index.ts
@@ -241,6 +241,11 @@ interface GraphQLIntrospectionOptions {
 	skipRenameRootFields?: string[];
 }
 
+// TODO: better name for "internal" ?
+export interface GraphQLIntrospectionInternal extends GraphQLIntrospection {
+	kind: 'graphql';
+}
+
 export interface GraphQLIntrospection extends GraphQLUpstream, GraphQLIntrospectionOptions {
 	isFederation?: boolean;
 }
@@ -249,6 +254,10 @@ export interface GraphQLFederationUpstream extends Omit<Omit<GraphQLUpstream, 'i
 	name?: string;
 	loadSchemaFromString?: GraphQLIntrospectionOptions['loadSchemaFromString'];
 	introspection?: GraphqlIntrospectionHeaders;
+}
+
+export interface GraphQLFederationIntrospectionInternal extends GraphQLFederationIntrospection {
+	kind: 'graphql-federation';
 }
 
 export interface GraphQLFederationIntrospection extends IntrospectionConfiguration {
@@ -261,6 +270,16 @@ export interface ReplaceJSONTypeFieldConfiguration {
 	fieldName: string;
 	inputTypeReplacement: string;
 	responseTypeReplacement: string;
+}
+
+export type IntrospectionUnion =
+	| OpenAPIIntrospectionInternal
+	| GraphQLIntrospectionInternal
+	| GraphQLFederationIntrospectionInternal
+	| DatabaseIntrospectionInternal;
+
+export interface DatabaseIntrospectionInternal extends DatabaseIntrospection {
+	kind: 'database';
 }
 
 export interface DatabaseIntrospection extends IntrospectionConfiguration {
@@ -364,6 +383,10 @@ export type OpenAPIIntrospectionSource =
 	| OpenAPIIntrospectionFile
 	| OpenAPIIntrospectionString
 	| OpenAPIIntrospectionObject;
+
+export interface OpenAPIIntrospectionInternal extends OpenAPIIntrospection {
+	kind: 'openapi';
+}
 
 export interface OpenAPIIntrospection extends HTTPUpstream {
 	source: OpenAPIIntrospectionSource;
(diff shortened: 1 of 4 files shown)


packages/sdk/src/definition/index.ts
```diff
@@ -241,6 +241,11 @@ interface GraphQLIntrospectionOptions {
 	skipRenameRootFields?: string[];
 }
 
+// TODO: better name for "internal" ?
```
StarpTech, line 244: Hint

packages/sdk/src/definition/introspection-cache.ts
```diff
@@ -124,12 +125,18 @@ export const introspectInInterval = async <Introspection extends IntrospectionCo
...
-export const introspectWithCache = async <Introspection extends IntrospectionConfiguration, A extends ApiType>(
+export const introspectWithCache = async <Introspection extends IntrospectionUnion, A extends ApiType>(
 	introspection: Introspection,
 	generator: (introspection: Introspection) => Promise<Api<A>>
 ): Promise<Api<A>> => {
 	const cacheKey = objectHash(introspection);
 
+	if (introspection.kind == 'openapi') {
+		if (introspection.source.kind === 'file') {
+			// TODO: read file, hash file and combine it with cacheKey
```
StarpTech, line 136: @fiam, here you can handle all properties pointing to external files and respect them in the final hash.

fiam:
@StarpTech Please, don't mix formatting changes with actual code changes, it makes telling what's actually changed more complicated.

If we wanted to go with this approach, I would instead make `introspectWithCache()` take an additional optional argument that indicates which files to consider for hashing, then make `introspect.openapi()` pass those. That way we remove all the plumbing for wrapping other introspection types.

However, as we discussed this morning, this simplest approach might be to skip the cache altogether for openapi data sources. What do you think of something like this?

```diff
diff --git a/packages/sdk/src/definition/index.ts b/packages/sdk/src/definition/index.ts
index 83232bc..9df9603 100644
--- a/packages/sdk/src/definition/index.ts
+++ b/packages/sdk/src/definition/index.ts
@@ -603,11 +603,10 @@ export const introspect = {
                        return new MongoDBApi(schema, dataSources, fields, types, interpolateVariableDefinitionAsJSON);
                }),
        federation: introspectFederation,
-       openApi: async (introspection: OpenAPIIntrospection): Promise<RESTApi> =>
-               introspectWithCache(introspection, async (introspection: OpenAPIIntrospection): Promise<RESTApi> => {
-                       const spec = loadOpenApi(introspection);
-                       return await openApiSpecificationToRESTApiObject(spec, introspection);
-               }),
+       openApi: async (introspection: OpenAPIIntrospection): Promise<RESTApi> => {
+               const spec = loadOpenApi(introspection);
+               return await openApiSpecificationToRESTApiObject(spec, introspection);
+       },
 };
 ```

We lose support for polling by doing that, but maybe the correct way handle schemas from static files is to make `wunderctl` watch them. What about watching `.wundergraph` and excluding `generated`/ `cache`? That would also remove a lot of complexity from the watcher.

StarpTech:
I didn't format the code without reasons. It's reformatted based on my code changes.

We should not remove the polling functionality for openapi. The easiest approach for now is to skip the cache for openapi with file as a source and regernerate the config.

fiam:
> I didn't format the code without reasons. It's reformatted based on my code changes.
> 
> We should not remove the polling functionality for openapi. The easiest approach for now is to skip the cache for openapi with file as a source and regernerate the config.

You're right, my bad. There are a lot of functions that are reformatted because they're indented to one more level. I hadn't noticed that.

About skipping the cache only when the source is a file, wouldn't it be simpler to do it this way than with the wrapper types? Something like e.g.

```diff
diff --git a/packages/sdk/src/definition/index.ts b/packages/sdk/src/definition/index.ts
index 83232bc..8843bde 100644
--- a/packages/sdk/src/definition/index.ts
+++ b/packages/sdk/src/definition/index.ts
@@ -603,11 +603,16 @@ export const introspect = {
                        return new MongoDBApi(schema, dataSources, fields, types, interpolateVariableDefinitionAsJSON);
                }),
        federation: introspectFederation,
-       openApi: async (introspection: OpenAPIIntrospection): Promise<RESTApi> =>
-               introspectWithCache(introspection, async (introspection: OpenAPIIntrospection): Promise<RESTApi> => {
+       openApi: async (introspection: OpenAPIIntrospection): Promise<RESTApi> => {
+               const generator = async (introspection: OpenAPIIntrospection): Promise<RESTApi> => {
                        const spec = loadOpenApi(introspection);
                        return await openApiSpecificationToRESTApiObject(spec, introspection);
-               }),
+               };
+               if (introspection.source.kind === 'file') {
+                       return generator(introspection);
+               }
+               return introspectWithCache(introspection, generator);
+       },
 };
 ```

StarpTech:
> About skipping the cache only when the source is a file, wouldn't it be simpler to do it this way than with the wrapper types? Something like e.g.

Yes, my example should outline the approach to identifying different introspection types in a type-safe way.

PR #309, opened 2022-11-02
fix: skip introspection cache when OpenAPI source is a file

fiam:
This prevents the cache from getting in the way when the underlying file
changes. Since all the data to generate the introspection is local, we
don't need the cache to able to introspect without hitting the network.

### DIFF
--- a/packages/sdk/src/definition/index.ts
+++ b/packages/sdk/src/definition/index.ts
@@ -603,11 +603,16 @@ export const introspect = {
 			return new MongoDBApi(schema, dataSources, fields, types, interpolateVariableDefinitionAsJSON);
 		}),
 	federation: introspectFederation,
-	openApi: async (introspection: OpenAPIIntrospection): Promise<RESTApi> =>
-		introspectWithCache(introspection, async (introspection: OpenAPIIntrospection): Promise<RESTApi> => {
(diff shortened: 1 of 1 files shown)


packages/sdk/src/definition/index.ts
```diff
@@ -603,11 +603,16 @@ export const introspect = {
...
 	federation: introspectFederation,
-	openApi: async (introspection: OpenAPIIntrospection): Promise<RESTApi> =>
-		introspectWithCache(introspection, async (introspection: OpenAPIIntrospection): Promise<RESTApi> => {
+	openApi: async (introspection: OpenAPIIntrospection): Promise<RESTApi> => {
+		const generator = async (introspection: OpenAPIIntrospection): Promise<RESTApi> => {
 			const spec = loadOpenApi(introspection);
 			return await openApiSpecificationToRESTApiObject(spec, introspection);
-		}),
+		};
+		if (introspection.source.kind === 'file') {
```
StarpTech, line 611: Please add a comment on why we do this.

StarpTech (changes requested):

packages/sdk/src/definition/index.ts
fiam: Good idea. I've added a comment explaining the rationale.

StarpTech (approved):
LGTM

PR #310, opened 2022-11-02
feat: disable response caching when returning HTTP errors

fiam:
- feat: add cachecontrol package
- refactor: use cachecontrol package to set headers
- feat: add httperror package
- feat: make all returned HTTP errors  use httperror package
- feat: add middleware to make all responses default to use no cache

### DIFF
--- a/pkg/apihandler/apihandler.go
+++ b/pkg/apihandler/apihandler.go
@@ -47,6 +47,7 @@ import (
 	"github.com/wundergraph/wundergraph/pkg/engineconfigloader"
 	"github.com/wundergraph/wundergraph/pkg/graphiql"
 	"github.com/wundergraph/wundergraph/pkg/hooks"
+	"github.com/wundergraph/wundergraph/pkg/httperror"
 	"github.com/wundergraph/wundergraph/pkg/inputvariables"
 	"github.com/wundergraph/wundergraph/pkg/interpolate"
 	"github.com/wundergraph/wundergraph/pkg/loadvariable"
@@ -200,7 +201,7 @@ func (r *Builder) BuildAndMountApiHandler(ctx context.Context, router *mux.Route
 						return
 					}
 				}
-				http.Error(w, fmt.Sprintf("Host not found: %s", r.Host), http.StatusNotFound)
+				httperror.Error(w, fmt.Sprintf("Host not found: %s", r.Host), http.StatusNotFound)
 			})
 		})
 	}
@@ -697,7 +698,7 @@ func (h *GraphQLHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	defer pool.PutBytesBuffer(buf)
 	_, err := io.Copy(buf, r.Body)
 	if err != nil {
-		http.Error(w, "bad request", http.StatusBadRequest)
+		httperror.Error(w, "bad request", http.StatusBadRequest)
 		return
 	}
 
@@ -726,7 +727,7 @@ func (h *GraphQLHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	shared.Parser.Parse(shared.Doc, shared.Report)
 
 	if shared.Report.HasErrors() {
-		http.Error(w, "bad request", http.StatusBadRequest)
+		httperror.Error(w, "bad request", http.StatusBadRequest)
 		return
 	}
 
@@ -734,11 +735,11 @@ func (h *GraphQLHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	err = shared.Printer.Print(shared.Doc, h.definition, shared.Hash)
 	if err != nil {
-		http.Error(w, "bad request", http.StatusBadRequest)
+		httperror.Err(w, err, "")
 		return
 	}
 	if shared.Report.HasErrors() {
-		http.Error(w, "bad request", http.StatusBadRequest)
+		httperror.Error(w, "bad request", http.StatusBadRequest)
 		return
 	}
 
@@ -772,7 +773,7 @@ func (h *GraphQLHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 				return
 			}
 			h.log.Error("ResolveGraphQLResponse", abstractlogger.Error(err))
-			http.Error(w, "bad request", http.StatusBadRequest)
+			httperror.Err(w, err, "")
 			return
 		}
 		_, err = executionBuf.WriteTo(w)
@@ -783,7 +784,7 @@ func (h *GraphQLHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	case *plan.SubscriptionResponsePlan:
 		flushWriter, ok := getFlushWriter(shared.Ctx, r, w)
 		if !ok {
-			http.Error(w, "Connection not flushable", http.StatusBadRequest)
+			httperror.Error(w, "Connection not flushable", http.StatusBadRequest)
 			return
 		}
 
@@ -796,7 +797,7 @@ func (h *GraphQLHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 	case *plan.StreamingResponsePlan:
-		http.Error(w, "not implemented", http.StatusNotFound)
+		httperror.Error(w, "not implemented", http.StatusNotFound)
 	}
 }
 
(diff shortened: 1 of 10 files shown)


pkg/authentication/authentication.go
```diff
@@ -740,7 +742,8 @@ func (u *CookieUserHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	w.Header().Set("Content-Type", "application/json")
 	w.Header()["ETag"] = []string{user.ETag}
-	w.Header().Set("Cache-Control", "private, max-age=0, stale-while-revalidate=60")
+	var h cachecontrol.Header
+	h.Private().MaxAge(0).StaleWhileRevalidate(60).Set(w)
```
StarpTech, line 746: Why setting SWR when the content is not cached at all?

pkg/cachecontrol/cachecontrol.go
```diff
@@ -0,0 +1,100 @@
...
+func (h *Header) Public() *Header {
+	return h.filterOut("private").append("public")
+
+}
+
+func (h *Header) Private() *Header {
+	return h.filterOut("public").append("private")
+}
+
+// PublicOrPrivate is a shorthand for calling Public() iff
```
StarpTech, line 76: "iff"

pkg/cachecontrol/cachecontrol.go
```diff
@@ -0,0 +1,100 @@
...
+func (h *Header) Middleware() func(next http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			h.Set(w)
+			next.ServeHTTP(w, r)
+		})
+	}
+}
+
+func (h *Header) Disabled() *Header {
```
StarpTech, line 62: `Disabled` sounds unspecific. Maybe `DisableCache`, `NoCache` ?

pkg/cachecontrol/cachecontrol.go
```diff
@@ -0,0 +1,100 @@
...
+	"strings"
+)
+
+// See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control for
+// information on the different cache values. Not all
+
+// Header represents a series of values for a Cache-Control header.
+// The zero value is a valid an empty Cache-Control header. To set the
+// header use String(), Set() or Middleware().
+type Header struct {
```
StarpTech, line 16: Honestly, I don't see value in providing a header builder for just a few constants. Just use `fmt.Sprintf`.

StarpTech (changes requested):

pkg/cachecontrol/cachecontrol.go
fiam: https://www.merriam-webster.com/dictionary/iff

pkg/apihandler/apihandler.go
```diff
@@ -1123,11 +1129,11 @@ func (h *QueryHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 			w.Header()["ETag"] = []string{ETag}
 
-			if h.cacheConfig.public {
-				w.Header().Set("Cache-Control", fmt.Sprintf("public, max-age=%d, stale-while-revalidate=%d", h.cacheConfig.maxAge, h.cacheConfig.staleWhileRevalidate))
-			} else {
-				w.Header().Set("Cache-Control", fmt.Sprintf("private, max-age=%d, stale-while-revalidate=%d", h.cacheConfig.maxAge, h.cacheConfig.staleWhileRevalidate))
-			}
+			var cch cachecontrol.Header
```
jensneuse, line 1132: Wouldn't it make sense to build the CC headers when the Handler is configured?
It looks like this config is static in regards of the request, so I think we can do it at configuration time instead of request time for each request.

jensneuse (changes requested):

pkg/cachecontrol/cachecontrol.go
fiam: Using `fmt.Sprintf` can lead to typos, copy pasting and repetition. See the headers that I replaced that were identical except for `public / private`. Also, if we want to make changes to the headers, having them defined in a central place makes it easier.

StarpTech: A builder only makes sense if we have something complex to "build". We could provide a simple function for 2 use cases and are done.

- No cache
- Cache

pkg/authentication/authentication.go
fiam: It's actually cached. Basically, if the response is less than 60 seconds old, you're allowing the browser to reuse whatever cached response it has while at the same time asking it to refresh the cached data from the network in a non-blocking way.

pkg/apihandler/apihandler.go
fiam: Good idea. We can set it up during startup and wrap the handler into the middleware. Error responses with explicit headers will override the header for responses that should not be cached.

pkg/cachecontrol/cachecontrol.go
fiam: Cache responses are determined based on what the user configures via the SDK, so we need to keep some state. For now we only have two configurable values, but expecting it to grow doesn't seem unreasonable to me. If fact, if we have a more robust way than handling them on the server side than manipulating string, we could allow more flexibility.

Regarding unit testing the cache values, what ends up happening with long strings is that people mistype/misremember the string in the code and then copy it into the test. Let's say you have `%s, max-age=%d, state-while-revalidate=%d`, then someone copies it into the test and checks that passing `(true, 10, 20)` ends up producing `public, max-age=10, state-while-revalidate=20` which will pass the test but end up with an invalid header.

On the other hand, having just 2 free functions to set the cache leaves the call sites quite ambiguous. If you see a call to `cachecontrol.EnableCache(true, 30, 30)` is not clear what it does.

With that said, if you want it with free functions I will change it to do so.

pkg/apihandler/apihandler.go
fiam: Looking at it again, there are a lot of paths in that function and we don't have robust tests for it. Changing for a function based approach, there's no much to be gained from a performance perspective either. So I think I would leave the logic as-is.

pkg/cachecontrol/cachecontrol.go
StarpTech: Just go with a reasonable approach. I leave let it up to you to use just constants or functions.

pkg/cachecontrol/cachecontrol.go
StarpTech: Never heard :smile:

pkg/authentication/authentication.go
StarpTech: No, SWR means you can work with stale data for a further 60s until the data is invalidated asynchronously, but we have never put the data into the cache. It makes no sense to me.

> stale-while-revalidate tells caches that they may continue to serve a response after it becomes stale for up to the specified number of seconds, provided that they work asynchronously in the background to fetch a new one.

StarpTech: I'd even say for auth I'd disable cache.

jensneuse: max-age=0 combined with stale-while-revalidate=60 will result in the following behavior:

After 0 seconds (max-age,immediately), the browser will read the user object from the HTTP cache for 60 seconds.
Each time the browser reads the cache, it also fires off a request to the server to revalidate the cache.
Combined with the ETag, the server will most likely return with a 304.

This means, we can efficiently revalidate if the user is still logged in on each browser focus event, while most of the time, the answer is just a 304 not modified.

Once we logout, we delete the in-memory cache of the browser client and fire off a get user request, which will invalidate the http cache.

I think this is the most efficient pattern to leverage the HTTP cache and keep the client in sync across tabs and windows.

StarpTech: Thanks, both. Can we document this in the form of comment?

pkg/authentication/authentication.go
```diff
@@ -740,7 +742,7 @@ func (u *CookieUserHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	w.Header().Set("Content-Type", "application/json")
 	w.Header()["ETag"] = []string{user.ETag}
-	w.Header().Set("Cache-Control", "private, max-age=0, stale-while-revalidate=60")
+	cachecontrol.EnableCache(w, false, 0, 60)
```
StarpTech, line 745: Sry, but this is hard to read with 4 arguments. Can we use a string here?

pkg/authentication/authentication.go
fiam: I've linked Jens response in a comment

pkg/authentication/authentication.go
fiam: Changed to new function called `EnableForAuthentication()`, as discussed

PR #313, opened 2022-11-03
fix: health check with retry client

thisisnithin:
#### Checklist

- [x] run `make test`

### DIFF
--- a/pkg/hooks/hooks.go
+++ b/pkg/hooks/hooks.go
@@ -212,10 +212,11 @@ func (c *Client) doRequest(ctx context.Context, action string, hook MiddlewareHo
 }
 
 func (c *Client) DoHealthCheckRequest(timeout time.Duration) (status bool) {
-	client := http.DefaultClient
-	client.Timeout = timeout
+	retryClient := retryablehttp.NewClient()
+	retryClient.RetryMax = 3
+	retryClient.HTTPClient.Timeout = timeout
 
-	resp, err := client.Get(fmt.Sprintf("%s/health", c.serverUrl))
+	resp, err := retryClient.Get(fmt.Sprintf("%s/health", c.serverUrl))
 	if err != nil || resp.StatusCode != 200 {
 		return
 	}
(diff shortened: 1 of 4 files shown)


pkg/hooks/hooks.go
```diff
@@ -212,10 +212,11 @@ func (c *Client) doRequest(ctx context.Context, action string, hook MiddlewareHo
 }
 
 func (c *Client) DoHealthCheckRequest(timeout time.Duration) (status bool) {
-	client := http.DefaultClient
-	client.Timeout = timeout
+	retryClient := retryablehttp.NewClient()
```
StarpTech, line 215: Please create a client once when the client is created. We can use this client for the client of the whole hook.

pkg/hooks/hooks.go
```diff
@@ -212,10 +212,11 @@ func (c *Client) doRequest(ctx context.Context, action string, hook MiddlewareHo
...
+	retryClient.RetryMax = 3
+	retryClient.HTTPClient.Timeout = timeout
```
StarpTech, line 217: set retryClient.logger = null to avoid logging with the default library logger. In the long term, we should revisit those places and implement an adapter to our logger.

pkg/node/nodetemplates/resources/index.html
```diff
@@ -48,6 +48,10 @@ <h2 class="font-medium text-gray-600 text-xl">Powered by</h2>
 			>
```
StarpTech, line 48: Let's also rename "DEATH" to "INIT" for intialiazing

StarpTech (changes requested):

pkg/hooks/hooks.go
thisisnithin: used the existing client

pkg/hooks/hooks.go
thisisnithin: done

pkg/node/nodetemplates/resources/index.html
thisisnithin: renamed it to Ready/Not Ready

PR #318, opened 2022-11-07
chore: improve error handling

jensneuse:
This PR gracefully handles errors caused by canceled contexts when the client disconnects.

### DIFF
--- a/pkg/apihandler/apihandler.go
+++ b/pkg/apihandler/apihandler.go
@@ -1086,6 +1086,10 @@ func (h *QueryHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		hookData := hookBaseData(r, hookBuf.Bytes(), ctx.Variables, nil)
 		out, err := h.hooksClient.DoOperationRequest(ctx.Context, h.operation.Name, hooks.MockResolve, hookData)
 		if err != nil {
+			if ctx.Err() != nil {
+				// context was canceled (e.g. client disconnected)
+				return
+			}
 			h.log.Error("MockResolve query hook", abstractlogger.Error(err), abstractlogger.Bool("isLive", isLive))
 			http.Error(w, "MockResolve query hook failed", http.StatusInternalServerError)
 			return
@@ -1153,6 +1157,10 @@ func (h *QueryHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		hookData := hookBaseData(r, hookBuf.Bytes(), ctx.Variables, nil)
 		out, err := h.hooksClient.DoOperationRequest(ctx.Context, h.operation.Name, hooks.PreResolve, hookData)
 		if err != nil {
+			if ctx.Err() != nil {
+				// context was canceled (e.g. client disconnected)
+				return
+			}
 			h.log.Error("PreResolve query hook failed", abstractlogger.Error(err))
 			http.Error(w, "PreResolve query hook failed", http.StatusInternalServerError)
 			return
@@ -1170,6 +1178,10 @@ func (h *QueryHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		hookData := hookBaseData(r, hookBuf.Bytes(), ctx.Variables, nil)
 		out, err := h.hooksClient.DoOperationRequest(ctx.Context, h.operation.Name, hooks.MutatingPreResolve, hookData)
 		if err != nil {
+			if ctx.Err() != nil {
+				// context was canceled (e.g. client disconnected)
+				return
+			}
 			h.log.Error("MutatingPreResolve query hook failed", abstractlogger.Error(err))
 			http.Error(w, "MutatingPreResolve query hook failed", http.StatusInternalServerError)
 			return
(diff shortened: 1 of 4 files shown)


pkg/apihandler/apihandler.go
```diff
@@ -1086,6 +1086,10 @@ func (h *QueryHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		hookData := hookBaseData(r, hookBuf.Bytes(), ctx.Variables, nil)
 		out, err := h.hooksClient.DoOperationRequest(ctx.Context, h.operation.Name, hooks.MockResolve, hookData)
 		if err != nil {
+			if ctx.Err() != nil {
```
fiam, line 1089: Maybe checking `err` with `errors.Is` would be more robust and obviate the comment?

pkg/apihandler/apihandler.go
```diff
@@ -1743,6 +1787,10 @@ func (h *SubscriptionHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
 		hookData := hookBaseData(r, hookBuf.Bytes(), ctx.Variables, nil)
 		out, err := h.hooksClient.DoOperationRequest(ctx.Context, h.operation.Name, hooks.MutatingPreResolve, hookData)
 		if err != nil {
+			if ctx.Err() != nil {
```
StarpTech, line 1790: I think it's not correct to swallow the error from the request in case of context cancellation.

StarpTech (changes requested):

pkg/apihandler/apihandler.go
jensneuse: There can be multiple causes that cancel the ctx. Contexts can be "stacked".
What matter is that the ctx was cancelled, which results in ctx.Err() not being nil.

pkg/apihandler/apihandler.go
jensneuse: I think it is. The client disconnected. It's a regular operation that is "propagated" through context cancellation.
The purpose of the task was to get rid of false negatives in the logs. (Error logs that should be no logs at all, as it's caused by client disconnected)

pkg/apihandler/apihandler.go
jensneuse: So, checking errors.Is is actually less robust as there will be many cases where it will not match.

fiam: If you cancel the parent then all of the stack should be cancelled. Checking for something like `errors.Is(err, context.Canceled)` should work and help with self-documentation.

pkg/apihandler/apihandler.go
jensneuse: I think you're misunderstanding this code. if ctx.Err() is not nil, something cancelled the operation (e.g. the client, a timeout, etc..) The operation by itself did not return any error, it was the http client cancellation by the ctx.

fiam: I think early returning is fine in these. If we want to log client disconnects, we can do that at the top level in a single place (e.g. with middleware)

YuriBuerov (approved):

pkg/apihandler/apihandler.go
StarpTech: If the request fails, it is still an error. This is my proposal:

```go
if err != nil {
    h.log.Error("MutatingPostResolve mutation hook", abstractlogger.Error(err))
	http.Error(w, "MutatingPostResolve mutation hook failed", http.StatusInternalServerError)
   return
}
if ctx.Err() != nil {
    h.log.Debug("MutatingPostResolve mutation hook", abstractlogger.Error(ctx.Err()))
	http.Error(w, "MutatingPostResolve mutation hook failed", http.StatusInternalServerError)
    return
}
```

StarpTech (changes requested):

pkg/apihandler/apihandler.go
jensneuse: I understand the intention, but this code will never reach `if ctx.Err()` because err is never nil if ctx.Err() is != nil

StarpTech: Are you sure?. I thought someone could transport different errors through ctx and return value.

pkg/apihandler/apihandler.go
```diff
@@ -1414,12 +1339,16 @@ func (h *QueryHandler) handleLiveQuery(r *http.Request, w http.ResponseWriter, c
 		var hookError bool
 		response, err := h.handleLiveQueryEvent(ctx, r, requestBuf, hookBuf)
 		if err != nil {
-			if ctx.Err() != nil {
+			if errors.Is(err, context.Canceled) {
 				// context was canceled (e.g. client disconnected)
+				// we've already flushed a header, so we simply return
```
StarpTech, line 1344: Where did we already flush the header? Does it return 499 to the client?

StarpTech (changes requested):

pkg/apihandler/apihandler.go
jensneuse: In line 1077

jensneuse: You cannot return a header when we flushed it already.

StarpTech: ok

StarpTech (approved):
LGTM

fiam (approved):
LGTM

PR #323, opened 2022-11-08
feat: logging, request id

YuriBuerov:
#### Checklist

- [ ] run `make test`


cli/commands/generate.go
```diff
@@ -70,8 +70,8 @@ var generateCmd = &cobra.Command{
 			err := configRunner.Stop()
 			if err != nil {
 				log.Error("Stopping runner failed",
-					abstractlogger.String("runnerName", "config-runner"),
```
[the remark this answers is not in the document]
YuriBuerov: correct

pkg/apihandler/apihandler.go
```diff
@@ -61,8 +61,12 @@ const (
 	WG_LIVE         = WG_PREFIX + "live"
 	WG_VARIABLES    = WG_PREFIX + "variables"
 	WG_CACHE_HEADER = "X-Wg-Cache"
+
+	requestID = "requestID"
 )
 
+type requestIDKey struct{}
```
StarpTech, line 68: I'd favor `type requestIDKey string`

```
k := favContextKey("language")
ctx := context.WithValue(request.Context(), k, requestID)
```

StarpTech (changes requested):

pkg/apihandler/apihandler.go
YuriBuerov: why? https://pkg.go.dev/context#WithValue

StarpTech: Because it is more explicit and easier to debug. You use a unique struct as an identifier although a string is sufficient.  In the example of `WithValue` https://pkg.go.dev/context#WithValue they use the same approach.

StarpTech: Check out the example. `type requestIDKey string` is not of type string.

jensneuse (changes requested):
Nice and clean implementation of injecting the request ID into the context.

3 things I'd add:
1. inject the request ID into hook middleware requests
2. inject the request ID into internal requests (when a hook uses the internal client)
3. inject the request ID into origin requests via the transport (the ctx automatically injects the request ID into the transport, so we can pick it up in the transport and inject it into origin http requests)

pkg/node/node.go
```diff
@@ -347,10 +354,12 @@ func (n *Node) GetHealthReport(hooksClient *hooks.Client) (*HealthCheckReport, b
...
 	logLevel := nodeConfig.Api.Options.Logging.Level
 	if n.options.enableDebugMode {
-		logLevel = abstractlogger.DebugLevel
+		logLevel = zapcore.DebugLevel
 	}
 
-	n.log = abstractlogger.NewZapLogger(logging.Zap().With(zap.String("component", "@wundergraph/node")), logLevel)
+	n.log = logging.
+		New(n.options.prettyLogging, n.options.enableDebugMode, logLevel).
+		With(zap.String("component", "@wundergraph/node"))
```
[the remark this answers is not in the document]
StarpTech: Yes. We call it `wundernode` in the cloud to differentiate between a node with a running hooks server and without.

StarpTech (approved):
LGTM

packages/sdk/src/middleware/server.ts
```diff
@@ -162,6 +163,9 @@ export const createServer = async ({
 
 	const fastify = Fastify({
 		logger,
+		genReqId: (req) => {
+			return req.headers['x-request-id']?.toString() || uuidv4();
```
StarpTech, line 167: We don't need to fall back here because the hooks server can't be called alone.

pkg/authentication/authentication.go
```diff
@@ -304,6 +304,9 @@ func (h *Hooks) handleMutatingPostAuthentication(ctx context.Context, user User)
 	}
 	out, err := h.Client.DoAuthenticationRequest(ctx, hooks.MutatingPostAuthentication, hookData)
 	if err != nil {
+		if ctx.Err() != nil {
```
StarpTech, line 307: @jensneuse

pkg/hooks/hooks.go
```diff
@@ -184,15 +185,22 @@ func (c *Client) doRequest(ctx context.Context, action string, hook MiddlewareHo
 	if err != nil {
 		return nil, err
 	}
+
 	r.Header.Set("Content-Type", "application/json")
+	if requestID := logging.RequestIDFromContext(ctx); requestID != "" {
```
StarpTech, line 190: It should never happen that we have no request id. I'd log a warning here.

StarpTech (changes requested):

pkg/apihandler/apitransport.go
```diff
@@ -116,6 +117,7 @@ type Claims struct {
 }
 
 func (t *ApiTransport) RoundTrip(request *http.Request) (*http.Response, error) {
+	request.Header.Set("X-Request-Id", logging.RequestIDFromContext(request.Context()))
```
StarpTech, line 120: Missing `!= ""` check with warning log.

pkg/scriptrunner/scriptrunner.go
```diff
@@ -141,40 +141,40 @@ func (b *ScriptRunner) Run(ctx context.Context) chan struct{} {
...
 			} else {
 				b.log.Debug("Script runner is done",
-					abstractlogger.String("runnerName", b.name),
-					abstractlogger.Int("exit", status.Exit),
-					abstractlogger.Any("startTs", status.StartTs),
-					abstractlogger.Any("stopTs", status.StopTs),
-					abstractlogger.Bool("complete", status.Complete),
+					zap.String("runnerName", b.name),
+					zap.Int("exit", status.Exit),
+					zap.Any("startTs", status.StartTs),
```
StarpTech, line 166: I think zap has a concrete type for this.

StarpTech (changes requested):

packages/sdk/src/middleware/server.ts
YuriBuerov: `return req.headers['x-request-id']!.toString()` ?

pkg/authentication/authentication.go
jensneuse: @YuriBuerov we should use the same pattern like here:
https://github.com/wundergraph/wundergraph/blob/68b34d7b5436dfe8d88c7e3a27e0975a263e5034/pkg/apihandler/apihandler.go#L2243

YuriBuerov: thanks, I'll fix it, I didn't change it though

jensneuse: Yes, I know. I forgot to change it.

pkg/apihandler/apitransport.go
YuriBuerov: @jensneuse  where should we place it? `logging` package?

jensneuse: sounds good

YuriBuerov: and what is our final decision regarding logging warning in case if req id is empty? (my comment above)

jensneuse: How about adding a test that ensures this behaviour is actually impossible? 
A test like this:
https://github.com/wundergraph/wundergraph/blob/ad02a27af5dddbcf8e1126e5bf32c949005630b2/pkg/node/node_test.go#L154

Create a mock origin. Send a request to the gateway (including request ID), expect request ID to be sent to mock origin.

YuriBuerov:
> @YuriBuerov did you test that the request-id is correctly propagated from Node -> hook -> another-hook -> onOriginResponse -> postHook ?

@StarpTech  yes

StarpTech (approved):
LGTM

jensneuse (approved):

PR #328, opened 2022-11-10
feat: remove /<app-name>/main prefix from router and pb definition

fiam:
Some minor changes to the SDK are also included, to accomodate for the
missing fields in the types generated from .proto


fiam:
Superseded by https://github.com/wundergraph/wundergraph/pull/335

PR #330, opened 2022-11-10
fix: make baseURL optional

Pagebakers:
The problem was that config is optional, but baseURL was required when adding a config, while baseURL has a default value. So when you add a customFetch for usage in nodejs, you also had to add baseURL which is incorrect.


fiam (approved):
LGTM

jensneuse (approved):

PR #338, opened 2022-11-15
chore: limit CI runs to 20 minutes

fiam:


StarpTech:
@fiam could you add this to all workflows? I think 20min is a good default.

fiam:
@StarpTech Sure

StarpTech (approved):
LGTM

PR #342, opened 2022-11-17
feat: add better error messages for input validation

jensneuse:
This PR adds better error messages in case input validation fails. This gives useful information to the user of the API to send the correct inputs.

### DIFF
--- a/pkg/inputvariables/validator.go
+++ b/pkg/inputvariables/validator.go
@@ -3,34 +3,66 @@ package inputvariables
 import (
 	"context"
 	"encoding/json"
-	"fmt"
+	"io"
+	"net/http"
 
 	"github.com/qri-io/jsonschema"
 )
 
 type Validator struct {
-	schema          jsonschema.Schema
-	schemaString    string
-	enableDebugMode bool
+	schema               jsonschema.Schema
+	schemaString         string
+	disableVerboseErrors bool
 }
 
-func NewValidator(schema string, enableDebugMode bool) (*Validator, error) {
+func NewValidator(schema string, disableVerboseErrors bool) (*Validator, error) {
 	var validator Validator
 	err := json.Unmarshal([]byte(schema), &validator.schema)
 	if err != nil {
 		return nil, err
 	}
 	validator.schemaString = schema
-	validator.enableDebugMode = enableDebugMode
+	validator.disableVerboseErrors = disableVerboseErrors
 	return &validator, nil
 }
 
-func (v *Validator) Validate(ctx context.Context, variables []byte) bool {
+type ValidationError struct {
+	Message string
+	Input   json.RawMessage
+	Errors  []jsonschema.KeyError
+}
+
+func (v *Validator) Validate(ctx context.Context, variables []byte, errOut io.Writer) (valid bool) {
 	errs, err := v.schema.ValidateBytes(ctx, variables)
-	valid := err == nil && len(errs) == 0
-	if !valid && v.enableDebugMode {
-		errDump, _ := json.Marshal(errs)
-		fmt.Printf("\n\nVariable Validation: invalid\n\nJSON Schema:\n\n%s\n\nVariables:\n\n%s\n\nErrors:\n\n%s\n\n", v.schemaString, string(variables), string(errDump))
+	if err == nil && len(errs) == 0 {
+		return true
 	}
-	return valid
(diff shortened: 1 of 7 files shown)


pkg/inputvariables/validator.go
```diff
@@ -3,34 +3,66 @@ package inputvariables
...
 	errs, err := v.schema.ValidateBytes(ctx, variables)
-	valid := err == nil && len(errs) == 0
-	if !valid && v.enableDebugMode {
-		errDump, _ := json.Marshal(errs)
-		fmt.Printf("\n\nVariable Validation: invalid\n\nJSON Schema:\n\n%s\n\nVariables:\n\n%s\n\nErrors:\n\n%s\n\n", v.schemaString, string(variables), string(errDump))
+	if err == nil && len(errs) == 0 {
+		return true
+	}
+	if v.disableVerboseErrors {
+		_, _ = errOut.Write([]byte("Bad Request: Invalid input"))
```
fiam, line 41: It would be better to use `io.WriteString` instead of casting to `[]byte`, please.

pkg/inputvariables/validator.go
```diff
@@ -3,34 +3,66 @@ package inputvariables
...
+
+func (v *Validator) Validate(ctx context.Context, variables []byte, errOut io.Writer) (valid bool) {
 	errs, err := v.schema.ValidateBytes(ctx, variables)
-	valid := err == nil && len(errs) == 0
-	if !valid && v.enableDebugMode {
-		errDump, _ := json.Marshal(errs)
-		fmt.Printf("\n\nVariable Validation: invalid\n\nJSON Schema:\n\n%s\n\nVariables:\n\n%s\n\nErrors:\n\n%s\n\n", v.schemaString, string(variables), string(errDump))
+	if err == nil && len(errs) == 0 {
+		return true
+	}
```
fiam, line 39: Type-assert `errOut` for `http.ResponseWriter` and set an appropriate response code: e.g.

```go
if rw, ok := errOut.(http.ResponseWriter); ok {
    rw.WriteHeader(http.StatusBadRequest)
}
```

Then you can drop `ValidationWriter`

pkg/inputvariables/validator_test.go
```diff
@@ -17,29 +19,47 @@ func TestValidator_Validate(t *testing.T) {
...
+	out.Reset()
+	actual = validator.Validate(context.Background(), []byte(`{"id":true}`), out)
+	assert.Equal(t, false, actual)
+	assert.Equal(t, `{"Message":"Bad Request: Invalid input","Input":{"id":true},"Errors":[{"propertyPath":"/id","invalidValue":true,"message":"type should be string, got boolean"}]}
+`, out.String())
+
+	out.Reset()
+	actual = validator.Validate(context.Background(), []byte(`{}`), out)
+	assert.Equal(t, false, actual)
+	assert.Equal(t, `{"Message":"Bad Request: Invalid input","Input":{},"Errors":[{"propertyPath":"/","invalidValue":{},"message":"\"id\" value is required"}]}
```
Pagebakers, line 61: What do you think about using dot notation for the path? It will have better support with libraries like lodash but also form libs like react-hook-form out of the box.

instead of propertyPath '/' for input field 'id'

`"propertyPath": "id"`

or for objects

`"propertyPath": "user.id"`

```ts
import get from 'lodash/get'

const id = get({user: {id: 1}}, 'user.id')
```

or with hook form

```ts
const { setError } = useForm();

setError(propertyPath, {message}) // user.id
```

Pagebakers (changes requested):
Some remarks about the api.

pkg/inputvariables/validator_test.go
jensneuse: It's the default error format from the library we're using.
So far, we don't have any requirements for the use case you're mentioning.

In your case, you can do client-side validation using a generated validator with the json schema that we generate per operation.

fiam (changes requested):

pkg/inputvariables/validator.go
jensneuse: changed it

pkg/inputvariables/validator.go
jensneuse: I'd like to keep the validator itself independent from http.

Pagebakers (approved):
LGTM

PR #344, opened 2022-11-17
docs: advanced s3 spec

Pagebakers:
#### Checklist

- [ ] run `make test`


specs/advanced-s3-support.md
```diff
@@ -0,0 +1,177 @@
...
+This configuration can then be used for validation on the server and client side.
+
+In case no `uploads` configuration is added, upload behaves as the current implementation, where files are upload in the bucket root with their original filename.
+
+```ts
+configureWunderGraphApplication({
+  s3UploadProvider: [
+    new S3UploadProvider({
+      name: 'minio',
+      uploads: {
```
jensneuse, line 40: I'd call this uploadProfiles.

Btw. what happens if there's no upload profile at all?

specs/advanced-s3-support.md
```diff
@@ -0,0 +1,177 @@
...
+        avatar: {
+          maxAllowedUploadSizeBytes: 1024 * 1024 * 10, // 10 MB, optional, defaults to 25 MB
+          maxAllowedFiles: 1, // limit the number of files to 1, leave undefined for unlimited files
+          allowedMimeTypes: ['image/png', 'image/jpeg'], // wildcard is supported, e.g. 'image/*', leave empty/undefined to allow all
+          allowedFileExtensions: ['png', 'jpg'], // leave empty/undefined to allow all
+        },
+        coverPicture: {
+          meta: {
```
jensneuse, line 48: I'm not sure if this will work.
I think it would be easier to have something like zod to generate a JSON Schema, which can be used to generate the types for the wundergraph.server.ts.

jensneuse (changes requested):

specs/advanced-s3-support.md
Pagebakers: I like it.

When no profiles are configured, it will fallback to the default functionality as described.

specs/advanced-s3-support.md
Pagebakers: I didn't want to make a hard dependency on Zod yet since that would be an RFC on itself I think, but I would love to see it. The current meta object in the spec is a JSON schema defined as a js object, so it can be typed.

Pagebakers: We could create a helper for this that returns the JSON schema as a string. It could accept a simple object by default and it allows to use any other implementation like Zod.

```ts
meta: createMetaSchema({
  postId: 'string'
})
```

or we accept this as the default (I like the simplicity of this)

```ts
meta: {
   postId: 'string
}
```

And allow schema to be transformed from zod or other libs.

```ts
meta: metaSchemaFromZod(z.object({
  
}))
```

jensneuse: ok, makes sense

jensneuse (approved):

fiam (approved):
LGTM. I like the recap of the problems at the end, good job.

PR #345, opened 2022-11-17
feat: add WunderGraph testing library

fiam:
- feat: add wunderctlSubprocess
- feat: add initial testing framework
- feat: add tests to simple example


docs-website/src/pages/docs/guides/testing.md
```diff
@@ -0,0 +1,164 @@
+---
+title: Testing
+pageTitle: WunderGraph - Testing
+description: This guide shows how to test your WunderGraph applications
+---
+
+## Headless testing
+
+WunderGraph comes with builtin support for running tests against your APIs, using a our type-safe
```
Pagebakers, line 9: I think we can drop type-safe no? Or rephrase.

`fully typesafe using our TypeScript client.`

Pagebakers (changes requested):
Really great work! A few small comments.

packages/sdk/src/codegen/templates/typescript/client.template.ts
```diff
@@ -101,7 +100,7 @@ export class WunderGraphClient extends Client {
 	}
 }
 
-export const createClient = (config?: Partial<Omit<ClientConfig, PrivateConfigProperties>>) => {
+export const createClient = (config?: ClientConfigInit) => {
```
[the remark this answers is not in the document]
fiam: Good idea, I've changed it.

docs-website/src/pages/docs/guides/testing.md
fiam: I agree, I think it reads better if we drop `type-safe`

docs-website/src/pages/docs/guides/testing.md
```diff
@@ -0,0 +1,164 @@
...
+
+```typescript
+import { Server } from '@wundergraph/sdk/testing'
+import { createClient } from '../.wundergraph/generated/client'
+
+// Imports from Jest
+import { describe, expect, test } from '@jest/globals'
+```
+
+You can create as many `Server` instances as you like, they will run in parallel without interfering
```
StarpTech, line 55: I should rephrase it. Right now, there is no reason to start more than one, and you have to create multiple if you want to split your tests into multiple files.

packages/sdk/src/testing/index.ts
```diff
@@ -0,0 +1,262 @@
...
+			} catch (e: any) {
+				if (!(e instanceof Error && e.message.indexOf('SIGKILL') >= 0)) {
+					console.error(`error shutting down ${e}`);
+				}
+			}
+			this.subprocess = undefined;
+		}
+	}
+
+	private async beforeTest(): Promise<void> {
```
StarpTech, line 171: Just `start` and `stop`?

StarpTech (changes requested):

docs-website/src/pages/docs/guides/testing.md
```diff
@@ -0,0 +1,164 @@
...
+description: This guide shows how to test your WunderGraph applications
+---
+
+## Headless testing
+
+WunderGraph comes with builtin support for running tests against your APIs, using a our type-safe
+TypeScript client. Our `Server` implementation sets up your WunderGraph server, runs your tests,
+and finally tears everything down.
+
+Our testing library is framework agnostic, so you can use it coupled with your preferred test
```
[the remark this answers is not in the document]
fiam: Fixed

docs-website/src/pages/docs/guides/testing.md
```diff
@@ -0,0 +1,164 @@
...
+)
+```
+
+If you prefer to start and stop the server with Jest instead of relying on the automatic
+management by the library, use Jest's `beforeAll()` and `afterAll()`, the you don't need
+to wrap your test functions with `wg.test()`:
+
+```typescript
+beforeAll(async () => {
+  await wg.setupUp()
```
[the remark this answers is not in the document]
fiam: Fixed

docs-website/src/pages/docs/guides/testing.md
fiam: I've rephrased it from a different angle, encouraging users to create as little instances as possible. Let me know what you think.

docs-website/src/pages/docs/guides/testing.md
```diff
@@ -0,0 +1,164 @@
...
+```
+
+Finally, define your tests as a Jest test function but wrapped with `wg.test()`. This will take care
+of automatically setting up and tearing down the server.
+
+```typescript
+test(
+  'country by code',
+  // Wrap your function with wg.test()
+  wg.test(async () => {
```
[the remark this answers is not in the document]
fiam: I agree, having a single way to do something is better than having two, it makes everything clearer.

Give me another day to think to try to get the automatic API right. The biggest problem right now is the ergonomics with testing frameworks that require test functions to accept an argument (like ava). If I can't find a suitable solution, 

If we can't find a way to use the automatic API with concurrent tests, we'll go with the "manual" one (using beforeAll/afterAll or equivalent)

docs-website/src/pages/docs/guides/testing.md
```diff
@@ -0,0 +1,164 @@
+---
+title: Testing
+pageTitle: WunderGraph - Testing
+description: This guide shows how to test your WunderGraph applications
+---
+
+## Headless testing
```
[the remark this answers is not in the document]
StarpTech: I've never heard the term headless in testing expect if it comes to frontend testing e.g with playwright. I'd go with integration which is a term.

docs-website/src/pages/docs/guides/testing.md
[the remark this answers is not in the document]
fiam: Removed automatic API

docs-website/src/pages/docs/guides/testing.md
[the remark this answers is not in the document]
fiam: Changed to `integration`

Pagebakers (approved):
awesome

jensneuse (changes requested):

docs-website/src/pages/docs/guides/testing.md
```diff
@@ -0,0 +1,160 @@
...
+// Imports from Ava
+import { test } from 'ava'
+```
+
+We recommended creating as little `Server` instances as possible and sharing them between
+multiple tests. These minimizes the number of times the server starts and stops, making your tests
+faster.
+
+```typescript
+const wg = new Server({ createClient })
```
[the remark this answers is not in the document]
fiam: I'll rename it

docs-website/src/pages/docs/guides/testing.md
```diff
@@ -0,0 +1,160 @@
...
+// Imports from Jest
+import { describe, expect, test } from '@jest/globals'
+```
+
+We recommended creating as little `Server` instances as possible and sharing them between
+multiple tests. These minimizes the number of times the server starts and stops, making your tests
+faster.
+
+```typescript
+const wg = new Server({ createClient })
```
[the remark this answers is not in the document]
fiam: This is a bit tricker because there's no fixed location for `client.ts`, we're relying on the user to provide the correct path in the import. Do you think we should generate a `tests.ts` file with a template that imports `Client`? Seems like a lot of work for a minor gain to me, but we can do it that way.

docs-website/src/pages/docs/guides/testing.md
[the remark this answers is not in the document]
fiam: Renamed

docs-website/src/pages/docs/guides/testing.md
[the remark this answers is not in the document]
fiam: I've prototyped this with an additional `templates.typescript.testing` template and I'm linking it a lot. Not just for avoiding passing the `createClient`, but also as plumbing for additional testing boilerplate (e.g. mocks). I'll clean it up and push.

examples/simple/package.json
```diff
@@ -5,16 +5,21 @@
...
     "generate": "wunderctl generate up --debug",
-    "check": "tsc --noEmit"
+    "check": "tsc --noEmit",
+    "test": "jest"
   },
   "author": "",
   "license": "ISC",
   "dependencies": {
     "@wundergraph/sdk": "^0.124.0",
+    "@types/node-fetch": "^2.6.2",
```
StarpTech, line 15: Should be in devDeps

StarpTech (changes requested):

examples/simple/package.json
fiam: Good catch, fixed

packages/sdk/src/testing/index.ts
```diff
@@ -0,0 +1,215 @@
...
+				const resp = await this.options.fetch(health, { signal: controller.signal });
+				clearTimeout(id);
+				if (resp.status == 200) {
+					const data = (await resp.json()) as ServerHealth;
+					if (data.nodeStatus === readyStatus && data.serverStatus === readyStatus) {
+						break;
+					}
+				}
+			} catch (e: any) {
+				if (maxWaitMs > 0 && new Date().getTime() - started > maxWaitMs) {
```
StarpTech, line 135: Please log the original error

packages/sdk/src/testing/index.ts
```diff
@@ -0,0 +1,215 @@
...
+		}
+		const health = this.url('/health');
+		const started = new Date().getTime();
+		const maxWaitMs = this.options.startupTimeoutSeconds * 1000;
+		while (true) {
+			try {
+				const controller = new AbortController();
+				const id = setTimeout(() => controller.abort(), 500);
+				const resp = await this.options.fetch(health, { signal: controller.signal });
+				clearTimeout(id);
```
StarpTech, line 127: You should clean the timeout also in catch case.

StarpTech (changes requested):

packages/sdk/src/testing/index.ts
fiam: We need to add more context, otherwise it's just going to say `fetch() failed` or something like that which might look like something else failed, right?

packages/sdk/src/testing/index.ts
fiam: Fixed, PTAL

packages/sdk/src/testing/index.ts
StarpTech: This is the `start` method. I think you mixed it up.

fiam: No, I knew which method you were referring to. This loop is waiting for the server to come up, checking the health endpoint. The only errors we might get here are either fetch failing or a problem decoding/checking the JSON returned from the health endpoint. I think those need to be augmented with some context, otherwise the user will have no clue were they came from and might even think the problem is in their tests/code.

packages/sdk/src/testing/index.ts
```diff
@@ -0,0 +1,216 @@
...
+				cmd.push('--cli-log-level', 'warning');
+			}
+			subprocess = wunderctlSubprocess({ cmd, env });
+			subprocess?.stdout?.pipe(process.stdout);
+			subprocess?.stderr?.pipe(process.stderr);
+		}
+		const health = this.url('/health');
+		const started = new Date().getTime();
+		const maxWaitMs = this.options.startupTimeoutSeconds * 1000;
+		while (true) {
```
StarpTech, line 122: I'd use this lib to cover retry and max retries https://github.com/normartin/ts-retry-promise

StarpTech (changes requested):

StarpTech (commented):
LGTM

jensneuse (approved):

packages/sdk/src/codegen/templates/typescript/testing.template.ts
```diff
@@ -0,0 +1,17 @@
+//language=handlebars
+export const handlebarTemplate = `
+import { ServerOptions, WunderGraphTestServer as WunderGraphTestServerInternal } from '@wundergraph/sdk/testing';
```
StarpTech, line 3: Code that is not tightly coupled to the type generation should always live in a codebase for maintainability reasons. It's great to have one single import, but the downsides are not worth it. There are two ways.

1. You put this code in the SDK and export only the class
2. You put everything in the SDK

StarpTech (changes requested):

StarpTech (approved):
LGTM

PR #349, opened 2022-11-18
feat(create-wundergraph-app): download latest tag instead of main

fiam:
Downloading from the main branch prevents us from making breaking
changes, since the examples would stop working until the next SDK with
the breaking changes is released.

When using the wundegraph repository as the example source, first
retrieve the tags and use the latest tag starting with @wundergraph/sdk

### DIFF
--- a/packages/create-wundergraph-app/src/helpers/getRepository.ts
+++ b/packages/create-wundergraph-app/src/helpers/getRepository.ts
@@ -42,7 +42,7 @@ const resolveRepository = async ({ exampleName, githubLink }: { exampleName?: st
 			return {
 				repoOwnerName: 'wundergraph',
 				repoName: 'wundergraph',
-				ref: 'main',
+				ref: ref,
 				filePath: `examples/${selectedExampleName}`,
 			};
 		} else if (githubLink) {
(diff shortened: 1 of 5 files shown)


packages/create-wundergraph-app/src/helpers/getRepository.ts
```diff
@@ -22,16 +30,19 @@ const resolveRepository = async ({ exampleName, githubLink }: { exampleName?: st
...
-			const selectedExampleName = await checkIfValidExample(exampleName);
+			if (!ref) {
+				ref = await resolveLatestWundergraphRef();
+			}
+			const selectedExampleName = await checkIfValidExample(exampleName, ref);
 			return {
 				repoOwnerName: 'wundergraph',
 				repoName: 'wundergraph',
-				branch: 'main',
+				ref: 'main',
```
JivusAyrus, line 45: Shouldn't this be ' ref: ref || "main" '

JivusAyrus (changes requested):

Pagebakers (approved):
Nice one!

packages/create-wundergraph-app/src/helpers/getRepository.ts
fiam: You're right, good catch. I'll update it.

fiam: Fixed, PTAL

JivusAyrus (approved):
LGTM

PR #335, opened 2022-11-11
feat: remove app/main from URL structure

fiam:
Fixes ENG-701, ENG-703, ENG-704, ENG-705, ENG-706, ENG-730

- feat: remove /<app-name>/main prefix from router and pb definition
- chore(docs): update docs to reflect changes to WunderGraph protocol
- chore(docs): update docs to reflect changes to WunderGraph protocol
- fix: update code generation (#333)
- feat: update URLs in testapps for new style
- feat: remove /<app-name>/main prefix from router and pb definition
- fix: update code generation (#333)
- feat: drop Client applicationPath and associated properties
- fix: port apollo-federation example to new URL/API
- fix: update /auth0-oidc-authentication to new URL/API
- fix: update caching example to new URL/API
- fix: update cross-api-joins to new URL/API
- fix: update faunadb-nextjs to new URL/API
- fix: update fragments example to new URL/API
- fix: update golang-client example to new URL/API
- fix: update graphql-sse-subscriptions example to new URL/API
- fix: update graphql-subscriptions-hooks example to new URL/API
- fix: update graphql-ws-subscriptions example to new URL/API
- fix: update graphql-yoga-sse-subscriptions example to new URL/API
- fix: update hooks example to new URL/API
- fix: update inject-bearer example to new URL/API
- fix: update keycloak-oidc-authentication example to new URL/API
- fix: update migrate-from-apollo example to new URL/API
- fix: update nextjs example to new URL/API
- fix: update nextjs-postgres-prisma to new URL/API
- fix: update nextjs-swr example to new URL/API
- fix: update rbac example to new URL/API
- fix: upadte  schema-extension example to use new URL/API
- fix: update simple example to new URL/API
- fix: update webhooks example to new URL/API
- fix: update URLs in README files for new style
- fix: update publish-install-api to new URL/API
- fix: update graphql-hasura-subscriptions to new URL/API
- fix: permissions in graphql-apollo-subscriptions example
- fix: update graphql-apollo-subscriptions example to new URL/API
- fix: typo in graphql-apollo-subscriptions README
- fix: update postgres example to new URL/API
- fix: update nextjs-postgres-user-filters example to new URL/API
- docs: fix old-style URLs in a couple of pages
- feat: add redirect for old-style URLs

PR #353, opened 2022-11-20
fix: remove deprecated prefix from cookie path

alasdairtran:
PR #335 breaks our auth flow with Auth0 - cookies are missing after logging in.

After digging around a bit, I think you guys simply forgot to remove the old prefix from the cookie path.


jensneuse (approved):

jensneuse:
You're right! Thank you!

jensneuse:
Merged with #354

PR #365, opened 2022-11-23
fix: make prisma engine exit cleanly

fiam:
Fixes ENG-511
Fixes ENG-740
Fixes ENG-929

### DIFF
--- a/pkg/datasources/database/engine.go
+++ b/pkg/datasources/database/engine.go
@@ -14,6 +14,7 @@ import (
 	"path"
 	"strconv"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/go-cmd/cmd"
@@ -63,6 +64,7 @@ type Engine struct {
 	queryEnginePath         string
 	introspectionEnginePath string
 	url                     string
+	cmd                     *exec.Cmd
 	cancel                  func()
 	client                  *http.Client
 	log                     *zap.Logger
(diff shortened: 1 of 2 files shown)


pkg/datasources/database/engine.go
```diff
@@ -255,6 +260,10 @@ func (e *Engine) StartQueryEngine(schema string) error {
 
 func (e *Engine) ensurePrisma() error {
 
+	fmt.Println("ensurePrisma()", time.Now())
```
jensneuse, line 263: I'd rename this to "installing prisma dependencies"

jensneuse (changes requested):

pkg/datasources/database/engine.go
fiam: I don’t intend to merge this, at least for the time being. The messages are just for debugging where it gets stuck.

jensneuse (approved):
Amazing! Thank you.

fiam:
@jensneuse Wait a bit. I need to test on Windows.

fiam:
@jensneuse Changed `syscall.Kill()` to `os.Process.Kill()`, which provides a better implementation on Windows. Good to go! Thanks for the review 👍

pkg/datasources/database/engine.go
```diff
@@ -299,6 +302,23 @@ func (e *Engine) StopQueryEngine() {
 		return
 	}
 	e.cancel()
+	exitCh := make(chan error, 1)
```
StarpTech, line 304: Why buffered?

StarpTech (changes requested):

pkg/datasources/database/engine.go
fiam: The idea was to prevent the goroutine waiting for the exit to block, but it should be fine with an unbuffered channel because the select in the parent goroutine will be already waiting by that point.

StarpTech: If it is not needed we should remove it.

fiam: Changed 👍

PR #369, opened 2022-11-24
feat: add example for OpenAPI and mocking

fiam:

### DIFF
--- a/examples/openapi-mocking/.wundergraph/api.yaml
+++ b/examples/openapi-mocking/.wundergraph/api.yaml
@@ -12,6 +12,7 @@ paths:
   /notes/note/{noteID}:
     get:
       summary: Returns a note by ID
+      operationId: noteByID
       parameters:
         - name: noteID
           in: path
@@ -43,6 +44,7 @@ paths:
   /notes/new:
     post:
       summary: Creates a new note
+      operationId: newNote
       parameters:
         - name: text
           in: body
--- a/examples/openapi-mocking/.wundergraph/api.yaml
+++ b/examples/openapi-mocking/.wundergraph/api.yaml
@@ -1,0 +1,63 @@
+openapi: 3.0.0
+info:
+  title: Notes API
+  description: An API which supports storing and retrieving notes
+  version: '0.1'
+
+servers:
+  - url: http://api.example.com/v1
+    description: Example Server
+
+paths:
+  /notes/note/{noteID}:
+    get:
+      summary: Returns a note by ID
+      parameters:
+        - name: noteID
+          in: path
+          required: true
+          description: Note unique identifier
+          schema:
+            type: integer
+            format: int64
+            minimum: 1
+      responses:
+        '200':
+          description: A note object
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  id:
+                    type: integer
+                    format: int64
+                  text:
+                    type: string
+        '400':
+          description: The specified note ID is invalid (not a number).
+        '404':
+          description: A note with the specified ID was not found.
+        default:
+          description: Unexpected error
+  /notes/new:
+    post:
+      summary: Creates a new note
+      parameters:
+        - name: text
+          in: body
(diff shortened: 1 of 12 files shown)


StarpTech:
Please add the example to the index on the readme.

fiam:
> Please add the example to the index on the readme.

I had totally missed that list. Adding it, thanks!

examples/openapi-mocking/.wundergraph/api.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+servers:
+  - url: http://api.example.com/v1
+    description: Example Server
+
+paths:
+  /notes/note/{noteID}:
+    get:
+      summary: Returns a note by ID
+      parameters:
```
jensneuse, line 15: add the property "operationId" (type string) to give this field a better GraphQL-like name, e.g. "noteByID"

Also add a comment to the README and docs website about this.
It'll help users get a better generated GraphQL Schema.

examples/openapi-mocking/.wundergraph/api.yaml
```diff
@@ -0,0 +1,63 @@
...
+        '400':
+          description: The specified note ID is invalid (not a number).
+        '404':
+          description: A note with the specified ID was not found.
+        default:
+          description: Unexpected error
+  /notes/new:
+    post:
+      summary: Creates a new note
+      parameters:
```
jensneuse, line 46: Add "operationId" here as well.

examples/openapi-mocking/.wundergraph/api.yaml
jensneuse: Example: https://github.com/wundergraph/wundergraph/blob/main/testapps/default/json_placeholder.json#L18

jensneuse: Don't forget to re-generate the code and update the Operations after this change.

examples/openapi-mocking/.wundergraph/wundergraph.server.ts
```diff
@@ -0,0 +1,71 @@
...
+	hooks: {
+		queries: {
+			NoteByID: {
+				mockResolve: async (hook) => {
+					const noteID = hook.input.id;
+					const text = notes[noteID];
+					if (text != null) {
+						return {
+							data: {
+								notes_getNotesNoteByNoteID: {
```
jensneuse, line 18: needs update as well after changing operationId

examples/openapi-mocking/.wundergraph/wundergraph.server.ts
```diff
@@ -0,0 +1,71 @@
...
+							notes_postNotesNew: {
+								id: nextID,
+							},
+						},
+					};
+				},
+			},
+		},
+	},
+	graphqlServers: [
```
jensneuse, line 52: can be removed.

examples/openapi-mocking/README.md
```diff
@@ -0,0 +1,39 @@
...
+#### Create a new note
+
+```shell
+curl -d text=Hello http://localhost:9991/operations/NewNote
+```
+
+Notice the returned JSON with the newly created ID:
+
+```json
+{ "data": { "notes_postNotesNew": { "id": 1 } } }
```
jensneuse, line 24: update and add info how to set operationId

jensneuse (changes requested):

examples/openapi-mocking/.wundergraph/wundergraph.server.ts
fiam: Good catch, thanks!

examples/openapi-mocking/.wundergraph/api.yaml
fiam: I've added it in the 2 places you suggested and updated the code accordingly. PTAL

StarpTech (approved):
LGTM

PR #371, opened 2022-11-24
fix: remove colors from configurePublishWunderGraphAPI()

fiam:
Interferes with logging, printing color sequences to JSON logs. Also,
we now have colors via --pretty-logging


StarpTech (approved):
LGTM

spetrunin (approved):

PR #372, opened 2022-11-24
fix(ci): deprecation warnings

fiam:
upgrade pnpm/action-setup to v2.2.4
change ::set-output to new $GITHUB_OUTPUT API


StarpTech (approved):
LGTM

PR #373, opened 2022-11-25
fix: validate OpenID discovered configuration, warn when appropriate

fiam:
end_session_endpoint is not a mandatory field, so it should print a
warning instead of an error. Also, add validation for the required
fields according to the spec.

### DIFF
--- a/pkg/apihandler/apihandler.go
+++ b/pkg/apihandler/apihandler.go
@@ -1985,7 +1985,7 @@ func (r *Builder) configureOpenIDConnectIssuerLogoutURLs() map[string]string {
 			continue
 		}
 		if resp.StatusCode != http.StatusOK {
-			r.log.Error("failed to get openid-configuration",
+			r.log.Error("failed to get openid-configuration, provider returned an HTTP error",
 				zap.Int("status", resp.StatusCode),
 			)
 			continue
@@ -2006,11 +2006,12 @@ func (r *Builder) configureOpenIDConnectIssuerLogoutURLs() map[string]string {
 			)
 			continue
 		}
-		if config.EndSessionEndpoint == "" {
-			r.log.Error("failed to get openid-configuration",
-				zap.String("end_session_endpoint", config.EndSessionEndpoint),
-			)
+		if err := config.Validate(); err != nil {
+			r.log.Warn("failed to get openid-configuration", zap.Error(err))
 			continue
+		}
+		if config.EndSessionEndpoint == "" {
+			r.log.Warn("issuer doesn't support end_session_endpoint", zap.String("issuer", issuer))
 		}
 		issuerLogoutURLs[issuer] = config.EndSessionEndpoint
 	}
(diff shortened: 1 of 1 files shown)


pkg/apihandler/apihandler.go
```diff
@@ -2003,12 +2003,13 @@ func (r *Builder) configureOpenIDConnectIssuerLogoutURLs() map[string]string {
 			)
 			continue
 		}
-		if config.EndSessionEndpoint == "" {
-			r.log.Error("failed to get openid-configuration",
-				zap.String("end_session_endpoint", config.EndSessionEndpoint),
-			)
+		if err := config.Validate(); err != nil {
+			r.log.Warn("failed to get openid-configuration", zap.Error(err))
```
StarpTech, line 2007: No warn, debug. It is optional.

StarpTech (changes requested):

pkg/apihandler/apihandler.go
StarpTech: Sry, I mean for the EndSessionEndpoint endpoint

pkg/apihandler/apihandler.go
```diff
@@ -2003,12 +2003,13 @@ func (r *Builder) configureOpenIDConnectIssuerLogoutURLs() map[string]string {
...
 			continue
 		}
+		if config.EndSessionEndpoint == "" {
+			r.log.Warn("issuer doesn't support end_session_endpoint", zap.String("issuer", issuer))
```
jensneuse, line 2011: Could be info or debug I think.

pkg/apihandler/apihandler.go
```diff
@@ -2024,6 +2025,30 @@ type OpenIDConnectConfiguration struct {
...
+	}
+	if c.AuthorizationEndpoint == "" {
+		return openIDConnectConfigurationMissingFieldError("authorization_endpoint")
+	}
+	if c.TokenEndpoint == "" {
+		// TODO: This might be optional with Implicit Flow?
+		return openIDConnectConfigurationMissingFieldError("token_endpoint")
+	}
+	if c.JwksUri == "" {
+		return openIDConnectConfigurationMissingFieldError("jwks_uri")
```
jensneuse, line 2047: this is optional I think

pkg/apihandler/apihandler.go
```diff
@@ -2024,6 +2025,30 @@ type OpenIDConnectConfiguration struct {
...
+	// See https://openid.net/specs/openid-connect-discovery-1_0.html
+	if c.Issuer == "" {
+		return openIDConnectConfigurationMissingFieldError("issuer")
+	}
+	if c.AuthorizationEndpoint == "" {
+		return openIDConnectConfigurationMissingFieldError("authorization_endpoint")
+	}
+	if c.TokenEndpoint == "" {
+		// TODO: This might be optional with Implicit Flow?
+		return openIDConnectConfigurationMissingFieldError("token_endpoint")
```
jensneuse, line 2044: we don't want to support implicit flow, only PKCE

jensneuse (changes requested):

pkg/apihandler/apihandler.go
fiam: 👍

fiam: Changed to debug

pkg/apihandler/apihandler.go
fiam: Changed to debug

pkg/apihandler/apihandler.go
fiam: Changed to debug

pkg/apihandler/apihandler.go
fiam: Maybe I’m looking at the wrong spec, but it says it’s required here https://openid.net/specs/openid-connect-discovery-1_0.html

jensneuse: I'm just not sure if it will always be present...

fiam: Made it optional

jensneuse (approved):

StarpTech (approved):
LGTM

PR #374, opened 2022-11-25
feat: make detecting invalid operations easier

fiam:
- When an operation is not valid, either by being empty or by producing
an error during parsing, generate a handler that just returns an error
for it.


pkg/loadoperations/loader.go
```diff
@@ -159,6 +142,43 @@ func (l *Loader) Load(operationsRootPath, fragmentsRootPath, schemaFilePath stri
 	return string(encodedOutput)
 }
 
+type infoError string
```
jensneuse, line 145: What's the purpose of this type alias?

pkg/loadoperations/loader.go
```diff
@@ -159,6 +142,43 @@ func (l *Loader) Load(operationsRootPath, fragmentsRootPath, schemaFilePath stri
...
+func (l *Loader) loadOperation(file GqlFile, normalizer *astnormalization.OperationNormalizer, fragments string, schemaDocument *ast.Document) (string, error) {
+	content, err := ioutil.ReadFile(file.FilePath)
+	if err != nil {
+		return "", fmt.Errorf("error reading file: %w", err)
+	}
+	doc, report := astparser.ParseGraphqlDocumentString(string(content) + fragments)
+	if report.HasErrors() {
+		return "", fmt.Errorf("error parsing operation: %s", report.Error())
+	}
+	if len(doc.OperationDefinitions) > 1 {
```
jensneuse, line 159: This is actually wrong, although it looks correct.

You need to iterate through the doc.RootNodes and count the number of root nodes that are of kind OperationDefinition.

Counting the number of OperationDefinitions is correct most of the time, but not always.
E.g. we "delete" an OperationDefinition from an AST by deleting the reference in the RootNodes.
By deleting it, the Walker will no longer walk into the OperationDefinition, hence it "doesn't exist" anymore.

By directly using doc.OperationDefinitions, all you do is access pre-allocated space, that might or might not be correct.

jensneuse (changes requested):

pkg/loadoperations/loader.go
fiam: It's used to report errors that should be passed to the TS side as `info` instead of `error`, see the type assertion on line 124 after calling `l.loadOperation()`

pkg/apihandler/apihandler.go
```diff
@@ -390,29 +394,38 @@ func (r *Builder) registerWebhook(config *wgpb.WebhookConfiguration) error {
...
 
+func (r *Builder) operationApiPath(name string) string {
+	return fmt.Sprintf("/operations/%s", name)
+}
+
+func (r *Builder) registerInvalidOperation(name string) {
+	apiPath := r.operationApiPath((name))
+	route := r.router.Methods(http.MethodGet, http.MethodPost, http.MethodOptions).Path(apiPath)
+	route.Handler(&EndpointUnavailableHandler{
+		OperationName: name,
```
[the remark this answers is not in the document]
fiam: We can definitely log an error there. I'll add it.

pkg/loadoperations/loader.go
fiam: I tested it locally with a few files and it seemed to work. I'll look into it and count the nodes as you suggest.

pkg/loadoperations/loader.go
```diff
@@ -116,37 +117,19 @@ func (l *Loader) Load(operationsRootPath, fragmentsRootPath, schemaFilePath stri
...
-			out.Errors = append(out.Errors, fmt.Sprintf("graphql document must contain exactly one operation: %s\n", file.FilePath))
-			continue
-		}
-
-		normalizer.NormalizeOperation(&doc, &schemaDocument, &report)
-		if report.HasErrors() {
-			out.Errors = append(out.Errors, fmt.Sprintf("error normalizing operation: %s, operationFilePath: %s", report.Error(), file.FilePath))
-			continue
+			out.Invalid = append(out.Invalid, file.OperationName)
+			if ierr, ok := err.(infoError); ok && ierr.IsInfo() {
```
jensneuse, line 124: it would be a bit more idiomatic to just use `errors.Is(infoError,infoErrorType)`

jensneuse (changes requested):

pkg/loadoperations/loader.go
jensneuse: It will always work unless someone manipulates the AST.
I only wanted to make you aware of the behaviour of the doc object as it might bite you when working with graphql-go-tools.
Right after parsing, the AST should be correct, no deletions happened.
So you can leave the code as is, but just keep it in mind.

pkg/apihandler/apihandler.go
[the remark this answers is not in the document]
fiam: Added

pkg/loadoperations/loader.go
fiam: `errors.Is` test against a given error value, not its type. I've changed the check to use `errors.As` which can be used to test its type

pkg/loadoperations/loader.go
fiam: I've changed it to walk the AST and count OperationDefinitions nodes

jensneuse (approved):

StarpTech:
> Allow .graphql files with zero operations, mark them as invalid. This
allows commenting them out

Honestly, I don't like that approach. You can easily move the file to a different location. Next.js also didn't try to solve this "issue".

fiam:
> > Allow .graphql files with zero operations, mark them as invalid. This
> > allows commenting them out
> 
> Honestly, I don't like that approach. You can easily move the file to a different location. Next.js also didn't try to solve this "issue".

It feels a bit weird to me that an entirely commented file (or an empty one) is considered an error, but maybe that's more frequent in frameworks I don't have experience with. I've changed it so empty files are treated as files with errors. PTAL.

StarpTech:
> It feels a bit weird to me that an entirely commented file (or an empty one) is considered an error, but maybe that's more frequent in frameworks I don't have experience with. I've changed it so empty files are treated as files with errors. PTAL.

The more I think about it, the less I'm opinionated. I see it from this perspective: An empty string is not valid graphql and an operation contains just graphql. If we throw an error, we can ensure that everything that can be picked as an operation is a valid operation. The user is free to design their operations in a different directory if they can't provide the right query yet.

fiam:
> > It feels a bit weird to me that an entirely commented file (or an empty one) is considered an error, but maybe that's more frequent in frameworks I don't have experience with. I've changed it so empty files are treated as files with errors. PTAL.
> 
> The more I think about it, the less I'm opinionated. I see it from this perspective: An empty string is not valid graphql and an operation contains just graphql. If we throw an error, we can ensure that everything that can be picked as an operation is a valid operation. The user can design their operations in a different directory if they can't provide the right query yet.

Honestly, the more I think about it the better it seems to report them as errors.

I don’t think we should discuss wether an empty string is or is not valid GraphQL, that’s going down the realm of semantics. Instead, think of the practical implications.

The reason I wanted to make empty files acceptable is that commenting out a file is way faster than moving it (and simpler, since you don’t need to think where to move it). However, there’s a drawback: if you accidentally leave an empty file because you forgot to save its contents wunderctl generate will report a message but not stop with an error. For this reason, I think is better if we err on the side of correctness even if it comes with a cost in terms of convenience.

pkg/apihandler/apihandler.go
```diff
@@ -393,29 +397,39 @@ func (r *Builder) registerWebhook(config *wgpb.WebhookConfiguration) error {
...
+		Logger:        r.log,
+	})
+	r.log.Error("EndpointUnavailableHandler",
+		zap.String("Operation", name),
+		zap.String("Endpoint", apiPath),
+		zap.String("Help", "The Operation is not properly configured. This usually happens when there is a mismatch between GraphQL Schema and Operation. Please make sure, the Operation is valid. This can be supported best by enabling intellisense for GraphQL within your IDE."),
```
StarpTech, line 414: The last phrase is not helpful. Can we point to some docs?

StarpTech (changes requested):

packages/sdk/src/graphql/operations.ts
```diff
@@ -1276,16 +1283,22 @@ export const loadOperations = (schemaFileName: string): string => {
...
 	const output = result?.stdout;
+	let content = '';
+	let invalidOperationNames: string[] = [];
 	if (output) {
 		const out = JSON.parse(output) as LoadOperationsOutput;
 		out.info?.forEach((msg) => Logger.info(msg));
 		out.errors?.forEach((msg) => Logger.error(msg));
-		if ((out.errors?.length ?? 0) > 0 && out?.errors?.[0]) {
+		if (WG_THROW_ON_OPERATION_LOADING_ERROR && (out.errors?.length ?? 0) > 0 && out?.errors?.[0]) {
 			throw new Error(out.errors[0]);
```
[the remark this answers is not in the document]
fiam: I did, it exists with 1 because of the thrown exception.

pkg/apihandler/apihandler.go
fiam: I agree with its lack of usefulness, but I also think in most cases it would be impossible to find the relevant doc. What if we direct them to check the logs, which will have the relevant error?

fiam: I've updated the message, PTAL

PR #380, opened 2022-11-29
fix: fauna-db example

kaleem68:
#### Checklist

- [ ] run `make test`

### DIFF
--- a/examples/faunadb-nextjs/pages/index.tsx
+++ b/examples/faunadb-nextjs/pages/index.tsx
@@ -1,5 +1,5 @@
 import { NextPage } from 'next';
-import { useQuery, withWunderGraph } from '../components/generated/nextjs';
+import { useQuery } from '../lib/wundergraph';
 
 const Home: NextPage = () => {
 	const stores = useQuery({
@@ -98,4 +98,4 @@ const Home: NextPage = () => {
 	);
 };
 
-export default withWunderGraph(Home);
+export default Home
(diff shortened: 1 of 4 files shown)


StarpTech (approved):
LGTM

examples/faunadb-nextjs/pages/index.tsx
```diff
@@ -98,4 +98,4 @@ const Home: NextPage = () => {
 	);
 };
 
-export default withWunderGraph(Home);
+export default Home
```
jensneuse, line 101: why remove withWunderGraph? no SSR?

jensneuse (changes requested):

examples/faunadb-nextjs/pages/index.tsx
kaleem68: TL;DR: `withWunderGraph` is not exported from `lib/wundergraph.ts`.

If we need `withWunderGraph` then it will be imported from a different file.

The changes will be as following:

```diff
+ import {withWunderGraph} from "../components/generated/nextjs";
  import { useQuery } from '../lib/wundergraph';
```
```diff
- export default Home
+ export default withWunderGraph(Home);
```

Do we want to do the above? 

I copied `lib/wundergraph.ts` from `examples/nextjs-swr` to make this example work and in the `examples/nextjs-swr` `withWunderGraph` is not exported.

StarpTech: I'd favor doing not more than showing that WG works with faunadb. For advanced next.js stuff, we have a dedicated example.

jensneuse (approved):

jensneuse:
I've just changed to "fix: ..." because "chore: fix..." feels wrong.

examples/faunadb-nextjs/pages/index.tsx
kaleem68: @StarpTech Are you saying we are good with existing changes?

PR #396, opened 2022-12-05
feat: upgrade to Node.js LTS (18)

StarpTech:
#### Checklist

- [ ] run `make test`


fiam (approved):
I would submit the documentation changes in a separate PR, but let’s get it done. LGTM

.github/actions/node/action.yaml
```diff
@@ -18,7 +18,7 @@ runs:
   steps:
     - uses: actions/setup-node@v3
       with:
-        node-version: '16'
+        node-version: '18'
```
thisisnithin, line 21: Can we do node-version: ${{ inputs.node-version }} instead? the default is already set to LTS.

thisisnithin (changes requested):

.github/actions/node/action.yaml
StarpTech: Good catch. Added.

Pagebakers (approved):
LGTM

thisisnithin (approved):
LGTM

PR #416, opened 2022-12-08
fix: return a 400 bad request when POST'ed JSON is invalid

fiam:
Response was failing to serialize because we were trying to include the
JSON back as a json.RawMessage, which failed to serialize. To avoid silently
ignoring another problem here in the future, make sure any errors writing
the error to the output are propagated to the caller.

### DIFF
--- a/pkg/inputvariables/validator.go
+++ b/pkg/inputvariables/validator.go
@@ -26,28 +26,53 @@ func NewValidator(schema string, disableVerboseErrors bool) (*Validator, error) 
 	return &validator, nil
 }
 
+type jsonError struct {
+	Message string `json:"message"`
+}
+
+func (e *jsonError) Error() string {
+	return e.Message
+}
+
 type ValidationError struct {
 	Message string
 	Input   json.RawMessage
-	Errors  []jsonschema.KeyError
+	Errors  []error
 }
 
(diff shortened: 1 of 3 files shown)


pkg/inputvariables/validator.go
```diff
@@ -26,43 +25,51 @@ func NewValidator(schema string, disableVerboseErrors bool) (*Validator, error)
...
 
+type jsonError struct {
+	Message string `json:"message"`
+}
+
+func (e *jsonError) Error() string {
+	return e.Message
+}
+
 type ValidationError struct {
```
jensneuse, line 36: As you're touching this already, can you make all fields pascal case, including the fields from jsonschema.KeyError, which would require to create an adapter type. This was already requested by @Pagebakers and would make the response shape in sync with all other responses.

jensneuse (changes requested):

pkg/inputvariables/validator.go
fiam: On it

fiam: Done

jensneuse (approved):
LGTM

Pagebakers (approved):
LGTM

PR #420, opened 2022-12-08
chore: remove version numbers from generated PB Go file

fiam:
Although version numbers could provide some value to detect version
incompatibility, we're only using PB to generate schema definitions
for JSON files, so this is a minor concern. Additionally, we do not
have version numbers in the TS side (because the generator doesn't
write them down). On the other hand, including the version number
causes significant friction for contributors using slightly different
PB versions while checking that the generated files are up to date)


types/generate.sh
```diff
@@ -1,9 +1,19 @@
...
   --go_opt=module=github.com/wundergraph/wundergraph \
   --go_opt=paths=import \
   --experimental_allow_proto3_optional \
   wundernode_config.proto
+
+outfile='../pkg/wgpb/wundernode_config.pb.go'
+
+# Remove version numbers. Generating the backup is required
+# for compatibility with both GNU and BSD sed variants.
+sed -i.bak '2,4d' ${outfile}
```
StarpTech, line 18: This thread convinced me. https://github.com/golang/protobuf/issues/1185

Your regex is a bit dangerous. Could you use this one:

```sh
sed -i -e '1h;2,$H;$!d;g' -re 's|// versions.+// source:|// source:|' pkg/wgpb/*.pb.go
```

StarpTech (changes requested):

types/generate.sh
StarpTech: Whereby this is still valid https://github.com/golang/protobuf/issues/1185#issuecomment-1305074470

> Note that this does not apply semver; the promise is expressed in terms of a time window.

types/generate.sh
```diff
@@ -1,9 +1,18 @@
...
   --go_out=.. \
   --go_opt=module=github.com/wundergraph/wundergraph \
   --go_opt=paths=import \
   --experimental_allow_proto3_optional \
   wundernode_config.proto
+
+outfile="../pkg/wgpb/wundernode_config.pb.go"
+tmp="${outfile}.tmp"
+
+# Remove comments with version numbers
```
StarpTech, line 17: Could you add comment why we are doing this?

StarpTech (changes requested):

types/generate.sh
fiam: Changed to grep

types/generate.sh
fiam: I've added a comment explaining the rationale for dropping the version numbers as well as for choosing grep as the tool to do it. PTAL

StarpTech (approved):
LGTM

jensneuse (approved):

PR #421, opened 2022-12-08
chore: enforce formatting of prettier-formatted files in CI

fiam:
- Fix patterns in package.json, so prettier doesn't exit with an
error when some of the file patterns have no results
- Add generated PB files to .prettierignore
- Format files in CI before checking for dirty ones


.github/workflows/packages-ci.yaml
```diff
@@ -75,6 +75,9 @@ jobs:
       - name: Build all libs + apps
         run: pnpm build
 
+      - name: Format source code
+        run: pnpm format && pnpm format:examples
```
StarpTech, line 79: As discussed: Please only check for incorrectly formatted files and only files that were touched in the commit.

StarpTech (changes requested):

Pagebakers (approved):

StarpTech (approved):
LGTM

PR #423, opened 2022-12-09
chore: add nextauth example

Pagebakers:
#### Checklist

- [ ] run `make test`


pkg/authentication/authentication.go
```diff
@@ -687,13 +675,14 @@ type CookieUserHandler struct {
...
-func (u *CookieUserHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+func (u *UserHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	user := UserFromContext(r.Context())
 	if user == nil {
 		http.NotFound(w, r)
 		return
 	}
-	if u.HasRevalidateHook && r.URL.Query().Get("revalidate") == "true" {
+
+	if user.FromCookie && u.HasRevalidateHook && r.URL.Query().Get("revalidate") == "true" {
```
jensneuse, line 685: We could also revalidate with a token I think. This would allow us to call the userInfo endpoint again.

jensneuse (changes requested):

docs-website/src/pages/docs/guides/token-based-authentication-with-next-auth.md
```diff
@@ -0,0 +1,301 @@
...
+export function middleware(request: NextRequest) {
+  // retrieve the session token from the cookie
+  const token = request.cookies.get('next-auth.session-token')?.value
+
+  let pathname = request.nextUrl.pathname.replace('/api/wg', '')
+
+  // rewrite the api url to the WunderGraph API
+  const url = new URL(
+    pathname + request.nextUrl.search,
+    'http://127.0.0.1:9991'
```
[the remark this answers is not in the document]
Pagebakers: Yes, sure, but for the sake of the example I didn't add any env variables. I'll add an extra not to clarify this.

pkg/authentication/authentication.go
Pagebakers: Yeah but response caching is never enabled for token, since it doesn't have an ETag

Pagebakers: So revalidate with token would then reset the internal userInfo cache?

jensneuse: We cache the response of the token internally. So if the token doesn't change and we have a cache entry, we're not calling the userInfo endpoint again. In case of changing some user details, it would be nice if we could automatically revalidate the cache entry. So, by setting revalidate to true, we could skip the cache and refresh it.

docs-website/src/pages/docs/guides/token-based-authentication-with-next-auth.md
```diff
@@ -0,0 +1,301 @@
...
+npm install @wundergraph/sdk @wundergraph/nextjs swr
+```
+
+Add the following WunderGraph commands to your package.json
+
+```json
+{
+  "scripts": {
+    "wundergraph": "wunderctl up",
+    "generate": "wunderctl generate"
```
StarpTech, line 32: Let's use `build` as the script name. We use it now as default for all examples.

packages/sdk/src/client/client.ts
```diff
@@ -26,11 +26,14 @@ export class Client {
...
 
 		this.extraHeaders = { ...options.extraHeaders };
+
+		this.csrfEnabled = options.csrfEnabled ?? true;
 	}
 
 	private readonly baseHeaders: Headers = {};
 	private extraHeaders: Headers = {};
 	private csrfToken: string | undefined;
+	private csrfEnabled: boolean = true;
```
StarpTech, line 36: Why should it be possible to disable it?

testapps/nextauth/pages/api/auth/userInfo.ts
```diff
@@ -0,0 +1,14 @@
+import { getToken } from 'next-auth/jwt';
+
+export default async (req, res) => {
+	// If you don't have NEXTAUTH_SECRET set, you will have to pass your secret as `secret` to `getToken`
+	const token = await getToken({ req });
```
StarpTech, line 5: According to the code, this is the only place where we verify the JWT. From a security perspective, it would make sense to verify the token when it arrives at our node to not make unauthenticated requests. Currently, we reuse the userInfoEndpoint to do both but we aren't in control of it from wundergraph perspective.

StarpTech (changes requested):

packages/sdk/src/client/client.ts
Pagebakers: Because its now always enabled, even when no auth is enabled, or for server side requests

testapps/nextauth/pages/api/auth/userInfo.ts
StarpTech: We can go this way, but we should ensure that we don't make decisions at the node based on the authenticity of the token.

packages/sdk/src/client/client.ts
StarpTech: ok

testapps/nextauth/pages/api/auth/userInfo.ts
Pagebakers: As discussed in the standup. It's OK to leave the validation to the userInfo endpoint according to OIDC spec.

pkg/authentication/authentication.go
Pagebakers: Add revalidate support for the userInfo cache

StarpTech (approved):
LGTM

fiam (approved):

PR #424, opened 2022-12-09
feat: implement telemetry

StarpTech:
This PR implements a basic telemetry client that tracks the usage and duration of marked commands.

During the implementation, I also removed the hub code that is no longer used because the way how we want to enable users to share and consume API as packages have significantly changed. That allowed me to clean up several places.

Telemetry is enabled by default. The user can disable it with `wunderctl generate --telemetry=false`. To get an overview of what data is collected on each invocation, the user can debug it with `wunderctl generate --telemetry-debug `.

To distinguish between different clients we store a unique ID to `~/wundergraph/config.json` that is sent with each metric. That ID is persistent across runs.

Following commands are tracked:

- wunderctl start
- wunderctl generate
- wunderctl up
- wunderctl node start
- wunderctl server start

#### Checklist

- [ ] run `make test`


cli/commands/root.go
```diff
@@ -119,55 +109,87 @@ You can opt out of this by setting the following environment variable: WUNDERGRA
...
+				metricName := cmd.Name()
+				if cmd.HasParent() {
+					metricName = telemetry.CmdMetricNameWithParent(cmd.Parent().Name(), cmd.Name())
+				}
+
+				metricUsageName := telemetry.CmdUsageMetricName(metricName)
+				TelemetryClient.Counter(metricUsageName)
+
+				// Send telemetry data
+				go func() {
```
jensneuse, line 140: Is it intentional to do this async? This might have the effect that it doesn't complete if the actual action is faster.
Should you not make sure that the process doesn't exit before sending the telemetry data is complete?

cli/commands/root.go
```diff
@@ -119,55 +109,90 @@ You can opt out of this by setting the following environment variable: WUNDERGRA
...
+						} else {
+							log.Info("Telemetry data sent")
+						}
+					}
+
+					// Track command duration.
+					// We have to do it after the flush to not send the metric too early
+					// The data is sent on the next flush
+					metricDurationName := telemetry.CmdDurationMetricName(metricName)
+					TelemetryClient.Duration(metricDurationName)
```
jensneuse, line 154: Can you rename the functions to give them more meaning?
From looking at the code, I don't understand it.
Also, `TelemetryClient.Duration(metricDurationName)` makes no sense to me.
Maybe it's the lack of verbs. I don't understand what happens here.

jensneuse: This is an example that is easier to understand for me:

https://github.com/open-telemetry/opentelemetry-go/blob/4e763472eee5ef03b97274694d2d1b07f0b3cd11/example/jaeger/main.go#L85

jensneuse (changes requested):

cli/commands/root.go
StarpTech: I don't want to block the actual usage. Wunderctl should work offline, and it could block for a noticeable time.

cli/commands/root.go
StarpTech: Please check

cli/commands/root.go
StarpTech: > Should you not make sure that the process doesn't exit before sending the telemetry data is complete?

I do https://github.com/wundergraph/wundergraph/pull/424/files#diff-6cef6e709aaccb651a756f23bbae52a22cbe8f82fbe90258831d7c189b42915aR172

StarpTech: If the process is exited before aster the data is sent to the client the data is also flushed on exit. See link.

StarpTech: But there is another flaw. I do count too late. The duration should start immediately not after the first metric was sent.

StarpTech: Please check

cli/commands/root.go
```diff
@@ -119,55 +111,97 @@ You can opt out of this by setting the following environment variable: WUNDERGRA
...
-		err = man.PersistChanges()
-		if err != nil {
-			return fmt.Errorf("unable to persist manifest changes")
-		}
-		err = man.WriteIntegrationsFile()
-		if err != nil {
-			return fmt.Errorf("unable to write integrations file")
+		// Check if telemetry is enabled
+		if rootFlags.Telemetry {
+			TelemetryClient = telemetry.NewClient(
```
jensneuse, line 116: Does it make sense to init the client when the telemetry annotation is not true?

cli/commands/root.go
```diff
@@ -119,55 +111,97 @@ You can opt out of this by setting the following environment variable: WUNDERGRA
...
+					if err != nil {
+						log.Error("Could not send telemetry data", zap.Error(err))
+					} else {
+						log.Info("Telemetry data sent")
+					}
+				}
+
+				// Track command duration.
+				// The data is sent on the next flush
+				// which is done after the command is finished
```
jensneuse, line 158: how? using `FlushTelemetry()` in the Execute func?

cli/commands/root.go
```diff
@@ -119,55 +111,97 @@ You can opt out of this by setting the following environment variable: WUNDERGRA
...
-	if err := rootCmd.Execute(); err != nil {
-		fmt.Println(err)
-		os.Exit(1)
+	err = rootCmd.Execute()
+}
+
+func FlushTelemetry() {
+	if TelemetryClient != nil && rootFlags.Telemetry {
+		// Create a batch of metrics which will be sent to the server
+		flusher := TelemetryClient.PrepareBatch()
```
jensneuse, line 195: Can you explain why we create a batch here and in line 144 as well?

jensneuse (changes requested):

cli/commands/root.go
StarpTech: yes

cli/commands/root.go
```diff
@@ -119,55 +111,85 @@ You can opt out of this by setting the following environment variable: WUNDERGRA
...
-		if err != nil {
-			return fmt.Errorf("unable to persist manifest changes")
-		}
-		err = man.WriteIntegrationsFile()
-		if err != nil {
-			return fmt.Errorf("unable to write integrations file")
+		// Check if telemetry is enabled
+		if rootFlags.Telemetry {
+			// Check if we want to track telemetry for this command
+			if cmd.Annotations["telemetry"] == "true" {
```
jensneuse, line 117: doesn't have to be 2 nested if, can be just one

pkg/config/config.go
```diff
@@ -61,7 +62,16 @@ func initViper() {
 		fmt.Println("Error loading config", err)
 	}
 
-	viper.SetEnvPrefix("WUNDERGRAPH")
+	// Init anonymous user id that is used to distinguish between different users
+	if !viper.InConfig("anonymousID") {
```
jensneuse, line 66: Should we disable this if "WG_TELEMETRY_DISABLED" env var is true?

cli/commands/root.go
```diff
@@ -190,22 +212,15 @@ func wunderctlBinaryPath() string {
...
 
-	viper.SetDefault("OAUTH_CLIENT_ID", "wundergraph-production-cli")
-	viper.SetDefault("OAUTH_BASE_URL", "https://accounts.wundergraph.com/auth/realms/master")
-	viper.SetDefault("API_URL", "https://api.wundergraph.com")
+	viper.SetDefault("API_URL", "https://gateway.wundergraph.com")
 
 	rootCmd.PersistentFlags().StringVarP(&rootFlags.CliLogLevel, "cli-log-level", "l", "info", "sets the CLI log level")
 	rootCmd.PersistentFlags().StringVarP(&DotEnvFile, "env", "e", ".env", "allows you to set environment variables from an env file")
 	rootCmd.PersistentFlags().BoolVar(&rootFlags.DebugMode, "debug", false, "enables the debug mode so that all requests and responses will be logged")
+	rootCmd.PersistentFlags().BoolVar(&rootFlags.Telemetry, "telemetry", os.Getenv("WG_TELEMETRY_DISABLED") == "", "enables telemetry. Telemetry allows us to accurately gauge WunderGraph feature usage, pain points, and customization across all users.")
```
jensneuse, line 220: rename to "WG_DISABLE_TELEMETRY"?

pkg/telemetry/client.go
```diff
@@ -0,0 +1,160 @@
...
+	req.Header.Set("User-Agent", userAgent)
+	req.Header.Add("Content-Type", "application/json")
+
+	resp, err := c.httpClient.Do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != 200 {
```
jensneuse, line 136: you should also check if the response contains the root field errors

jensneuse (changes requested):

cli/commands/root.go
StarpTech: Why?

pkg/config/config.go
StarpTech: Yes, good point.

pkg/telemetry/client.go
StarpTech: Good catch

cli/commands/root.go
```diff
@@ -2,27 +2,21 @@ package commands
...
 	"github.com/wundergraph/wundergraph/pkg/logging"
-	"github.com/wundergraph/wundergraph/pkg/manifest"
 	"github.com/wundergraph/wundergraph/pkg/node"
-	"github.com/wundergraph/wundergraph/pkg/v2wundergraphapi"
+	"github.com/wundergraph/wundergraph/pkg/telemetry"
+	"go.uber.org/zap"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"time"
```
fiam, line 19: Sort these impors with `make golang-ci-fix`please

testapps/default/.wundergraph/wundergraph.server.ts
```diff
@@ -24,7 +24,7 @@ export default configureWunderGraphServer<HooksConfig, InternalClient, WebhooksC
 	},
 	hooks: {
 		authentication: {
-			postAuthentication: async ({ user }) => {},
+			postAuthentication: async ({ user, internalClient }) => {},
```
fiam, line 27: This seems to be added accidentally

fiam (changes requested):

cli/commands/root.go
StarpTech: Can we run it as part of the recommit hook?

testapps/default/.wundergraph/wundergraph.server.ts
StarpTech: Yes, will delete it

fiam:
> During the implementation, I also removed the hub code that is no longer used because the way how we want to enable users to share and consume API as packages have significantly changed. That allowed me to clean up several places.

Please, try to avoid that in the future. It makes the review more complicated (hence more likely for reviewers to miss something). Also, if we want to bring back the code at some point in the future we can no longer git-revert.

cli/commands/root.go
fiam: We can add a check to enforce the formatting, but I don't think that changing the code automatically is a good idea.

pkg/telemetry/client.go
jensneuse: so?

testapps/default/.wundergraph/wundergraph.server.ts
StarpTech: Done

StarpTech:
> Please, try to avoid that in the future. It makes the review more complicated (hence more likely for reviewers to miss something). Also, if we want to bring back the code at some point in the future we can no longer git-revert.

I see your point, and I agree, but in this case, we agreed on it.

fiam (approved):

pkg/telemetry/client.go
StarpTech: Adressed

jensneuse (approved):

PR #426, opened 2022-12-09
chore(RFC): functions

jensneuse:


rfc/functions/[RFC] functions - 001 - jensneuse.md
```diff
@@ -0,0 +1,305 @@
...
+## Solution
+
+Folders will be reflected in the URL of the operation, and in the generated types.
+
+Given the following folder structure:
+
+```
+.wundergraph
+├── operations
+│   ├── users
```
Comment (StarpTech) [inline rfc/functions/[RFC] functions - 001 - jensneuse.md:41]: It should also be possible to add a folder to implement REST-like versioning e.g. `/v1/users/create`. This is very powerful.

rfc/functions/[RFC] functions - 001 - jensneuse.md
```diff
@@ -0,0 +1,305 @@
...
+/operations/posts/create
+/operations/posts/delete
+/operations/posts/get
+/operations/posts/update
+```
+
+Calling the users/create Operation from the TypeScript client will look like this:
+
+```ts
+const out = await client.mutate({
```
Comment (StarpTech) [inline rfc/functions/[RFC] functions - 001 - jensneuse.md:70]: How is DELETE, and PUT request methods mapped? Or do we ignore them in wundergraph RPC protocol?

rfc/functions/[RFC] functions - 001 - jensneuse.md
```diff
@@ -0,0 +1,305 @@
...
+│   │   ├── get.graphql
+│   │   └── update.graphql
+│   └── posts
+│       ├── create.graphql
+│       ├── delete.graphql
+│       ├── get.graphql
+│       └── update.graphql
+```
+
+The following endpoints will be created:
```
Comment (StarpTech) [inline rfc/functions/[RFC] functions - 001 - jensneuse.md:53]: I prefer to manage functions in a different directory. It might be confusing from the user perspective that only GraphQL operations can have hooks. If a function is never the same as a regular GraphQL operation, we should not mix them. I like to have a functions directory.

rfc/functions/[RFC] functions - 001 - jensneuse.md
```diff
@@ -0,0 +1,305 @@
...
+});
+
+const output = z.object({
+  deleted: z.boolean(),
+});
+
+export default createOperation({
+  input,
+  output,
+  type: 'mutation',
```
Comment (StarpTech) [inline rfc/functions/[RFC] functions - 001 - jensneuse.md:200]: I prefer to have different methods for queries and mutations `createMutationOperation`. This makes it a bit more convenient.

rfc/functions/[RFC] functions - 001 - jensneuse.md
```diff
@@ -0,0 +1,305 @@
...
+├── wundergraph.context.ts
+```
+
+The `wundergraph.context.ts` file will look like this:
+
+It creates a global context object on application start,
+and has an additional handler to inject per-request meta data into the context.
+
+```ts
+import { createContext } from '@wundergraph/sdk/functions';
```
Comment (StarpTech) [inline rfc/functions/[RFC] functions - 001 - jensneuse.md:237]: IMO this is a premature optimization. The user can do whatever he wants in the handler. Creating a TCP connection on every cold start could result in very bad performance.

rfc/functions/[RFC] functions - 001 - jensneuse.md
```diff
@@ -0,0 +1,305 @@
...
+  type: 'mutation',
+  requiresAuth: true,
+  handler: async ({ input, context }) => {
+    if (!context.user.roles.includes('admin')) {
+      throw new Error('Not authorized');
+    }
+    const { id } = input;
+    const user = await context.db.users.findOne({ id });
+    if (!user) {
+      throw new Error('User not found');
```
Comment (StarpTech) [inline rfc/functions/[RFC] functions - 001 - jensneuse.md:209]: It should be possible to work with typed errors that are part of the output type definition. Similar to GraphQL unions.

Comment (StarpTech) [inline rfc/functions/[RFC] functions - 001 - jensneuse.md:209]: What's the expected response when an error is thrown?

Comment (StarpTech) [reply rfc/functions/[RFC] functions - 001 - jensneuse.md:209]: We could provide an onError hook where the user can decide how and when to communicate an error to the client.

rfc/functions/[RFC] functions - 001 - jensneuse.md
```diff
@@ -0,0 +1,305 @@
...
+```
+
+We simply use the full path (excluding the "/operations/" prefix) to create the URL and the operationName, instead of
+just using the filename.
+
+This solves problem "3".
+
+Next, let's say we'd like to add hooks to the users/create Operation to set the input name to uppercase.
+
+We'd extend the existing folder structure by adding a `create.hooks.ts` file:
```
Comment (StarpTech) [inline rfc/functions/[RFC] functions - 001 - jensneuse.md:90]: I think this has nothing to do with functions. I'd handle it in a separate RFC.

rfc/functions/[RFC] functions - 001 - jensneuse.md
```diff
@@ -0,0 +1,305 @@
...
+const output = z.object({
+  deleted: z.boolean(),
+});
+
+export default createOperation({
+  input,
+  output,
+  type: 'mutation',
+  requiresAuth: true,
+  handler: async ({ input, context }) => {
```
Comment (StarpTech) [inline rfc/functions/[RFC] functions - 001 - jensneuse.md:202]: It should be possible to call other functions through the context `context.internalClient`

rfc/functions/[RFC] functions - 001 - jensneuse.md
```diff
@@ -0,0 +1,305 @@
...
+  email: z.string(),
+  title: z.string(),
+  content: z.string(),
+  published: z.boolean(),
+});
+
+const createUser = user.omit({ id: true });
+
+export default createOperation({
+  input: createUser,
```
Comment (StarpTech) [inline rfc/functions/[RFC] functions - 001 - jensneuse.md:155]: I'd put validation in an object

```ts
{
  validation:  {  input, output }
}
```

StarpTech (changes requested):
Great proposal.

rfc/functions/[RFC] functions - 001 - jensneuse.md
```diff
@@ -0,0 +1,305 @@
...
+```
+
+The route for this operation will be `/operations/users/create`.
+
+The `create.ts` file will look like this:
+
+```ts
+import { createOperation } from '../../generated/functions';
+
+const user = z.object({
```
Comment (StarpTech) [inline rfc/functions/[RFC] functions - 001 - jensneuse.md:142]: Some other points:
- We need to figure out how and when to generate the JSON schema from the functions.
- The more functions we create, the slower the cold start because all files are loaded synchronously. Ideally, we can lazy load it or we bundle all functions into a single bundle.

StarpTech (changes requested):

rfc/functions/[RFC] functions - 001 - jensneuse.md
Comment (Pagebakers) [reply rfc/functions/[RFC] functions - 001 - jensneuse.md:209]: When we throw a regular Error or anything else we parse it to `{error: 'message'}`, but it should also be possible to throw built-in error objects like `throw new GraphqlError()` so you can communicate more info in the error.

rfc/functions/[RFC] functions - 001 - jensneuse.md
Comment (Pagebakers) [reply rfc/functions/[RFC] functions - 001 - jensneuse.md:202]: Since we define them as operations this is possible.

rfc/functions/[RFC] functions - 001 - jensneuse.md
Comment (Pagebakers) [reply rfc/functions/[RFC] functions - 001 - jensneuse.md:53]: I think they are 2 distinct things, there's some overlap technically.

1. Functional (custom) operations.
I think this is a very natural and powerful pattern, you write your operation resolvers directly in TS, instead of graphql (later, this hooks in perfectly with the ts ORM) and can continue using the WG client (query/mutate) like you would normally. From a DX perspective it is very similar to writing resolvers and like other frameworks like trpc. They also have a more opiniated API with input validation for example.

2. Generic functions
These are more like lambda (or edge / nextjs api) functions, very similar to webhooks.

rfc/functions/[RFC] functions - 001 - jensneuse.md
Comment (Pagebakers) [reply rfc/functions/[RFC] functions - 001 - jensneuse.md:90]: I like the idea, because the current hooks setup is not ideal. It could be handled in a separate RFC, but it's a good moment to think about the design.

rfc/functions/[RFC] functions - 001 - jensneuse.md
Comment (Pagebakers) [inline rfc/functions/[RFC] functions - 001 - jensneuse.md:155]: Passing the schema directly to input like this makes zod a hard dependency, we'd also want to support other validation libraries.

rfc/functions/[RFC] functions - 001 - jensneuse.md
Comment (Pagebakers) [reply rfc/functions/[RFC] functions - 001 - jensneuse.md:237]: Having access directly to prisma via context is productive, but depending on the application it might add unneeded extra load to all operations.

rfc/functions/[RFC] functions - 001 - jensneuse.md
Comment (jensneuse) [reply rfc/functions/[RFC] functions - 001 - jensneuse.md:41]: you can do this already with this RFC

rfc/functions/[RFC] functions - 001 - jensneuse.md
Comment (jensneuse) [reply rfc/functions/[RFC] functions - 001 - jensneuse.md:70]: We could later on add different HTTP Verbs, although this wouldn't affect the generated clients as the HTTP method is an internal detail. Currently, I'd not try to solve this problem.

rfc/functions/[RFC] functions - 001 - jensneuse.md
Comment (jensneuse) [reply rfc/functions/[RFC] functions - 001 - jensneuse.md:53]: I think it needs to be in the same folder as GraphQL Operations, because in the end, both are just operations. Putting them in different directories would cause conflicts. We also don't yet know how this will be used. I'd just release it as is and see how people react.. Everything else is assumptions.

rfc/functions/[RFC] functions - 001 - jensneuse.md
Comment (jensneuse) [reply rfc/functions/[RFC] functions - 001 - jensneuse.md:200]: Yes, good idea.

rfc/functions/[RFC] functions - 001 - jensneuse.md
Comment (jensneuse) [reply rfc/functions/[RFC] functions - 001 - jensneuse.md:237]: We can actually add this feature later.

rfc/functions/[RFC] functions - 001 - jensneuse.md
Comment (jensneuse) [reply rfc/functions/[RFC] functions - 001 - jensneuse.md:209]: I agree

rfc/functions/[RFC] functions - 001 - jensneuse.md
Comment (jensneuse) [reply rfc/functions/[RFC] functions - 001 - jensneuse.md:90]: Yes, can be done in a separate RFC/Implementation.

rfc/functions/[RFC] functions - 001 - jensneuse.md
Comment (jensneuse) [reply rfc/functions/[RFC] functions - 001 - jensneuse.md:202]: context/ctx should be the ctx like in all other hooks, so yes, we'd inject the internal client

rfc/functions/[RFC] functions - 001 - jensneuse.md
Comment (jensneuse) [reply rfc/functions/[RFC] functions - 001 - jensneuse.md:142]: JSON Schema should be generated during development (up) or build step (generate).

rfc/functions/[RFC] functions - 001 - jensneuse.md
Comment (StarpTech) [reply rfc/functions/[RFC] functions - 001 - jensneuse.md:155]: It's just another fixed field. I don't understand how it will cause issues or make it "a lot more" verbose.

StarpTech (changes requested):

rfc/functions/[RFC] functions - 001 - jensneuse.md
```diff
@@ -0,0 +1,305 @@
...
+This means, you'll easily run into conflicts when you have two operations with the same name in different folders.
+
+That said, there's another big drawback with the current approach.
+You're only able to create operations by defining GraphQL Operations.
+This means, the frontend contract will always depend on types from the virtual Graph.
+While quite efficient in most use cases,
+there are situations where you'd like to have clear abstraction between public and internal APIs,
+or you might just want to define custom business logic without depending on one single Operation from the virtual Graph.
+
+To sum up the problems:
```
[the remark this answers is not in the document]
Comment (jensneuse) [reply rfc/functions/[RFC] functions - 001 - jensneuse.md:21]: I think they solve quite different use cases.
A WebHook is more or less an HTTP handler.
An Operation Handler is a statically typed event handler, with a typed input and output.
Operation handlers can define all things for Operations, like Authentication, etc...
There's actually less overlap than one might think.
I think it's best to keep them separate.

rfc/functions/[RFC] functions - 001 - jensneuse.md
Comment (StarpTech) [reply rfc/functions/[RFC] functions - 001 - jensneuse.md:53]: It will be possible to configure function operations in "wundergraph.operation.ts" correct?

rfc/functions/[RFC] functions - 001 - jensneuse.md
Comment (StarpTech) [reply rfc/functions/[RFC] functions - 001 - jensneuse.md:70]: ok

rfc/functions/[RFC] functions - 001 - jensneuse.md
Comment (StarpTech) [reply rfc/functions/[RFC] functions - 001 - jensneuse.md:41]: ok, so empty directories aren't ignored, and the total file path is mapped to the operation name.

rfc/functions/[RFC] functions - 001 - jensneuse.md
Comment (jensneuse) [reply rfc/functions/[RFC] functions - 001 - jensneuse.md:155]: At the same time, it's not clear how this is really different then the original idea.
Keep in mind that it's more than just validation.
The input and output definition is responsible to infer the types for the handler function as well as for the generation of the JSON Schema for the clients AND input validation, but not just validation by itself.

rfc/functions/[RFC] functions - 001 - jensneuse.md
Comment (jensneuse) [reply rfc/functions/[RFC] functions - 001 - jensneuse.md:209]: As a user, I don't really want to know anything about GraphQL.
I think there should be a way to return a public error, which will then be returned in the errors object.
For internal errors we could allow to Throw.

rfc/functions/[RFC] functions - 001 - jensneuse.md
Comment (jensneuse) [reply rfc/functions/[RFC] functions - 001 - jensneuse.md:53]: What do you mean by "configure"? In theory yes, but the function could configure itself, e.g. caching. There's no need to add configuration in multiple places.

rfc/functions/[RFC] functions - 001 - jensneuse.md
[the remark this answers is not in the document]
Comment (fiam) [reply rfc/functions/[RFC] functions - 001 - jensneuse.md:21]: I would also keep them separate

rfc/functions/[RFC] functions - 001 - jensneuse.md
Comment (fiam) [reply rfc/functions/[RFC] functions - 001 - jensneuse.md:53]: I agree with Eelco here. I would implement both  “typescript operations” (these are analog to graphql operations and we should be able to translate a graphql operation to a typescript one automatically) and “functions” (which take the raw request and produce a response)

rfc/functions/[RFC] functions - 001 - jensneuse.md
Comment (fiam) [reply rfc/functions/[RFC] functions - 001 - jensneuse.md:90]: Separate RFC and PR, but we can discuss it here. And let it be notes that I like this design.

rfc/functions/[RFC] functions - 001 - jensneuse.md
Comment (fiam) [reply rfc/functions/[RFC] functions - 001 - jensneuse.md:209]: Throwing an error should be never reported to the client, otherwise it's very easy to leak information on unintentional errors. They should be logged and shown in the dashboard, though. If the intent is to return an error, we should have some mechanism for that.

rfc/functions/[RFC] functions - 001 - jensneuse.md
Comment (StarpTech) [reply rfc/functions/[RFC] functions - 001 - jensneuse.md:155]: Even so, I'd wish to have better names for it. Lets take fastify as an example

```ts
const schema = {
  body: bodyJsonSchema,
  querystring: queryStringJsonSchema,
  params: paramsJsonSchema,
  headers: headersJsonSchema
}

fastify.post('/the/url', { schema }, handler)
```

Maybe we call it `schema`?

rfc/functions/[RFC] functions - 001 - jensneuse.md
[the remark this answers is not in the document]
Comment (StarpTech) [reply rfc/functions/[RFC] functions - 001 - jensneuse.md:21]: Good points, but to make webhooks more convenient, we also need to consider type safety and validation.

rfc/functions/[RFC] functions - 001 - jensneuse.md
Comment (JivusAyrus) [reply rfc/functions/[RFC] functions - 001 - jensneuse.md:53]: Even I agree with Eelco. When I heard functions, I thought it would be like the second point Eelco mentioned.
But I like the approach. Does this mean we will remove the support for .graphql files when .ts files can do everything the graphql files can? Or will we have support for both?

Comment (jensneuse) [reply rfc/functions/[RFC] functions - 001 - jensneuse.md:53]: No, we will not remove support for GraphQL. You can use both.

rfc/functions/[RFC] functions - 001 - jensneuse.md
Comment (YuriBuerov) [reply rfc/functions/[RFC] functions - 001 - jensneuse.md:209]: @fiam  100% true

PR #428, opened 2022-12-09
feat(ci): add CI workflow for examples

fiam:


scripts/test-example.sh
```diff
@@ -0,0 +1,109 @@
+#!/bin/sh
+
+usage()
```
StarpTech, line 3: Awesome work! But we can drastically simplify it by using https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#expanding-or-adding-matrix-configurations

In that way, we can have one workflow that runs all examples in parallel with propert action integration (status, logging....)

StarpTech (changes requested):

scripts/test-example.sh
StarpTech: Why you can't use one job per example? In that way, you don't need to set up stuff that only one example required of 5. We can have up to 256 parallel jobs :smile:

StarpTech: Done

YuriBuerov (approved):

StarpTech (approved):
LGTM

PR #432, opened 2022-12-10
fix: add missing dependency to our nextjs package

fiam:
object-hash was accidentally removed at some point, causing import
errors


StarpTech (approved):
LGTM

PR #435, opened 2022-12-11
fix: missing generated files in golang-client example

fiam:


examples/golang-client/.wundergraph/wundergraph.config.ts
```diff
@@ -14,6 +14,9 @@ configureWunderGraphApplication({
 	server,
 	operations,
 	codeGenerators: [
+		{
+			templates: templates.typescript.all,
```
StarpTech, line 18: Why is that necessary?

StarpTech (changes requested):

examples/golang-client/.wundergraph/wundergraph.config.ts
fiam: `wundergraph.operations.ts` and `wundergraph.server.ts` reference the generated typescript files, which makes `tsc` fail

StarpTech: ok

StarpTech (approved):

Pagebakers (approved):

Issue #436, opened 2022-12-11
OAS: error on parameter names and enum names with '.'


emrosenf:
### Bug description

A parameter name with a '.', e.g. `sort.order`, or an enum name like `sort_order.asc` will throw an error:

```
 FAIL  ./index.test.ts
  ✕ arbitrary type (68 ms)

  ● arbitrary type

    GraphQLError: Syntax Error: Unexpected character: ".".

      25 |      let customScalarTypeFields: SingleTypeField[] = [];
      26 |
    > 27 |      const ast = parse(schemaSDL);
         |                       ^
      28 |      const cleanAst = visit(ast, {
      29 |              ObjectTypeDefinition: {
      30 |                      enter: (node) => {

      at syntaxError (../../../../node_modules/.pnpm/graphql@16.6.0/node_modules/graphql/error/syntaxError.js:15:10)
      at readNextToken (../../../../node_modules/.pnpm/graphql@16.6.0/node_modules/graphql/language/lexer.js:420:40)
      at Lexer.lookahead (../../../../node_modules/.pnpm/graphql@16.6.0/node_modules/graphql/language/lexer.js:84:29)
      at Lexer.advance (../../../../node_modules/.pnpm/graphql@16.6.0/node_modules/graphql/language/lexer.js:67:38)
      at Parser.advanceLexer (../../../../node_modules/.pnpm/graphql@16.6.0/node_modules/graphql/language/parser.js:1536:31)
      at Parser.expectToken (../../../../node_modules/.pnpm/graphql@16.6.0/node_modules/graphql/language/parser.js:1393:12)
      at Parser.parseName (../../../../node_modules/.pnpm/graphql@16.6.0/node_modules/graphql/language/parser.js:108:24)
      at Parser.parseInputValueDef (../../../../node_modules/.pnpm/graphql@16.6.0/node_modules/graphql/language/parser.js:869:23)
      at Parser.optionalMany (../../../../node_modules/.pnpm/graphql@16.6.0/node_modules/graphql/language/parser.js:1492:28)
      at Parser.parseArgumentDefs (../../../../node_modules/.pnpm/graphql@16.6.0/node_modules/graphql/language/parser.js:855:17)
      at Parser.parseFieldDefinition (../../../../node_modules/.pnpm/graphql@16.6.0/node_modules/graphql/language/parser.js:837:23)
[...]

PR #437, opened 2022-12-11
fix: allow periods in parameter and enum names #436

emrosenf:
This is a partial fix for #436. It is directionally correct and causes the test to pass.

However, `index.ts` is not well documented enough for me to know if we need additional work on L300:L345 or L924 to map the new GraphQL Schema parameter or enum name back to the original name with periods from the OAS


jensneuse:
@emrosenf please update the test snapshots to pass
Make sure that no existing behavior is broken, e.g. existing tests (snapshots) should not change.

jensneuse:
So, I've checked the code. We need to introduce a "mapping" to the OAS DataSource so that we're mapping the enums from the REST API to the GraphQL enums and vice versa. This needs to be done in the Go side. Once implemented, we have to expose an API to the SDK so it can be configured via TypeScript. Then we can update the OAS config function to also add the mappings. Quite some work actually...

emrosenf:
Just my luck that the first OAS I try has this issue. LMK if I can help, but it sounds like this might be a job for someone who better understands all of the internal architecture. At least it's flagged and you know how to fix it.

PR #441, opened 2022-12-12
chore(RFC): typesafe environment variables

Pagebakers:
#### Checklist

- [ ] run `make test`


rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md 
```diff
@@ -0,0 +1,84 @@
...
+A new `wundergraph.env.ts` configuration where all available environment variables can be configured. The configuration supports a schema, so the values can be validated.
+
+### Configuration
+
+```ts
+import { z } from 'zod';
+import { configureEnv } from '@wundergraph/sdk';
+
+const schema = z.object({
+	NODE_ENV: z.enum(['development', 'test', 'production']),
```
Comment (jensneuse) [inline rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md :27]: It would be nice if we could also add meta data, like a description, which can be used in WunderGraph cloud so the user can better understand the purpose of the variable.

rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md 
```diff
@@ -0,0 +1,84 @@
...
+
+Instead of using `process.env.VAR` we now have typesafe access to our environment variables like so;
+
+```ts
+import { env } from '.wundergraph/wundergraph.env';
+
+env.WG_NODE_URL; // http://localhost:9991
+env.OAUTH_CLIENT_ID;
+
+new EnvironmentVariable(env.WG_SERVER_PORT, '9992');
```
Comment (jensneuse) [inline rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md :58]: I have an idea of how we improve the DX of this.
We should force the user to use `env.XXX` in `wundergraph.config.ts` instead of allowing strings.
Now, when running inside `wundergraph.config.ts` using an `env.XXX` variable will resolve to what before was `new EnvironmentVariable`.
When using `env.XXX` in the fastify server, it will resolve to the actual value. WDYT?
@StarpTech is this possible (technically)

rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md 
Comment (jensneuse) [reply rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md :27]: It should also be possible to create a default value.
E.g.:

```typescript
const schema = createSchema({
  OAUTH_CLIENT_ID: {
    schema: z.string.optional(),
    default: "foo",
    description: "This is the Auth99 client ID"
  }
});
```

Comment (jensneuse) [reply rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md :27]: Maybe this is a bit verbose and we can improve it?

rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md
```diff
@@ -24,12 +24,12 @@ import { z } from 'zod';
 import { configureEnv } from '@wundergraph/sdk';
 
 const schema = z.object({
-	NODE_ENV: z.enum(['development', 'test', 'production']),
-	OAUTH_CLIENT_ID: z.string(),
+	NODE_ENV: z.enum(['development', 'test', 'production']).default('development'),
+	OAUTH_CLIENT_ID: z.string().describe('OAuth Client ID'),
 	SQLITE_DB: z.string().optional(),
 });
```
Comment (jensneuse) [inline rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md:30]: It looks perfect!

jensneuse:
@Pagebakers great work!

rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md
```diff
@@ -0,0 +1,84 @@
...
+| `WG_SERVER_URL`      | The URL of the `WunderGraph Server`.                    | `http://localhost:9992` |
+| `WG_SERVER_HOST`     | The host of the `WunderGraph Server`.                   | `127.0.0.1`             |
+| `WG_SERVER_PORT`     | The port of the `WunderGraph Server`.                   | `9992`                  |
+
+### Usage
+
+Instead of using `process.env.VAR` we now have typesafe access to our environment variables like so;
+
+```ts
+import env from '.wundergraph/wundergraph.env';
```
Comment (StarpTech) [inline rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md:53]: I don't like the fact that we workaround from using process.env which is very natural. IMO we only need to ensure that the env schema is correct at build and runtime. Why should we provide a different way to access them?

rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md
```diff
@@ -0,0 +1,84 @@
...
+
+A new `wundergraph.env.ts` configuration where all available environment variables can be configured. The configuration supports a schema, so the values can be validated.
+
+### Configuration
+
+```ts
+import { z } from 'zod';
+import { configureEnv } from '@wundergraph/sdk';
+
+const schema = z.object({
```
Comment (StarpTech) [inline rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md:26]: How can we distinguish between build and runtime env variables? We need that separation so the node doesn't throw for things that aren't needed at runtime.

StarpTech (changes requested):

rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md
Comment (fiam) [reply rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md:26]: I like the idea of a type-safe environment a lot, but I'm wondering: doesn't TypeScript type's system allow us to express 90% of the required use cases? For example, this specific environment could be defined as:

```typescript
interface Env {
    NODE_ENV: "development" | "test" | "production";
   OAUTH_CLIENT_ID: string;
   SQLITE_DB?: string;
}

Comment (jensneuse) [reply rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md:26]: what about default values? what about adding a description for the UI to explain the variable?

Comment (Pagebakers) [reply rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md:26]: that way we can't have runtime errors for missing variables, but also don't get the benefit of reading the schema in the cloud and prefil the environment form in the UI (nice DX).

File: rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md [hunk 6]
```diff
@@ -0,0 +1,102 @@
...
+| -------------------- | ------------------------------------------------------- | ----------------------- |
+| `WG_LOG_LEVEL`       | The log level of the `WunderNode`/`WunderGraph Server`. | `info`                  |
+| `WG_NODE_URL`        | The internal URL of the `WunderNode`.                   | `http://localhost:9991` |
+| `WG_PUBLIC_NODE_URL` | The publicly available URL of the `WunderNode`.         | `http://localhost:9991` |
+| `WG_NODE_HOST`       | The host of the `WunderNode`.                           | `127.0.0.1`             |
+| `WG_NODE_PORT`       | The port of the `WunderNode`.                           | `9991`                  |
+| `WG_SERVER_URL`      | The URL of the `WunderGraph Server`.                    | `http://localhost:9992` |
+| `WG_SERVER_HOST`     | The host of the `WunderGraph Server`.                   | `127.0.0.1`             |
+| `WG_SERVER_PORT`     | The port of the `WunderGraph Server`.                   | `9992`                  |
+
```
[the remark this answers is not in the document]
Comment (Pagebakers) [reply rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md:47]: They are already accessible to the user, and they can find them using the `wgEnv` variable exported from the SDK.

This would just add them to the type definition, so they are available in `env.`

rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md 
Comment (StarpTech) [reply rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md ]: That sounds to me like a magical thing. You want to parse `.env` usages and create the equivalent `new EnvironmentVariable` to the file? I'd try to keep it boring. We have one file where the user can specify build and runtime environment variables. We generate a file to access the actual value. This file is imported where needed, exactly as in your example.

```ts
import { configureEnv, z } from '@wundergraph/sdk';

// 1. Define you variables with validation and when they are needed (by all are required for all stages)
export default configureEnv({
  COOKIE_SECRET: z.requiredForRun().string().describe('Cookie secret'),
  DATABASE_URL: z.required().string().describe('Database URL')
});

// 2. Access it like a regular import

import { DATABASE_URL } from './generated/env';

// 3. From the wundernode perspective
// The node will validates only COOKIE_SECRET because it is required to run the server
```

rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md
Comment (StarpTech) [reply rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md:53]: I think the approach to using just process.env is too limited. What if we want to support coercion? string to int. It would be odd if process.env is not a string. Your argument regarding collision is also valid though.

rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md
Comment (fiam) [reply rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md:26]: You can use tsdoc comments to give them human-friendly names as well as specifying default values. The tsdoc parser is open source. This way developers don’t need to learn a new syntax, which removes a lot of friction.

rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md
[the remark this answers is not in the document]
Comment (fiam) [reply rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md:47]: My point is, does the app code need to care about those variables?  Users can set them to configure their app, but there’s not much that they can do with them at runtime.

rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md 
Comment (jensneuse) [reply rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md ]: not too bad

rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md
Comment (Pagebakers) [reply rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md:53]: Yeah I prefer having a custom env object for typesafe access. Coercion is a nice extra that way as well indeed. 

But then i'm not sure what you mean with `I don't like the fact that we workaround from using process.env `, your comments seem contradicting.

rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md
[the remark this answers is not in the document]
Comment (Pagebakers) [reply rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md:47]: Either way they are available in process.env no?

rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md 
Comment (StarpTech) [reply rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md ]: Decision: There is only one set of env variables (no separation between build and runtime), and we need to ensure that defaults are loaded into ENV before validation.

Comment (Pagebakers) [reply rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md ]: I rather not add proprietary functionality on top of zod.

If we make sure WG defaults are loaded into process.env is set before the validation is run we don't need to worry between runtime/buildtime required variables, we can just use one set.

StarpTech (approved):
LGTM

rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md
```diff
@@ -0,0 +1,91 @@
...
+new EnvironmentVariable('OAUTH_CLIENT_ID');
+```
+
+### EvironmentVariable
+
+When the config is generated we also generate a type declaration for `EnvironmentVariable`.
+
+```
+declare global {
```
Comment (StarpTech) [inline rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md:59]: Why should we generate it if we the user can import them directly?

StarpTech (changes requested):

rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md
Comment (StarpTech) [reply rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md:59]: Or do you mean for the type of `new EnvironmentVariable`?

StarpTech (changes requested):

rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md
Comment (Pagebakers) [reply rfc/typesafe-env-vars/[RFC] typesafe env vars - 001 - Pagebakers.md:59]: Yes that's for the first arg of EnvironmentVariable('this is typed')

StarpTech (approved):
LGTM

jensneuse (approved):

Pagebakers:
@jensneuse Can you merge this, the build_test doesn't run

PR #442, opened 2022-12-12
chore(RFC): wunderbase

jensneuse:


rfc/wunderbase/[RFC]wunderbase-001-jensneuse.md
```diff
@@ -0,0 +1,117 @@
...
+        ├── index.ts
+        └── schema.prisma
+```
+
+The config file will look like this:
+
+```ts
+import { configureWunderBase } from '@wundergraph/sdk';
+
+export default configureWunderBase({
```
Comment (Pagebakers) [inline rfc/wunderbase/[RFC]wunderbase-001-jensneuse.md:37]: Why a separate file and not define it in wundergraph.config.ts?

rfc/wunderbase/[RFC]wunderbase-001-jensneuse.md
```diff
@@ -0,0 +1,117 @@
...
+
+To create a new WunderBase,
+add a new subdirectory to wunderbase, e.g. `.wundergraph/wunderbase/app` and add a config file (`index.ts`) and a schema file `schema.prisma`.
+
+The file tree will look like this:
+
+```
+.wundergraph
+└── wunderbase
+    └── app
```
Comment (Pagebakers) [inline rfc/wunderbase/[RFC]wunderbase-001-jensneuse.md:27]: I thought we got rid of `app`. One wunderbase per project and one per environment seems OK for the use case no?

Pagebakers (changes requested):

rfc/wunderbase/[RFC]wunderbase-001-jensneuse.md
Comment (jensneuse) [reply rfc/wunderbase/[RFC]wunderbase-001-jensneuse.md:27]: Why not have multiple databases? E.g. an SQLite for user data and PostgreSQL for analytics

rfc/wunderbase/[RFC]wunderbase-001-jensneuse.md
Comment (YuriBuerov) [reply rfc/wunderbase/[RFC]wunderbase-001-jensneuse.md:37]: I agree with @Pagebakers, `wundergraph.config.ts` is not so huge, as for me it's more convenient to have configurations in one single place.

rfc/wunderbase/[RFC]wunderbase-001-jensneuse.md
Comment (YuriBuerov) [reply rfc/wunderbase/[RFC]wunderbase-001-jensneuse.md:27]: having multiple databases sounds reasonable.  And one more thing - `.wundergraph` is optional now (which I don't like, but the decision was made), and it's not well documented. For example, the WG project could look like this: https://github.com/bmcilw1/turbo-wunder-svelte/tree/main/apps/api. So when we use it in docs, and RFC, it could be confusing.

rfc/wunderbase/[RFC]wunderbase-001-jensneuse.md
Comment (Pagebakers) [reply rfc/wunderbase/[RFC]wunderbase-001-jensneuse.md:37]: Not that big of an issue here yet. Since wundergraph.config.ts is just a typescript file, you can stil split up datasource configs in a separate folder if needed and then import them in the main config..

rfc/wunderbase/[RFC]wunderbase-001-jensneuse.md
Comment (Pagebakers) [reply rfc/wunderbase/[RFC]wunderbase-001-jensneuse.md:27]: Right, you want to be able to define multiple prisma schema in wunderbase? not only for sqlite?

Comment (jensneuse) [reply rfc/wunderbase/[RFC]wunderbase-001-jensneuse.md:27]: exactly
in this case, what's the convention to put these prisma schemas?
why not just wunderbase and the convention makes it clear and the user doesn't have to think about file paths, etc...

rfc/wunderbase/[RFC]wunderbase-001-jensneuse.md
Comment (fiam) [reply rfc/wunderbase/[RFC]wunderbase-001-jensneuse.md:37]: I'd go with a `wunderbase.config.ts` file inside `.wundergraph` instead. We can make `configureWunderBase()` take an array of configuration objects explicitly linking to the schema e.g.

```configureWunderBase([
     {
          namespace: 'app',
          schema: './prisma.schema',
          ...
          <future-configuration-options e.g. for cloud automatic backups<
     }
])
```

This supports multiple databases and allows users to organize the files as they see fit

rfc/wunderbase/[RFC]wunderbase-001-jensneuse.md
```diff
@@ -0,0 +1,117 @@
...
+	ConfigurationVariable databaseURL = 1;
+	string prismaSchema = 2;
+	string graphqlSchema = 3;
+	// closeTimeoutSeconds define that the database connection will be closed after the given amount of seconds of inactivity
+	int64 closeTimeoutSeconds = 4;
+	repeated SingleTypeField jsonTypeFields = 5;
+	repeated string jsonInputVariables = 6;
+  bool isWunderBase = 7;
+  string wunderBaseURL = 8;
+}
```
Comment (fiam) [inline rfc/wunderbase/[RFC]wunderbase-001-jensneuse.md:103]: Do we really need so much information? I'd think that namespace and schema path should be enough, the rest can be read from the files.

Comment (jensneuse) [reply rfc/wunderbase/[RFC]wunderbase-001-jensneuse.md:103]: yes, it's needed

Pagebakers:
Closing this for now, if you continue working on this, re-open it in the rfcs repo

PR #447, opened 2022-12-13
chore(RFC): functions 004 jensneuse

jensneuse:


testapps/default/functionsv2.ts
```diff
@@ -0,0 +1,100 @@
...
+class Client<Queries> {
+	public useQuery<Query extends keyof Queries>(query: Query, variables: Queries[Query]['args']): Queries[Query]['response'] {
+		return Promise.resolve() as any;
+	}
+}
+
+const client = new Client<Queries>()
+
+const out = await client.useQuery('x', {
+	x: 1
```
jensneuse, line 88: This is currently optional.
How to make it mandatory?
@StarpTech @fiam @Pagebakers @thisisnithin @JivusAyrus

StarpTech: I don't understand the question. Do you mean the field is optional based on your types?

StarpTech: Maybe? Because `never` would allow no-value.
```ts
type ExtractInput<B> = B extends Builder<T, any>;
```

Pagebakers: Yeah it's an issue in the type it's using input? instead of a conditional type. Not strict enough.

testapps/default/functionsv2.ts
```diff
@@ -0,0 +1,100 @@
...
+				data: x,
+				xPlusOne: x + 1
+			}
+		})
+
+const userByID = 
+	createOperation()
+	.query()
+	.input(z.object({userByID: z.string()}).required())
+	.handler(async ({userByID}) => {
```
Pagebakers, line 60: handler -> query

testapps/default/functionsv2.ts
```diff
@@ -0,0 +1,100 @@
...
+		return {
+			id: userByID,
+			name: 'test',
+			age: 20,
+		}
+	})
+
+interface Queries {
+	x: {
+		args: ExtractInput<typeof getX>,
```
Pagebakers, line 70: why is it called args and not input?

testapps/default/functionsv2.ts
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	public mutation() {
+		this.type = 'mutation'
+		return this
+	}
+
+	public input<I extends any>(input: I) {
+		this.inputSchema = input
+		return this as Builder<z.infer<I>, Response>
```
Pagebakers, line 21: Ahhhh nice one

Pagebakers:
Im missing is RBAC

```ts
import { operation } from '../generated/operations';
export default operation()
  .input(z.object({ city: z.string() }))
  .authorize({
    requireMatchAll: ['superadmin', 'user'],
  })
  .query(async ({ input, internalClient }) => {
    const result = await internalClient.Weather({ name: input.city });
    return result;
  });
```

And what about subscriptions?

PR #456, opened 2022-12-15
fix: csrf and authenticated uploads

thisisnithin:
Will add example after merge as a separate PR

#### Checklist

- [x] run `make test`

### DIFF
--- a/packages/sdk/src/client/client.ts
+++ b/packages/sdk/src/client/client.ts
@@ -331,6 +331,13 @@ export class Client {
 		}
 	}
 
+	private async isAuthenticatedByCookie() {
+		const response = await this.fetchJson(`${this.options.baseURL}/auth/cookie/user`, {
+			method: 'GET',
+		});
+		return !!response.ok;
+	}
+
 	/**
 	 * Uploads one or more files to the server. Authentication is required. The method throws an error if the files
 	 * could not be uploaded for any reason. If the upload was successful, your return a list
@@ -343,19 +350,28 @@ export class Client {
 				formData.append('files', file);
 			}
 		}
-		const csrfToken = await this.getCSRFToken();
+
+		let csrfToken = '';
+		const isAuthenticated = await this.isAuthenticatedByCookie();
+		if (isAuthenticated) {
+			csrfToken = await this.getCSRFToken();
+		}
 
 		const params = new URLSearchParams({
 			wg_api_hash: this.options.applicationHash,
 		});
 
+		let headers = {};
+		if (csrfToken) {
+			headers = {
+				'X-CSRF-Token': csrfToken,
+			};
+		}
+
 		const response = await this.fetch(
 			this.addUrlParams(`${this.options.baseURL}/s3/${config.provider}/upload`, params),
 			{
-				headers: {
-					// Dont set the content-type header, the browser will set it for us + boundary
-					'X-CSRF-Token': csrfToken,
-				},
+				headers,
 				body: formData,
 				method: 'POST',
 				signal: config.abortSignal,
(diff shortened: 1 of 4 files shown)


StarpTech (approved):
Good work.

pkg/authentication/authentication.go
```diff
@@ -844,24 +844,20 @@ type CSRFConfig struct {
...
-					csrf.SameSite(csrf.SameSiteStrictMode),
-					csrf.ErrorHandler(&CSRFErrorHandler{
-						InsecureCookies: config.InsecureCookies,
-					}),
-				)
-				csrfMiddleware(unprotected).ServeHTTP(w, r)
-				return
-			}
-			unprotected.ServeHTTP(w, r)
+			domain := removeSubdomain(sanitizeDomain(r.Host))
```
fiam, line 847: This seems incorrect to me. This middleware is applied unconditionally [here](https://github.com/wundergraph/wundergraph/blob/737bc30682068a1f8af9c7e75ec6e88dbab284c2/pkg/apihandler/apihandler.go#L1693), so the effect is that all handlers will require CSRF protection.

Right now, CSRF is coupled with authentication, but if we're going to support it for unauthenticated users we need some means to signal the handlers that should be CSRF protected.

fiam (changes requested):

pkg/authentication/authentication.go
```diff
@@ -844,24 +844,20 @@ type CSRFConfig struct {
 func NewCSRFMw(config CSRFConfig) func(handler http.Handler) http.Handler {
 	return func(unprotected http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			if user := UserFromContext(r.Context()); user != nil && user.FromCookie {
```
StarpTech, line 847: As discussed in the daily. We misunderstood the situation. A CSRF should only be required when the User is authenticated. The generated client should not issue a CSRF request when the operation doesn't require auth.

StarpTech (changes requested):

pkg/authentication/authentication.go
thisisnithin: Reverted

pkg/authentication/authentication.go
thisisnithin: Reverted and added a check on the client

packages/sdk/src/client/client.ts
```diff
@@ -343,19 +350,28 @@ export class Client {
 				formData.append('files', file);
 			}
 		}
-		const csrfToken = await this.getCSRFToken();
+
+		let csrfToken = '';
+		const isAuthenticated = await this.isAuthenticatedByCookie();
```
StarpTech, line 355: I think that can be improved. Right now, we need auth for uploads. It would be enough to throw an error when the crsf endpoint does not return 200

```ts
throw new Error('You must be authenticated in order to upload files')
```

StarpTech (changes requested):

packages/sdk/src/client/client.ts
StarpTech: The upload endpoint will already throw an unauthentication error.

StarpTech (changes requested):

packages/sdk/src/client/client.ts
thisisnithin: There is no change to be done at all then on the client, making the route require auth on the server should be enough. CSRF function already throws the error, the point is to not make that request at all right?

thisisnithin: Done

StarpTech (approved):
LGTM

fiam (approved):
LGTM

PR #460, opened 2022-12-16
chore: add e2e tests to Next.js example

fiam:
Fixes ENG-880


scripts/test-examples/single.sh
```diff
@@ -116,6 +116,19 @@ if test ! -z "${docker_compose_yml}" && test -f ${docker_compose_yml}; then
     cd `dirname ${docker_compose_yml}` && docker-compose down && cd -
 fi
 
+
+# If we have playwright based tests, run them
+# Note that we always run these with npx
+if test -f playwright.config.ts; then
```
StarpTech, line 122: I refactored that in https://github.com/wundergraph/wundergraph/pull/461. We should not guess actions based on a file. We should use npm scripts with conventions. You can run the e2e tests as part of the `test` script with `run-p`

examples/nextjs/playwright.config.ts
```diff
@@ -0,0 +1,108 @@
...
+	/* Configure projects for major browsers */
+	projects: [
+		{
+			name: 'chromium',
+			use: {
+				...devices['Desktop Chrome'],
+			},
+		},
+
+		// {
```
StarpTech, line 55: Remove. Let's try to work with the simplest config file here because we have to duplicate it in every example.

examples/nextjs/playwright.config.ts
```diff
@@ -0,0 +1,108 @@
+import type { PlaywrightTestConfig } from '@playwright/test';
+import { devices } from '@playwright/test';
+
+/**
+ * Read environment variables from file.
+ * https://github.com/motdotla/dotenv
+ */
+// require('dotenv').config();
```
StarpTech, line 8: Enable it. This is compatible with the new approach in https://github.com/wundergraph/wundergraph/pull/461 where we load env from .env

examples/nextjs-react-query/playwright.config.ts
```diff
@@ -0,0 +1,94 @@
...
+		//   },
+		// },
+	],
+
+	/* Folder for test artifacts such as screenshots, videos, traces, etc. */
+	// outputDir: 'test-results/',
+
+	/* Run your local dev server before starting the tests */
+	webServer: {
+		command: 'npm run start',
```
StarpTech, line 89: I think this will collide. `start` was not introduced for CI, only for local development. This is great for local testing only. For CI, it should work because `start:services` and `wait-on:services` ensure that the env is bootstrapped.

examples/nextjs-react-query/pages/index.tsx
```diff
@@ -34,7 +34,7 @@ const Home: NextPage = () => {
 									This is the result of your{' '}
 									<code className="font-mono font-medium text-amber-500 font-bold">Dragons</code> operation.
 								</p>
-								<ul>
+								<ul id="result">
```
StarpTech, line 37: `data-test-id` attributes are used for those markers. Playwright has native support for it.

StarpTech (changes requested):

examples/nextjs/playwright.config.ts
fiam: We don't need `.env` because we're calling `npm start` which ends up calling `wunderctl up` and setting dummy secrets. Keep in mind that end to end tests might also need to run the Next.js server.

examples/nextjs-react-query/playwright.config.ts
fiam: `start` works fine for CI.  Playwright kills all the children and doesn't leave any orphans.

examples/nextjs-react-query/pages/index.tsx
fiam: Changed to `data-testid`

examples/nextjs/playwright.config.ts
fiam: Do you mean we should remove all comments?

StarpTech: I think the comments are good, but we should only have the props we need for the example to work.

examples/nextjs/playwright.config.ts
StarpTech: I thought it could be a quick win whenever you need to reference an ENV in your e2e tests you would be able to use .env.

examples/nextjs-react-query/playwright.config.ts
StarpTech: Then why don't we just use `start` for the CI?

examples/nextjs/playwright.config.ts
StarpTech: Delete it.

examples/nextjs/playwright.config.ts
fiam: Deleted

examples/nextjs-react-query/playwright.config.ts
fiam: Changed the script to run `start` once, then run both integration and end-to-end tests before stopping the server

scripts/test-examples/single.sh
fiam: Changed to package script

scripts/test-examples/single.sh
```diff
@@ -65,60 +59,54 @@ if test ${update_package_json} = "yes"; then
...
+npm start &
+pid=$!
 
-if grep -q '"setup"' package.json; then
-    ${npm} run setup
-fi
+trap "kill_with_children ${pid}" EXIT
+
+# Wait for code generation to complete
+while ! test -f .wundergraph/generated/wundergraph.schema.graphql; do
```
StarpTech, line 84: What happens when it errors? Forever loop?

StarpTech (changes requested):

scripts/test-examples/single.sh
StarpTech: Or does will it always fail before on any error?

fiam: There's a `set -e` after the initial checks, so any command that fails will abort the script with a non-zero exit code

StarpTech: ok

StarpTech (approved):
LGTM

scripts/test-examples/single.sh
fiam: > There's a `set -e` after the initial checks, so any command that fails will abort the script with a non-zero exit code

Replying to myself: this ended up to not be correct for background commands. Added a check while waiting for code generation to complete.

Aenimus (approved):

Aenimus (changes requested):
start:serverCI test is failing

Aenimus (approved):

PR #461, opened 2022-12-17
feat: setup a default .env file before start, improve tests

StarpTech:
#### Checklist

- [ ] run `make test`


jensneuse (approved):

jensneuse (approved):

scripts/test-examples/single.sh
```diff
@@ -107,13 +86,9 @@ elif grep -q '"build"' package.json; then
     ${npm} run build
 fi
 
-if test ! -z ${services_pid}; then
```
fiam, line 110: This needs to stay. Otherwise examples like that run servers (e.g. graphql-apollo-subscriptions/package.json) will leave processes around.

fiam (changes requested):

scripts/test-examples/single.sh
StarpTech: I kill it in the cleanup step.

examples/openapi-mocking/package.json
```diff
@@ -4,9 +4,9 @@
   "description": "",
   "scripts": {
     "start": "wunderctl up --debug",
-    "generate": "wunderctl generate up --debug",
+    "generate": "wunderctl generate --debug",
     "check": "tsc --noEmit",
-    "test": "ava"
```
fiam, line 9: Please, leave this example with ava tests. We want to have at least one as an example and also ensures our test library works with testing suites that run tests in parallel.

examples/graphql-apollo-subscriptions/package.json
```diff
@@ -6,6 +6,8 @@
     "start": "run-p start:services start:server",
     "start:server": "wait-on tcp:8085 && wunderctl up --debug",
     "start:services": "cd chat && go run server/server.go",
+    "cleanup": "kill-port 8085",
```
fiam, line 9: I don't think this is a good idea. We'd be replacing 4 lines in an sh script with a line that needs to be maintained for every current and future example as well an extra dependency (which also requires the full lsof on Linux - so it doesn't work in e.g. Alpine)

examples/hooks/.env.example
```diff
@@ -0,0 +1,3 @@
+WUNDERGRAPH_SECURE_COOKIE_HASH_KEY=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
```
fiam, line 1: This is not the right solution, since it requires managing and `.env` file in the repository just for development and adding an extra package script to every example that uses auth. If we think that setting these in `beforeAll()` is too verbose, there are other 2 solutions that we might apply that require no user intervention, either:

- If these variables are empty during the testing library startup, set them before calling `wunderctl start`.
- Alternatively, add a `--dev-mode` flag to `wunderctl start` and use it when starting the server.

StarpTech: > If these variables are empty during the testing library startup, set them before calling wunderctl start

How?

> Alternatively, add a --dev-mode flag to wunderctl start and use it when starting the server.

This will only work for values with a known default value.

StarpTech: > This is not the right solution, since it requires managing and .env file in the repository just for development and adding an extra package script to every example that uses auth

I don't understand this. I rename `.env.example` to `.env` on the `postinstall` hook. At this point, the user can work with the defaults. Having a local `.env` is the best practice.

examples/openapi-mocking/package.json
StarpTech: It doesn't work. I got an error regarding `new` with classes.

examples/graphql-apollo-subscriptions/package.json
StarpTech: I tried to restore your kill procedure, but it didn't work. Feel free to re-add it but for now, I move forward with the cleanup hook.

examples/openapi-mocking/package.json
fiam: It works in my machine and in CI. Please, check your local setup.

examples/graphql-apollo-subscriptions/package.json
fiam: Again, works in my machine and in CI

StarpTech: I reverted it. It was due to pnpm. I had to use npm. PNPM always results in "ELIFECYCLE  Command failed" even on success.

examples/openapi-mocking/package.json
StarpTech: Even if it works now, mixing test frameworks randomly is confusing. If we want to show ava, we should provide a dedicated example. Let's go with jest as a default test runner.

examples/hooks/.env.example
jensneuse: I actually think it's perfect as it allows the user to run `wunderctl start` in the project without needing to figure out how to set the environment variables. It's not just about the tests but also about running the app in production. By adding this env file, they can draw the connection between `new EnvironmentVariable('SECRET')` and the `.env` file.

Alternatively, we could "omit" the secrets altogether if we run `wunderctl start --insecure`, which would override the secrets with a default but static value.

StarpTech: I'd not say alternatively, because as you said it is not just for default values also for configuration in general https://github.com/wundergraph/wundergraph/pull/461#discussion_r1051654954

fiam (approved):
Pending dropping references to ava from `testing.md`

PR #463, opened 2022-12-18
test: add tests for webhooks example

fiam:


fiam:
Changing to draft while we decide how to handle secrets during tests

StarpTech (approved):
LGTM

spetrunin (changes requested):
Tests are failing

I suppose it requires example env file?

fiam:
@spetrunin Thanks for the heads-up, that was indeed the problem. Good catch!

spetrunin (approved):

PR #476, opened 2022-12-21
feat: add upload profiles for S3 providers

fiam:


examples/nextjs-file-upload/.wundergraph/wundergraph.server.ts
```diff
@@ -0,0 +1,49 @@
...
+		uploads: {
+			minio1: {
+				coverPicture: {
+					preUpload: ({ user, file, meta }) => {
+						console.log(`preUpload user: ${user}, file: ${file}, meta: ${meta}`);
+						if (!user) {
+							return { error: 'authenticate' };
+						}
+						return { fileKey: 'myfile.png' };
+						//						return {fileKey: file.name}
```
Pagebakers, line 19: Why so many tabs :D

examples/nextjs-file-upload/.wundergraph/wundergraph.server.ts
```diff
@@ -0,0 +1,49 @@
...
+					postUpload: async ({ user, file, meta, internalClient, error }) => {
+						console.log(
+							`postUpload user: ${user}, file: ${file}, meta: ${meta}, internalClient: ${internalClient}, error: ${error}`
+						);
+					},
+				},
+			},
+		},
+	},
+	graphqlServers: [
```
Pagebakers, line 30: Remove this, we shouldn't be including this in examples any longer.

packages/sdk/src/client/client.ts
```diff
@@ -383,6 +394,14 @@ export class Client {
 			wg_api_hash: this.options.applicationHash,
 		});
 
+		if ('profile' in config) {
+			headers['X-Upload-Profile'] = (config as any).profile ?? ('' as string);
```
Pagebakers, line 398: Do we need to cast config to any here?

`('' as string);` could be just `''` too in this case?

packages/sdk/src/client/client.ts
```diff
@@ -383,6 +394,14 @@ export class Client {
...
+		}
+
+		if ('meta' in config) {
+			headers['X-Metadata'] = (config as any).meta ? JSON.stringify((config as any).meta) : '';
```
Pagebakers, line 402: Idem

packages/sdk/src/client/client.ts
```diff
@@ -410,6 +429,42 @@ export class Client {
 		};
 	}
 
+	public validateFiles(config: UploadRequestOptions, validation?: UploadValidationOptions) {
+		if (validation?.maxAllowedFiles && config.files.length > validation.maxAllowedFiles) {
+			throw new Error(`uploading ${config.files.length} exceeds the maximum allowed (${validation.maxAllowedFiles})`);
```
Pagebakers, line 434: Now we have a new InputValidationError error type, I think having a UploadValidationError makes sense as well. We can add additional information to make it easier to display relevant error messages to users.

Actually now I think about it, this should've been part of the spec.

There's much more to it, eg I would like an error object for all files and their validation errors.

Let's do this in v2, I'll open an issue for it.

Pagebakers (changes requested):
Did you update the docs in another PR?

Something that needs to be documented as well is using paths in your fileKeys to create folders.

eg 
```
{
  fileKey: `${user.id}/${fileHash}.${file.extension}`
}
```

examples/nextjs-file-upload/.wundergraph/wundergraph.server.ts
fiam: Removed

packages/sdk/src/configure/index.ts
```diff
@@ -1,4 +1,18 @@
...
+	buildSchema,
+	FieldDefinitionNode,
+	InputValueDefinitionNode,
+	Kind,
+	parse,
+	parseType,
+	print,
+	visit,
+} from 'graphql';
+import { z } from 'zod';
```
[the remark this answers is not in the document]
fiam: Good idea, changed

examples/nextjs-file-upload/.wundergraph/wundergraph.server.ts
fiam: Commented via keyboard shortcut, it kept all the indentation. Fixed.

packages/sdk/src/client/client.ts
fiam: As discussed, we found no cleaner way to do this

packages/sdk/src/client/client.ts
fiam: Fixed

fiam:
> Did you update the docs in another PR?
> 
> Something that needs to be documented as well is using paths in your fileKeys to create folders.
> 
> eg
> 
> ```
> {
>   fileKey: `${user.id}/${fileHash}.${file.extension}`
> }
> ```

Where should we document this?

fiam:
@jensneuse Merge conflicts fixed, PTAL

Pagebakers:
> > Did you update the docs in another PR?
> > Something that needs to be documented as well is using paths in your fileKeys to create folders.
> > eg
> > ```
> > {
> >   fileKey: `${user.id}/${fileHash}.${file.extension}`
> > }
> > ```
> 
> Where should we document this?

In the uploads documentation I'd say.

More importantly, does it work already? (I believe it should by default, but I can't confirm)

pkg/apihandler/apihandler.go
```diff
@@ -255,13 +255,32 @@ func (r *Builder) BuildAndMountApiHandler(ctx context.Context, router *mux.Route
...
 	for _, s3Provider := range api.S3UploadConfiguration {
+		var profiles map[string]*s3uploadclient.UploadProfile
+		for name, profile := range s3Provider.UploadProfiles {
+			if profiles == nil {
+				profiles = make(map[string]*s3uploadclient.UploadProfile)
+			}
+			profiles[name] = &s3uploadclient.UploadProfile{
+				MaxFileSizeBytes:      int(profile.MaxAllowedUploadSizeBytes),
+				MaxAllowedFiles:       int(profile.MaxAllowedFiles),
+				AllowedMimeTypes:      append([]string(nil), profile.AllowedMimeTypes...),
```
YuriBuerov, line 266: could `profile.AllowedMimeTypes` be nil? (possible runtime error)

pkg/s3uploadclient/s3uploadclient.go
```diff
@@ -52,10 +100,66 @@ func NewS3UploadClient(endpoint string, s3Options Options) (*S3UploadClient, err
...
+		allowedMimeTypesRegexps := make([]*regexp.Regexp, len(profile.AllowedMimeTypes))
+		for ii, mimeType := range profile.AllowedMimeTypes {
+			normalized := strings.ToLower(mimeType)
+			allowedMimeTypes[ii] = normalized
+			if strings.IndexByte(normalized, '*') >= 0 {
+				re, err := regexp.Compile(strings.ReplaceAll(normalized, "*", ".*"))
+				if err != nil {
+					if s3Options.Logger != nil {
+						s3Options.Logger.Warn("error compiling mimetype wildcard", zap.String("mimetype", normalized), zap.Error(err))
+						continue
```
YuriBuerov, line 116: should we continue if `err != nil`? 
```
re, err := regexp.Compile(strings.ReplaceAll(normalized, "*", ".*"))
if err != nil {
    if s3Options.Logger != nil {
	s3Options.Logger.Warn("error compiling mimetype wildcard", zap.String("mimetype", normalized), zap.Error(err))
    }
    continue
}
```

Could you also explain in what cases we want `s3Options.Logger` to be nil, and why?

YuriBuerov: I would also put normalization into the separate helper

YuriBuerov (changes requested):
just a few comments

pkg/apihandler/apihandler.go
```diff
@@ -255,13 +255,32 @@ func (r *Builder) BuildAndMountApiHandler(ctx context.Context, router *mux.Route
 	}
 
 	for _, s3Provider := range api.S3UploadConfiguration {
+		var profiles map[string]*s3uploadclient.UploadProfile
+		for name, profile := range s3Provider.UploadProfiles {
+			if profiles == nil {
```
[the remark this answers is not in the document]
fiam: I'll change it to preallocate the map with the required size. If there are no profiles we waste 48 bytes, but if there are profiles it will end up using less memory and being faster.

pkg/s3uploadclient/s3uploadclient.go
fiam: I wouldn't expect that error to ever happen in a real scenario, but I'd rather log it just in case I got it wrong. Notice that the code handles having nils in allowedMimeTypesRegexps because not all mimetypes get compiled as a regular expression.

Pagebakers (approved):
LGTM, let's get this in the hands of developers.

Can you open an issue to allow for anonymous uploads? We can use the same api as functions `authenticationRequired: true/false` in the uploadProfile

fiam:
> LGTM, let's get this in the hands of developers.
> 
> Can you open an issue to allow for anonymous uploads? We can use the same api as functions `authenticationRequired: true/false` in the uploadProfile

Thanks! I’ll open an issue so we can get the ball rolling.

jensneuse (approved):

fiam:
@YuriBuerov can you please take another look? Thanks!

YuriBuerov (approved):

PR #477, opened 2022-12-21
feat: add @removeNullVariables directives

### LINKED ISSUE #114: null values in body are not passed through to graphql backend
The Wundergraph operation looks like this:
```
mutation someUpdate($id: Int!, $name: String, $description: String, $icon: URL, $image: URL, $labels: [String!]) {
    object: someUpdate(id: $id, name: $name, description: $description, icon: $icon, image: $image, labels: $labels)
    {
        id
    }
}
```
I have the semantics that we can clear a value by sending an update with a null value. 
This is different to sending an update when some values are not set, which means don't change the values not included in the call.

For example, in the below call, I want to clear the existing "name" attribute on the object.
```
--- ClientRequest start ---

POST /api/main/operations/someUpdate HTTP/1.1
Host: localhost:9991
Accept: */*
Content-Length: 23
Content-Type: application/json
User-Agent: curl/7.68.0

{"id": 0, "name": null}



--- ClientRequest end ---



--- DebugTransport ---

Request:

POST / HTTP/1.1
Host: localhost:8000
Accept: application/json
Content-Type: application/json

{"variables":{"id":0},"query":"mutation($id: Int!, $name: String, $description: String, $icon: URL, $image: URL, $labels: [String!]){object: someUpdate(id: $id, name: $name, description: $description, icon: $icon, image: $image, labels: $labels){id}}"}

Duration: 121 ms

Response:

HTTP/1.1 200 OK
Content-Length: 28
Content-Type: application/json
Date: Fri, 05 Aug 2022 04:52:27 GMT
Server: uvicorn

{"data":{"object":{"id":0}}}

--- DebugTransport
```

You can see by the above, the Wundergraph server removes the "name" attribute, as it is null. If I set it to something else, it works ok.

The desired behaviour is for the wundergraph server to pass through null-valued arguments.

spetrunin:
- Add @removeNullVariables directive
- Make input coersion properly work with variables count  >2

closes #114 

#### Checklist

- [ ] run `make test`


StarpTech (approved):
LGTM

Pagebakers (approved):
LGTM

PR #493, opened 2022-12-26
fix(node): listening on tcp4 for 0.0.0.0

StarpTech:
Go's behavior is not strict. Listening `0.0.0.0` (ipv4) on a `tcp` socket has produced network issues on a platform like fly.io. Go defaults to dual stack. As a fix, we will create only a single `tcp4` socket in that case. This PR also removes the discovery of multiple ipv4 and ipv6 addresses from localhost. 

Here are the referenced issues:

- https://github.com/golang/go/issues/17615
- https://github.com/golang/go/issues/9334

#### Checklist

- [ ] run `make test`


pkg/node/node.go
```diff
@@ -264,65 +264,28 @@ func (n *Node) Close() error {
...
-				return nil, err
-			}
-			listeners = append(listeners, &proxyproto.Listener{
-				Listener: listener,
-			})
-		}
+	bindProto := "tcp"
+	// If they've passed 0.0.0.0, we only want to bind on IPv4
+	// rather than golang's dual stack default
+	if host == "0.0.0.0" {
```
jensneuse, line 277: I would do this instead as I think it's pretty accurate and will work for all kinds of local IP addresses: https://stackoverflow.com/a/48519490/20693155

```go
func IsIPv4(address string) bool {
    return strings.Count(address, ":") < 2
}
```

jensneuse: So, we only listen ipv4 when the user explicitly wants to listen to an ipv6 address. WDYT?

jensneuse (changes requested):

Makefile
```diff
@@ -59,7 +59,8 @@ wunderctl: build
 run:
 	cd cmd/wunderctl && go run main.go
 
+# CGO_ENABLED is set to 0 to avoid linking to libc. This is required to run the binary on alpine.
 install:
-	cd cmd/wunderctl && go install
+	cd cmd/wunderctl && CGO_ENABLED=0 go install
```
fiam, line 64: Please, don’t force this here. We can set it during the build as needed. Keep in mind that CGO_ENABLED=0 is not an Alpine requirement, it’s instead a requirement when you compile in a system with glibc with the intention to run the binary on a musl based system.

pkg/node/node.go
StarpTech: Great, I think this will work.

Makefile
StarpTech: My intention here is that I don't need to set this flag manually when producing a test binary for the builder. Since we also have no cgo dependencies, we can set it without any issues.

> Keep in mind that CGO_ENABLED=0 is not an Alpine requirement, it’s instead a requirement when you compile in a system with glibc with the intention to run the binary on a musl based system.

Yes, I mean, in the case of alpine it is required because the binary was linked differently. In order to produce a cross-compatible image we have to set CGO_ENABLED=0 therefore disables native builds. That's everything I want to do here.

pkg/node/node.go
StarpTech: Done.

jensneuse (approved):

fiam (approved):

PR #494, opened 2022-12-27
chore: update react-query example

Pagebakers:
Updated the example to @wundergraph/react-query@0.2.0
Removed secrets configuration.


examples/nextjs-react-query/.wundergraph/wundergraph.config.ts
```diff
@@ -37,9 +29,6 @@ configureWunderGraphApplication({
 		cookieBased: {
 			providers: [authProviders.demo()],
 			authorizedRedirectUris: ['http://localhost:3000'],
-			secureCookieHashKey: new EnvironmentVariable('WUNDERGRAPH_SECURE_COOKIE_HASH_KEY'), // must be of length 32
```
JivusAyrus, line 40: We need this for the tests

JivusAyrus (changes requested):
Can we change the name in the package.json?  I forgot to do it last time

examples/nextjs-react-query/.wundergraph/wundergraph.config.ts
Pagebakers: This is included in all examples now?

JivusAyrus: All examples which use authentication

JivusAyrus (approved):

examples/nextjs-react-query/.wundergraph/wundergraph.config.ts
Pagebakers: Ok I thought it was some leftover code from copying another example. This one doesn't use auth. But I'll leave it in for now then.

fiam (approved):

PR #497, opened 2022-12-27
feat(node)!: restructure server imports

StarpTech:
This PR will restructure the import paths for the server component. In that way, the server does not require more modules and paths as needed. This has improved the cold start by multiple seconds. 

### TODO:

- [X] Updates testapps
- [ ] Updates examples
- [ ] Update docs

#### Checklist

- [ ] run `make test`


jensneuse (approved):

Pagebakers (changes requested):
Great improvement! Only thing I see it the webhook example isn't updated (only the testapp)

Actually we need to update all examples and docs :)

StarpTech:
> Great improvement! Only thing I see it the webhook example isn't updated (only the testapp)
Actually we need to update all examples and docs :)

Thanks! Yes, I'm going to update all testapps in this PR, but examples after the release.

Pagebakers (approved):
LGTM

jensneuse (approved):

PR #293, opened 2022-10-26
fix: operation metadata in react provider

rwest202:
Adds operation metadata to react provider's client instantiation.

#### Checklist

- [x] run `make test`
- [x] PR title must follow [conventional-commit-standard](https://github.com/wundergraph/wundergraph/blob/main/CONTRIBUTING.md#conventional-commit-standard)
- [x] tests and/or benchmarks are included
- [x] documentation is changed or added
- [x] commit message and code follows the [Code of conduct](https://github.com/wundergraph/wundergraph/blob/main/CODE_OF_CONDUCT.md)

PR #501, opened 2022-12-29
fix: operation metadata type imports

rwest202:
#### Checklist

- [x] run `make test`

Follow up to https://github.com/wundergraph/wundergraph/pull/293. Fix was broken, creates a compiler error in generated code.

```
Type error: Cannot find name 'OperationMetadata'.

  28 | }
  29 | 
> 30 | const operationMetadata: OperationMetadata = {
     |                          ^
  31 |   CreateTag: {
  32 |     requiresAuthentication: false,
  33 |   },
error Command failed with exit code 1.
```

This PR adds the OperationMetadata import to the top of the `react.provider.template.ts` and `react.native.provider.template.ts`


Pagebakers (approved):

Pagebakers:
Looks good, but are you still using these ReactProviders? They are obsolete now with our new SWR or React Query packages.

jensneuse:
@Pagebakers can it be merged?

rwest202:
@Pagebakers would be great to merge, I currently run a script to add the line to the generated file before starting or building nextjs.

I tried the SWR/React Query packages but I just end up creating useEffect madness in my components, is there an example repo with more complex logic? I find that having a promise that returns a single value is much easier to manage then exporting `{ data, error, mutate, mutateAsync }` for every mutation, especially if I have multiple mutations in a single component and have to rename all of those exports.

Pagebakers:
It should actually make your life much easier and prevent useEffect madness. Trigger (or mutate with React Query) returns a promise, you can act on that, besides that useMutation also has callbacks you can use to invalidate data, etc.

```ts
const mutation = useMutation({
  operationName: 'SetName',
  onSucces: () => // do something after the mutation was succesful
})

await mutation.trigger({name: 'Ryan'})
```

jensneuse (approved):

Pagebakers:
@rwest202 Can you update your branch with main? Then we get merge it in.

PR #512, opened 2023-01-03
fix: error messages in OpenIDConnectCookieHandler

fiam:
Log messages incorrectly mentioned GitHub, leading to difficulties
when troubleshooting


Pagebakers (approved):
Nice one

spetrunin (approved):

PR #515, opened 2023-01-04
chore: wip implement functions

jensneuse:


jensneuse:
Tests are failing because we first need to make a binary release.

packages/sdk/src/operations/operations.ts
```diff
@@ -0,0 +1,183 @@
+import { z } from 'zod';
+import * as fs from 'fs';
+import type { BaseRequestContext, InternalClient } from '../server';
+import type { User } from '../client';
```
StarpTech, line 4: Are you sure this is the client user? I'd assume since this the server you should use `WunderGraphUser`

```ts
BaseRequestContext<User, IC> & AuthenticationRequestContext<User>;
```

packages/sdk/src/operations/operations.ts
```diff
@@ -0,0 +1,183 @@
+import { z } from 'zod';
```
StarpTech, line 1: Please put everything to `/src/server`. This is the export for all server components.

packages/sdk/src/server/plugins/functions.ts
```diff
@@ -0,0 +1,109 @@
...
+						};
+
+						switch (implementation.type) {
+							case 'subscription':
+								if (!implementation.subscriptionHandler) {
+									return reply.status(500);
+								}
+								const subscribeOnce = request.headers['x-wg-subscribe-once'] === 'true';
+								const gen = await implementation.subscriptionHandler(ctx);
+								reply.hijack();
```
StarpTech, line 50: I'd call this before `subscriptionHandler` so the code in the catch handler can hijack also when `subscriptionHandler` throws.

packages/sdk/src/server/plugins/functions.ts
```diff
@@ -0,0 +1,109 @@
...
+								reply.raw.on('close', () => {
+									gen.return(0);
+								});
+								while (true) {
```
StarpTech, line 54: This is blocking the event loop all the time. For/Loop iterates until all values have been consumed. E.g.

```ts
let generator = generateSequence();

for await (let value of generator) {
     
}
```

StarpTech (changes requested):

fly-machine-test.sh
```diff
@@ -0,0 +1,15 @@
...
+		exit 1
+fi
+
+fly apps create --machines --name "$FLY_APP_NAME"
+fly ips allocate-v4 -a "$FLY_APP_NAME"
+fly m run --dockerfile Dockerfile-Cloud-Starter . -p 443:9991/tcp:tls:http --region fra -a "$FLY_APP_NAME"
+
+# wait for user to press enter
+read -p "Press enter to delete the app after testing"
+fly apps destroy "$FLY_APP_NAME" --yes
```
Pagebakers, line 15: Same as the Dockerfile-Cloud-Starter, don't think this should be here

Dockerfile-Cloud-Starter
```diff
@@ -0,0 +1,34 @@
...
+COPY testapps/cloud-starter/package.json testapps/cloud-starter/package-lock.json /app/
+RUN npm ci --cache /root/cache/npm --prefer-offline --no-audit
+COPY testapps/cloud-starter /app
+COPY --from=build /go/bin/wunderctl /usr/local/bin/wunderctl
+WORKDIR /app
+RUN wunderctl generate --pretty-logging=true --wundergraph-dir=.wundergraph
+
+ENV NODE_ENV=production CI=false START_HOOKS_SERVER=true WG_ABS_DIR=/app/.wundergraph WG_NODE_URL=http://localhost:9991 WG_NODE_HOST=0.0.0.0 WG_NODE_PORT=9991 WG_SERVER_URL=http://localhost:9992 WG_SERVER_HOST=127.0.0.1 WG_SERVER_PORT=9992
+
+CMD ["/bin/sh","-c","node .wundergraph/generated/bundle/server.js & wunderctl start --exclude-server --pretty-logging=false --shutdown-after-idle=10 --wundergraph-dir=.wundergraph"]
```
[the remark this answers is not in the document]
jensneuse: this is only meant for testing fly machines, just ignore it

Dockerfile-Cloud-Starter
```diff
@@ -0,0 +1,34 @@
+FROM golang:1.18 as build
```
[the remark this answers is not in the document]
jensneuse: this is only meant for testing fly machines, just ignore it

packages/sdk/src/operations/operations.ts
jensneuse: Can you explain? Does this belong to server? It's about operations/functions?

StarpTech: To separate the server from the sdk we have introduced `@wundergraph/sdk/server` as the primary entry point for everything server related. Functions similar to webhooks are running on the server so I'd suggest to put it there as well.

packages/sdk/src/server/plugins/functions.ts
```diff
@@ -0,0 +1,109 @@
...
+								reply.send({
+									response: {
+										data: outMutation,
+									},
+								});
+								return;
+						}
+
+						return;
+					} catch (e) {
```
Pagebakers, line 92: Does this implement input validation errors?

```
{
	propertyPath: string;
	message: string;
	invalidValue: any;
}[]
```

Dockerfile-Cloud-Starter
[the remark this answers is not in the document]
StarpTech: ok, but I'd not put it at the root because it might confuse users.

Pagebakers (changes requested):
With folder based operations, what does this mean to .graphql operations?

/users/get.graphql translate to `operationName: 'users/get'` now as well?

In that case it's a breaking change, because we have some customers/users that group operations in sub folders. (Tripadvisor?)

Add migration instructions, and inform developers.

Dockerfile-Cloud-Starter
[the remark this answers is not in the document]
jensneuse: moved into test app

packages/sdk/src/operations/operations.ts
jensneuse: changed

StarpTech (approved):
LGTM

packages/sdk/src/server/plugins/functions.ts
jensneuse: it's handled in go by the gateway already

fly-machine-test.sh
jensneuse: moved

jensneuse:
> With folder based operations, what does this mean to .graphql operations?
> 
> /users/get.graphql translate to `operationName: 'users/get'` now as well?
> 
> In that case it's a breaking change, because we have some customers/users that group operations in sub folders. (Tripadvisor?)
> 
> Add migration instructions, and inform developers.

yup

PR #517, opened 2023-01-06
fix: listen on IPv6 when WG_NODE_HOST is a hostname

fiam:
- fix: when WG_NODE_HOST is a hostname resolve and listen in all addresses
- chore: dummy commit to trigger examples CI, UNDO BEFORE MERGE


examples/simple/README.md
```diff
@@ -33,3 +33,5 @@ Read the [Docs](https://wundergraph.com/docs).
 ## Got Questions?
 
 Join us on [Discord](https://wundergraph.com/discord)!
+
+dummy
```
StarpTech, line 37: Revert

pkg/node/node.go
```diff
@@ -264,35 +264,59 @@ func (n *Node) Close() error {
...
-	} else if IsIPv6(host) {
-		bindProto = "tcp6"
-		address = fmt.Sprintf("[%s]:%d", host, port)
-	}
-
-	listener, err := cfg.Listen(context.Background(), bindProto, address)
-	if err != nil {
-		return nil, err
+	var addrs []string
+	if net.ParseIP(host) == nil {
```
StarpTech, line 275: localhost is special. We should not inherit it for all names. This can be a pitfall. We only now that localhost is safe.

StarpTech (changes requested):

pkg/node/node.go
```diff
@@ -264,35 +264,59 @@ func (n *Node) Close() error {
...
+				continue
+			}
+			bindProto = "tcp6"
+			address = fmt.Sprintf("[%s]", addr)
+		} else {
+			panic(fmt.Errorf("%s doesn't look like an IPv4 nor an IPv6", addr))
+		}
+
+		toListen := fmt.Sprintf("%s:%d", address, port)
+		listener, err := cfg.Listen(context.Background(), bindProto, toListen)
```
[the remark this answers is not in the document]
fiam: Changed

pkg/node/node.go
```diff
@@ -264,35 +264,59 @@ func (n *Node) Close() error {
...
+			address = addr
+		} else if IsIPv6(addr) {
+			// Filter out link-local addresses
+			if strings.HasPrefix(addr, "fe80:") {
+				continue
+			}
+			bindProto = "tcp6"
+			address = fmt.Sprintf("[%s]", addr)
+		} else {
+			panic(fmt.Errorf("%s doesn't look like an IPv4 nor an IPv6", addr))
```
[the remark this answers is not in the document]
fiam: This is one of the cases where one should panic, because it's a programming error. The panic is to mark the section as unreachable (e.g. like `__builtin_unreachable()` GCC's builtin). However, I've just realized we can rewrite that to not need 3 branches and then we can remove the panic entirely.

examples/simple/README.md
fiam: I'll leave it in until we get the PR ready to be merged, so CI continues to trigger

pkg/node/node.go
fiam: There are more ways to specify a valid hostname that might make sense. `something.local` or even a public DNS name. I've also realized we should not manually decide when something is an IP or it's not. We can always resolve it and look at the returned values.

examples/simple/README.md
StarpTech: ok

pkg/node/node.go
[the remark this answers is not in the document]
StarpTech: Great

StarpTech:
I think I found another issue related to Ipv6. When the node only listen to ipv6 but the server on both I get 

```
2023-01-10T11:32:24+01:00 INFO: Server listening at http://[::1]:9992 {"component":"@wundergraph/server"}
2023-01-10T11:32:24+01:00 INFO: Server listening at http://127.0.0.1:9992 {"component":"@wundergraph/server"}


2023-01-10T11:29:01+01:00 ERROR: connect ECONNREFUSED ::1:9991 {"component":"@wundergraph/server","reqId":"2913d5a2-1a18-108e-df27-c716eb4c560c"}
    err: {
      "message": "connect ECONNREFUSED ::1:9991",
      "stack":
          Error: connect ECONNREFUSED ::1:9991

```

Are you able to reproduce?

StarpTech (approved):
LGTM

pkg/node/node.go
```diff
@@ -498,30 +522,38 @@ func (n *Node) startServer(nodeConfig WunderNodeConfig) error {
...
-			n.log.Debug("listener closed",
-				zap.String("addr", listener.Addr().String()),
+	for _, listener := range listeners {
+		l := listener
+		g.Go(func() error {
+			n.log.Info("listening on",
+				zap.String("addr", l.Addr().String()),
 			)
+
+			if err := n.server.Serve(l); err != nil {
```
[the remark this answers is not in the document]
fiam: If you do it that way then you can't use the shorter "if with assignment"

Aenimus (approved):

PR #519, opened 2023-01-07
fix: incorrectly derived filesystem paths on Windows

fiam:
Use filepath.Join() instead of path.Join() to construct filesystem paths.

### DIFF
--- a/cli/commands/generate.go
+++ b/cli/commands/generate.go
@@ -3,7 +3,7 @@ package commands
 import (
 	"context"
 	"fmt"
-	"path"
+	"path/filepath"
 
 	"github.com/spf13/cobra"
 	"go.uber.org/zap"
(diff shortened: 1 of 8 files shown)


fiam:
This needs to be tested on Windows before merging

fiam:
Windows binary attached, in case anyone has a chance to test it before I do
[wunderctl.exe.zip](https://github.com/wundergraph/wundergraph/files/10367168/wunderctl.exe.zip)

StarpTech (approved):
Good catch. I think we should also revisit other places.

fiam:
@johnrazmus Thanks for the report. There were several more occurrences of incorrectly used `path.Join()` calls. I've gone through all the occurrences and changed the ones referring to the filesystem to use `filepath.Join()` instead.

Here's an updated Windows binary
[wunderctl.exe.zip](https://github.com/wundergraph/wundergraph/files/10369499/wunderctl.exe.zip)

johnrazmus:
Pathings seem good now, ty!

fiam:
@johnrazmus Thanks for your help! The fix should be in the next release.

PR #525, opened 2023-01-09
feat: add support for disconnecting from Auth0

fiam:
Auth0 doesn't implement OIDC's end_session_endpoint, requiring a custom
API instead. Add machinery to support this kind of APIs and implement
a provider type to handle this.

Also, update the auth0-oidc-authentication example to use port 3000
and add some extra setup instructions

Fixes ENG-1016
Fixes ENG-968


jensneuse (approved):

jensneuse (changes requested):
see comments

pkg/authentication/oidc.go
```diff
@@ -313,3 +323,208 @@ func (h *OpenIDConnectCookieHandler) createProvider(ctx context.Context, config
...
+
+func (p *OpenIDConnectProvider) disconnectAuth0(ctx context.Context, user *User) (*OpenIDDisconnectResult, error) {
+	return &OpenIDDisconnectResult{
+		Redirect: fmt.Sprintf("%sv2/logout?client_id=%s", p.config.Issuer, p.clientID),
+	}, nil
+}
+
+type OpenIDConnectProviderSet struct {
+	mu sync.Mutex
+	m  map[string]*OpenIDConnectProvider
```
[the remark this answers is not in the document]
fiam: You're not missing anything, it's not required. I added it before I went through the rest of the code handling OIDC providers. I'll remove it.

examples/auth0-oidc-authentication/.wundergraph/wundergraph.config.ts
```diff
@@ -31,20 +31,20 @@ configureWunderGraphApplication({
...
 		...cors.allowAll,
-		allowedOrigins: ['http://localhost:3003'],
+		allowedOrigins: ['http://localhost:3000'],
 	},
 	authentication: {
 		cookieBased: {
 			providers: [
 				authProviders.demo(),
-				authProviders.openIdConnect({
+				authProviders.auth0({
```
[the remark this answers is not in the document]
fiam: That's correct, that's why we made `create-wundergraph-app` download from the last stable tag instead of fetching from main (otherwise we can't ever make breaking changes). But you've made a good point, because we totally forgot to do the same for the cloud templates.

packages/sdk/src/client/types.ts
```diff
@@ -147,5 +147,15 @@ export interface User<Role extends string = string> {
...
+	 * Wether to log out the user from the OpenID Connect provider.
+	 * Some providers might require the user to visit a URL. See
+	 * the redirect field.
+	 */
 	logoutOpenidConnectProvider?: boolean;
+	/**
+	 * Custom function for redirecting the client to the log out
+	 * URL. If not provided, window.location.href is updated.
+	 */
+	redirect?: (url: string) => Promise<boolean>;
```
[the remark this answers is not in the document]
fiam: Good idea. I’ll add it.

fiam:
@jensneuse PR updated, PTAL. Thanks!

pkg/authentication/oidc.go
```diff
@@ -313,3 +322,201 @@ func (h *OpenIDConnectCookieHandler) createProvider(ctx context.Context, config
...
+		return fmt.Errorf("duplicate OIDC provider ID %q", id)
+
+	}
+	s.m[id] = p
+	return nil
+}
+
+func (s *OpenIDConnectProviderSet) ByIssuer(issuer string) *OpenIDConnectProvider {
+	for _, provider := range s.m {
+		if provider.config.Issuer == issuer {
```
jensneuse, line 516: If we have 2 OIDC providers with a different id, but the same issue, we'd select the first, which might be wrong.
How about using the user object and selecting the provider by id? This way, you could also use the map actually.

jensneuse (changes requested):

pkg/authentication/oidc.go
fiam: I like that better. I though we were using the issuer URL because there might be circumstances where `User.ProviderID` is not set, but I guess I was wrong. I'll update it.

fiam: Updated, PTAL

pkg/authentication/oidc.go
```diff
@@ -511,12 +511,10 @@ func (s *OpenIDConnectProviderSet) Add(id string, p *OpenIDConnectProvider) erro
 	return nil
 }
 
-func (s *OpenIDConnectProviderSet) ByIssuer(issuer string) *OpenIDConnectProvider {
-	for _, provider := range s.m {
-		if provider.config.Issuer == issuer {
-			return provider
-		}
+func (s *OpenIDConnectProviderSet) ByID(id string) (*OpenIDConnectProvider, error) {
+	provider := s.m[id]
```
jensneuse, line 515: you can do `provider,exists := s.m[id]`
btw, can m have a real name, like providers? =)

jensneuse (approved):

jensneuse (approved):

jensneuse:
good work!

fiam:
> good work!

Thanks!

packages/sdk/src/client/client.ts
```diff
@@ -411,6 +414,26 @@ export class Client {
 			method: 'GET',
 		});
 
-		return response.ok;
+		if (!response.ok) {
+			return false;
+		}
+
+		let ok = true;
+		if (response.headers.get('Content-Type') === 'application/json') {
```
StarpTech, line 422: From my experience with https://github.com/StarpTech/apollo-datasource-http I know that content negotiation like this is error prone.

I recommend doing it like this https://github.com/StarpTech/apollo-datasource-http/blob/main/src/http-data-source.ts#L344

packages/sdk/src/client/client.ts
```diff
@@ -411,6 +414,26 @@ export class Client {
...
+			const data = (await response.json()) as LogoutResponse;
+			if (data.redirect) {
+				if (options?.redirect) {
+					ok = await options.redirect(data.redirect);
+				} else {
```
StarpTech, line 427: This only works on the browser. Is it intended to run this function on the server? At least, I'd check for the existence of the window object.

StarpTech (changes requested):

packages/sdk/src/client/client.ts
fiam: It's the same approach we use with `fetch`, we let it blow up in the server. I will add a check, but what should we do when not in a browser? Throw an error asking the user to pass the redirect function?

packages/sdk/src/client/client.ts
fiam: That's definitely more robust, I'll change it.

packages/sdk/src/client/client.ts
StarpTech: I'd do nothing. The user can already do something after logout with `after`. The redirect is async anyway.

fiam: The redirect is required for disconnecting from the auth provider and completing the log out when `logoutOpenidConnectProvider` is true, we need to keep it.

fiam: I've noticed we don't support `login()` and `logout()` in the server side. `login()` already has a check for it, I'll add the same in `logout()`

StarpTech: That's sad because you could issue a logout on the server and redirect yourself.

StarpTech (approved):
LGTM

PR #529, opened 2023-01-11
fix: prevent swallowing of input validation errors

Aenimus:
#### Checklist

- [ ] run `make test`


testapps/default/.wundergraph/wundergraph.config.ts
```diff
@@ -108,12 +108,12 @@ configureWunderGraphApplication({
 						// change this before deploying to production to the actual domain where you're deploying your app
 						'http://localhost:3000',
 				  ]
-				: ['http://localhost:3000'],
+				: ['http://localhost:3000', 'http://localhost:9991'],
```
Aenimus, line 111: I will revert this.

fiam (approved):
LGTM. Excited to get better error messages!

Pagebakers (changes requested):
Looks great!

One thing though. In our SWR and React Query clients we use GraphqlResponseError as the error type. I believe we should rewrite this to ClientResponseError instead. So useQuery and useMutation return the correct error type.

packages/react-query/src/types.ts
```diff
@@ -14,6 +13,7 @@ import {
 	UseQueryResult,
 	UseMutationResult,
 } from '@tanstack/react-query';
+import { ClientResponseError } from '@wundergraph/sdk/dist/client/ClientResponseError';
```
spetrunin, line 16: I believe import should not be from dist, but from package itself

```
@wundergraph/sdk/client/ClientResponseError'
```

maybe you will need to export it ether from client to been able to to this
```
import { ClientResponseError } from '@wundergraph/sdk/dist/client';
```

Pagebakers (approved):
Nice work!

PR #534, opened 2023-01-11
feat: allow to init wundergraph into existing projects

JivusAyrus:
#### Checklist

- [ ] run `make test`


fiam:
Looks good to me, but I’m not sure about using the simple example for that (there might be links to it that we need to update). Maybe we could add a dedicated one for this?

JivusAyrus:
We can add a new one also. But thought it wouldn't be any different from simple.

fiam:
> Looks good to me, but I’m not sure about using the simple example for that (there might be links to it that we need to update). Maybe we could add a dedicated one for this?

Replying to myself: the simple example is still working in this PR, but some deprecated parts are removed. All good!

fiam (approved):

Pagebakers (approved):
Love it

PR #541, opened 2023-01-13
chore: add support for extra telemetry fields

fiam:


Slickstef11 (approved):

cli/commands/root.go
```diff
@@ -209,6 +229,94 @@ func wunderctlBinaryPath() string {
...
+					}
+				}
+				if dataSourceName != "" {
+					metric := telemetry.NewDataSourceMetric(dataSourceName)
+					if urlVariable != nil {
+						urlValue := loadvariable.String(urlVariable)
+						if urlValue != "" {
+							urlHash, err := telemetry.Hash(urlValue)
+							if err != nil {
+								return nil, err
```
jensneuse, line 283: As a rule of thumb, I always try to have at max 2 levels of (if|else|switch) nesting. In very very rare cases 3 levels might be an escape hatch. 8 levels are absolutely unmaintainable. Please give this code some structure.

pkg/telemetry/client.go
```diff
@@ -88,13 +101,67 @@ func NewClient(address string, clientInfo MetricClientInfo, opts ...ClientOption
...
+
 type Metric struct {
-	Name  string  `json:"name"`
-	Value float64 `json:"value"`
+	Name  string      `json:"name"`
+	Value float64     `json:"value"`
+	Tags  []MetricTag `json:"tags,omitempty"`
+}
+
+// Equal returns true iff both Metric instances contain the same data
```
jensneuse, line 119: typo: iff -> if

cli/commands/root.go
```diff
@@ -209,6 +229,94 @@ func wunderctlBinaryPath() string {
...
+				case wgpb.DataSourceKind_POSTGRESQL,
+					wgpb.DataSourceKind_MYSQL,
+					wgpb.DataSourceKind_SQLSERVER,
+					wgpb.DataSourceKind_MONGODB,
+					wgpb.DataSourceKind_SQLITE:
+
+					dataSourceName = ds.Kind.String()
+					urlVariable = ds.CustomDatabase.GetDatabaseURL()
+				default:
+					if err != nil {
```
jensneuse, line 272: Do you want to return an error when the datasource is not supported? I'm not sure if the err check makes sense in this case.

jensneuse (changes requested):

pkg/telemetry/client.go
```diff
@@ -88,13 +101,67 @@ func NewClient(address string, clientInfo MetricClientInfo, opts ...ClientOption
...
+func (m *Metric) Equal(other *Metric) bool {
```
StarpTech, line 120: Please add tests for both functions.

pkg/telemetry/client.go
```diff
@@ -88,13 +101,67 @@ func NewClient(address string, clientInfo MetricClientInfo, opts ...ClientOption
...
+		return true
+	}
+	return false
+}
+
+// AddTag adds a tag to the Metric, validating its name and
+// ensuring there are no duplicates.
+func (m *Metric) AddTag(name string, value string) error {
+	const (
+		invalidChars       = ";!^="
```
StarpTech, line 143: This should be validated on the cloud. In the cloud, we can sanitize invalid tags. In wunderctl, we are in control of every metric. I see no value in adding it here.

cli/commands/root.go
```diff
@@ -209,6 +229,94 @@ func wunderctlBinaryPath() string {
 	return path
 }
 
+func dataSourcesTelemetryMetrics(info *telemetry.MetricClientInfo) ([]*telemetry.Metric, error) {
```
StarpTech, line 232: Please put this in a package. The root file contains already a lot of stuff.

cli/helpers/telemetry.go
```diff
@@ -0,0 +1,50 @@
+package helpers
+
+import "strings"
+
+type TelemetryAnnotation int
+
+const (
+	TelemetryAnnotationCommand TelemetryAnnotation = 1 << iota
```
StarpTech, line 8: What should that mean? Can we be more explicit e.g. `EnableTelemetry` I'd also suggest using string instead int which is easier to grasp.

StarpTech (changes requested):

cli/commands/root.go
fiam: The idea was to make it evident for future developers, but an error might be too harsh. Would you prefer a warning message in the logs while handling the known data source types?

pkg/telemetry/client.go
fiam: https://en.wikipedia.org/wiki/If_and_only_if

cli/commands/root.go
```diff
@@ -209,6 +229,94 @@ func wunderctlBinaryPath() string {
...
+					dataSourceName = ds.Kind.String()
+					urlVariable = ds.GetCustomGraphql().GetFetch().GetUrl()
+				case wgpb.DataSourceKind_POSTGRESQL,
+					wgpb.DataSourceKind_MYSQL,
+					wgpb.DataSourceKind_SQLSERVER,
+					wgpb.DataSourceKind_MONGODB,
+					wgpb.DataSourceKind_SQLITE:
+
+					dataSourceName = ds.Kind.String()
+					urlVariable = ds.CustomDatabase.GetDatabaseURL()
```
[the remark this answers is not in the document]
fiam: It's resolved here https://github.com/wundergraph/wundergraph/pull/541/files#diff-6cef6e709aaccb651a756f23bbae52a22cbe8f82fbe90258831d7c189b42915aR279 - Using that same function

cli/helpers/telemetry.go
fiam: It means that the command usage/duration should be reported via telemetry. `cmd.Annotations` is a map[string]string, but I introduced constants on top to avoid errors due to mistyped strings.

pkg/telemetry/client.go
fiam: I would validate it in both places. Otherwise, the failure report is too disconnected from the error and becomes difficult to diagnose when someone runs into it.

cli/commands/root.go
fiam: Moved it to `pkg/telemetry`, PTAL

cli/commands/root.go
```diff
@@ -209,6 +229,94 @@ func wunderctlBinaryPath() string {
...
+								return nil, err
+							}
+							if u, err := url.Parse(urlValue); err == nil {
+								// Try to parse as host:port, fallback to host
+								host, _, _ := net.SplitHostPort(u.Host)
+								if host == "" {
+									host = u.Host
+								}
+								if host == "localhost" {
+									if err := metric.AddTag("localhost", "true"); err != nil {
```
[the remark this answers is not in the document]
fiam: Updated to detect all private IPs, PTAL

pkg/telemetry/client.go
fiam: Added

cli/helpers/telemetry.go
fiam: Refactored this into a simpler approach

pkg/telemetry/client.go
```diff
@@ -72,6 +75,16 @@ func NewClient(address string, clientInfo MetricClientInfo, opts ...ClientOption
 		opt(c)
 	}
 
+	if clientInfo.GitRepoURLHash == "" {
+		gitRepoURLHash, err := gitRepoURLHash()
+		if err != nil {
+			if c.log != nil && c.debug {
+				c.log.Debug("error retrieving gitRepoURLHash: %w", zap.Error(err))
```
YuriBuerov, line 82: why is it needed to use wrap (`%w`) for the log message?

pkg/telemetry/client.go
```diff
@@ -88,13 +101,67 @@ func NewClient(address string, clientInfo MetricClientInfo, opts ...ClientOption
...
-	Name  string  `json:"name"`
-	Value float64 `json:"value"`
+	Name  string      `json:"name"`
+	Value float64     `json:"value"`
+	Tags  []MetricTag `json:"tags,omitempty"`
+}
+
+// Equal returns true iff both Metric instances contain the same data
+func (m *Metric) Equal(other *Metric) bool {
+	if m.Name == other.Name && m.Value == other.Value && len(m.Tags) == len(other.Tags) {
```
YuriBuerov, line 121: I would simplify it.
Two options:

1) using reflect, something like:
```
func (m *Metric) Equal(other *Metric) bool {
    if m.Name == other.Name && m.Value == other.Value {
        return reflect.DeepEqual(m.Tags, other.Tags)
    }
    return false
}
```

2)  with comparison func:
```
func (t MetricTag) Equal(other MetricTag) bool {
    return t.Name == other.Name && t.Value == other.Value
}

func (m *Metric) Equal(other *Metric) bool {
    if m.Name == other.Name && m.Value == other.Value && len(m.Tags) == len(other.Tags) {
        for i := range m.Tags {
            if !m.Tags[i].Equal(other.Tags[i]) {
                return false
            }
        }
        return true
    }
    return false
}
```

pkg/telemetry/client.go
fiam: Good catch. It was an error return at first. I’ll amend it.

pkg/telemetry/client.go
fiam: I'd rather not use reflect unless we strictly need it, I'll refactor the MetricTag comparison to a method.

fiam: I've just realized that reflect.DeepEqual wouldn't work because multiple Metric instances with the same tags but in different order should still be equal.

pkg/telemetry/client.go
fiam: Updated, PTAL

pkg/telemetry/client.go
fiam: Updated, PTAL

YuriBuerov: why not get rid of found -> break,  and just return like in my example?

fiam: Your example requires the tag order to match, but we want to ignore the order (because once these make their way into graphite, they're unordered)

pkg/telemetry/client.go
```diff
@@ -88,13 +101,73 @@ func NewClient(address string, clientInfo MetricClientInfo, opts ...ClientOption
...
+// ensuring there are no duplicates.
+func (m *Metric) AddTag(name string, value string) error {
+	const (
+		invalidChars       = ";!^="
+		invalidInitialChar = "~"
+	)
+	if name == "" {
+		return errors.New("tag name is empty")
+	}
+	if name[0] == invalidInitialChar[0] {
```
YuriBuerov, line 155: I would use `regexp` this way it will be more readable and even faster than `string.ContainsAny` below
like:

```
var invalidCharsRegex = regexp.MustCompile(`[;!^=]`)
var invalidInitialCharRegex = regexp.MustCompile(`^[~]`)
```

or alternatively `strings.IndexAny`

fiam: Regexp makes the check less readable and the performance difference for the string lengths we're using has to be negligible

pkg/telemetry/client_test.go
```diff
@@ -151,7 +152,72 @@ func TestSendDurationMetric(t *testing.T) {
...
+
+func TestMetricEqual(t *testing.T) {
+	tags := []MetricTag{{"b", "c"}, {"d", "e"}}
+	m1 := &Metric{"a", 1, tags}
+	m2 := &Metric{"a", 1, nil}
+	m3 := &Metric{"a", 3, nil}
+	m4 := &Metric{"b", 1, tags}
+	// Tags in reverse order vs m1, should compare as equal
+	m5 := &Metric{m1.Name, m1.Value, []MetricTag{tags[1], tags[0]}}
+	testCases := []struct {
```
YuriBuerov, line 168:
```
testCases := []struct {
		name string
		a    *Metric
		b    *Metric
		want bool
}{
	{name: "equal", a: m1, b: m1, want: true},
	{name: "not equal", a: m1, b: m2, want: false},
	{name: "not equal", a: m1, b: m3, want: false},
	{name: "not equal", a: m1, b: m4, want: false},
	{name: "equal", a: m1, b: m5, want: true},
}

for _, tc := range testCases {
	t.Run(tc.name, func(t *testing.T) {
		assert.Equal(t, tc.want, tc.a.Equal(tc.b))
		assert.Equal(t, tc.want, tc.b.Equal(tc.a))
	})
}
```

pkg/telemetry/client.go
YuriBuerov: I would not agree with less readable, but ok, it works for me

pkg/telemetry/client_test.go
```diff
@@ -151,7 +152,72 @@ func TestSendDurationMetric(t *testing.T) {
...
+
+func TestMetricAddTag(t *testing.T) {
+	const (
+		tagName  = "tag~"
+		tagValue = "tagValue"
+	)
+
+	m := &Metric{"metric", 1, nil}
+
+	assert.NotNil(t, m.AddTag("", tagValue), "should fail to add empty tag")
```
YuriBuerov, line 205: I would use `assert.Error`

pkg/telemetry/client_test.go
```diff
@@ -151,7 +152,72 @@ func TestSendDurationMetric(t *testing.T) {
...
+	assert.NotNil(t, m.AddTag("~tag", tagValue), "should fail to add invalid tag")
+	for _, ch := range ";!^=" {
```
YuriBuerov, line 207: it would better to reuse const here, so in test and implementation you will always have same values

pkg/telemetry/client_test.go
```diff
@@ -151,7 +152,72 @@ func TestSendDurationMetric(t *testing.T) {
...
+		assert.NotNil(t, m.AddTag("tag"+string(ch), "value"), "should fail to tag with invalid character")
+	}
+	assert.Nil(t, m.AddTag(tagName, tagValue), "should add tag")
+
+	assert.Equal(t, 1, len(m.Tags), "should have one tag")
```
YuriBuerov, line 212: `assert.Len`

YuriBuerov (changes requested):

pkg/telemetry/client.go
fiam: > I would not agree with less readable, but ok, it works for me

I didn't know you enjoyed Perl 😁

pkg/telemetry/client_test.go
fiam: Good idea, I'll move the constants outside the method and reference them from the test

pkg/telemetry/client_test.go
fiam: I'd rather use the struct dump as the test name, I feel it's easier to spot what's actually being tested, but I'll rename the boolean field.

pkg/telemetry/client.go
fiam: Added validation to cloud

jensneuse (approved):
amazing work!

StarpTech (approved):
LGTM

fiam:
@YuriBuerov PTAL

cli/commands/root.go
```diff
@@ -134,9 +122,40 @@ var rootCmd = &cobra.Command{
...
+				AnonymousID:      viper.GetString("anonymousid"),
+			}
+
+			// Check if this command should also send data source related telemetry
+			if telemetry.HasAnnotations(telemetry.AnnotationDataSources, cmd.Annotations) {
+				wunderGraphDir, err := files.FindWunderGraphDir(_wunderGraphDirConfig)
+				if err != nil {
+					return err
+				}
+				dataSourcesMetrics, err := telemetry.DataSourceMetrics(wunderGraphDir)
```
StarpTech, line 139: Loading and parsing the whole file twice on every start may make our cold start slower. Is there any room for improvement? Otherwise, I'd go with it and measure the impact.

StarpTech (changes requested):

cli/commands/root.go
jensneuse: How about this: send a signal when WunderNode is listening. Then do things that are not required for serving requests.

fiam: We're only sending the data source metrics in `wunderctl up`, so it shouldn't have any impact in production. We can add some means for the Node to emit an event when the configuration is parsed, thought, that might come in handy for eventually allowing the testing SDK to modify the config on the fly.

jensneuse (approved):

cli/commands/root.go
StarpTech: Sounds great.

StarpTech (approved):
LGTM

PR #543, opened 2023-01-14
docs: add twitter clone

Slickstef11:
#### Checklist

- [ ] run `make test`


Pagebakers (changes requested):
- `requiresAuthentication: false/true` is not needed on the hooks, it's only used internally and is auto generated.
- The conclusion is not finished?


- Not a blocker, I'd prefer to see Typescript version of this.

PR #546, opened 2023-01-16
fix: Fix query string with similar variables

erguotou520:
Here is the following user case:

```
mutation CreateUser($provider: String!, $providerId: String!, $providerUserId: String!) {
  local_createOneUser(
    data: {provider: $provider, providerId: $providerId, providerUserId: $providerUserId}
  ) {
    avatar
    id
  }
}
```

This will be replaced as 
```
mutation{createOneAppUser(data: {nickname: $$0$$,avatar: $$1$$,provider: $$2$$,providerId: $$2$$Id,providerUserId: $$2$$UserId}){avatar id}}
```

because `providerId` and `providerUserId` was replaced by `$provider` in early position.

So my solution is replace variables by not ending with characters `[^\w\d]`

#### Checklist

- [ ] run `make test`

### DIFF
--- a/pkg/datasources/database/datasource.go
+++ b/pkg/datasources/database/datasource.go
@@ -7,8 +7,8 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"regexp"
 	"runtime"
-	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


Pagebakers:
Thanks a lot for opening this PR. 

We confirm that this is an issue, but we need to do some more investigation to see if this is the right approach.

YuriBuerov:
Thanks for opening this PR and for the bug report. The fix was released. We would greatly appreciate any feedback you have.

PR #549, opened 2023-01-16
fix: test in nextjs-postgres-user-filters

fiam:
It used to work because it relied on mutatingPreResolve not working
for queries


Pagebakers (approved):

StarpTech (approved):
LGTM

PR #551, opened 2023-01-16
feat: make input required if there are required variables

Pagebakers:
This fixes `input` being optional when there are required fields for the TypeScript client, SWR, Next and React Query


StarpTech:
Can't we generate the right type at build time instead of solving that puzzle with ts?

StarpTech (approved):
LGTM

PR #554, opened 2023-01-16
feat: add services spec draft

fiam:

### DIFF
--- a/specs/services.md
+++ b/specs/services.md
@@ -1,0 +1,138 @@
+# WunderGraph Services
+
+This proposal includes a syntax for declaring service dependencies in WunderGraph applications, which
+might will only represent internal services. Network protocols used by these services are considered an
+implementation detail (e.g. FileStorage might use S3, cache might use memcache, etc... - but user should
+not rely on this)
+
+All services are identified by their unique `id`, which allows using them in a handler function with
+a very concise syntax. For this reason, `id` values must be valid TS identifiers. (e.g. `post-images` would
+be invalid and raise an error).
+
+```typescript
+export default createOperation.query({
+  input: z.object({
+    userId: z.number(),
+  }),
+  handler: async ({ services, input }) => {
+    // Use the 'images' service, strongly typed to FileStorage
+    const profilePicUrl = services.images.getPublicUrl(`avatar-${input.userId}`);
+    return {
+      url: profilePicUrl,
+    };
+  },
+});
+```
+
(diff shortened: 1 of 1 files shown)


specs/services.md
```diff
@@ -0,0 +1,138 @@
...
+// configureWunderGraph emits the configuration
+configureWunderGraphApplication({
+ apis: [jsp],
+ services: [images, videos, cache],
+});
+```
+
+## Deprecations
+
+Introducing this API would deprecate the current S3 providers API. Moving forward, plugging arbitrary S3 storage backends
```
Pagebakers, line 137: This would mean we would loose one very important feature of WunderGraph OSS, vendor agnostic.

specs/services.md
```diff
@@ -0,0 +1,138 @@
...
+const fileStorage = services.FileStorage({
+ id: 'images',
+ uploadProfiles: {
+    profilePic: {
+        meta: z.object({userId: z.number()}),
+        ...
+    },
+    ...
+ },
+});
```
YuriBuerov, line 46: could you extend it with examples of configuration for the following use cases:
- local storage provided by WG out of the box, cloud storage provided by WG cloud. (minimum or no configuration needed).
- local storage provided by WG, cloud storage custom 
- both local and cloud storages are custom 

Remember that we must initialize the upload client when we start the node.

specs/services.md
```diff
@@ -0,0 +1,138 @@
...
+Object cache provides the same API as a bare cache.
+
+## Images
+
+An image server that does dynamic image resizing/optimization.
+
+```typescript
+const images = services.Images({
+  id: 'imageServer',
+  storage: fileStorage, // Previously declared as services.FileStorage
```
YuriBuerov, line 110: could you explain how it should work internally and from a configuration perspective? Assuming that we already configured storage, and initialized the upload client, why do we need to pass `fileStorage`  to the image service?

fiam: The idea is that you can have multiple storage services (e.g. one for user images, another one for user videos) and the image server uses one of them specifically.

YuriBuerov: I would say it sounds like we need different buckets in this case, rather than servers. Maybe I'm missing something

fiam: Each service acts like a bucket. The idea is to remove the bucket as an intermediate configurable thing and tie them to a service.

Pagebakers:
Closing this for now, if you want to continue the work you can re-open it in the new rfcs repository.

PR #567, opened 2023-01-19
fix: avoid double escaping special characters in JSON

### LINKED ISSUE #377: Escape characters/unicode characters in operations don't seem to encode properly in request URLs
### Bug description

Escaped quotes/unicode quotes in an operation are *also* encoding the escape character `\` into the request URL.  e.g.

```
query foo_bar {
  data: foo_GetAllBars(
    startkey: "\"bar_2_\u0022",
    endkey: "bar_2_\ufff0"
  ) {
```

Both of the \" and \u0022 are also sending the `\` (encoded as `%5C`) while the `\ufff0` seems to get correctly encoded

_all_docs?endkey=bar_2_%EF%BF%B0&startkey=**%5C**%22bar_2_**%5C**%22

### How to reproduce

1. Add `\"` or `\u0022` to a REST/OpenAPI parameter in an operation
2. Issue a request against the operation
3. Observe the request includes the `\` character encoded as`%5C`

### Expected behavior

Escape characters shouldn't be encoded into the request URL

### Environment & setup

- OS: Mac OS
- Go version: 1.19.3 darwin/arm64
- Database: CouchDB (via OpenAPI/REST)
- Node.js version: 16.13.1

### WunderCtl Version

```
Version: 0.118.1
Commit: 507fcabf52cb0191ea99576f817e1d447e175072
Date: 2022-11-20T10:19:27Z
BuiltBy: ci
```

fiam:
Apply jsonparser.EachKey() returns escaped keys, so we need to unescape
them

Fixes #377


pkg/customhttpclient/customhttpclient.go
```diff
@@ -139,6 +129,36 @@ func respBodyReader(req *http.Request, resp *http.Response) (io.ReadCloser, erro
...
+		)
+		jsonparser.EachKey(value, func(i int, bytes []byte, valueType jsonparser.ValueType, err error) {
+			switch i {
+			case 0:
+				parameterName = unescapeJSON(bytes)
+			case 1:
+				parameterValue = unescapeJSON(bytes)
+			}
+		}, queryParamsKeys...)
+		if len(parameterName) != 0 && len(parameterValue) != 0 {
```
Aenimus, line 146: Are we able to unnest this?

fiam: I don't think so, `jsonparser.EachKey` uses a callback

Aenimus: I just mean `if len(parameterName) != 0 && len(parameterValue) != 0 {`, which is within `jsonparser.ArrayEach`

Aenimus: could we pass or return or otherwise do nothing and un-nest the if else?

jensneuse (changes requested):

pkg/customhttpclient/customhttpclient.go
```diff
@@ -139,6 +129,36 @@ func respBodyReader(req *http.Request, resp *http.Response) (io.ReadCloser, erro
...
+			if bytes.Equal(parameterValue[:1], literal.LBRACK) {
+				_, _ = jsonparser.ArrayEach(parameterValue, func(value []byte, dataType jsonparser.ValueType, offset int, err error) {
+					query.Add(string(parameterName), string(value))
+				})
+			} else {
+				query.Add(string(parameterName), string(parameterValue))
```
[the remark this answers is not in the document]
fiam: We can use unsafe string conversions because those strings are temporary until `encodeQueryParams()` returns. I'll update it.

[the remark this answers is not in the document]
fiam: Changed to `unsafebytes.BytesToString()`

pkg/customhttpclient/customhttpclient.go
```diff
@@ -27,6 +28,17 @@ var (
 	}
 )
 
+// unescapeJSON tries to unescape the given JSON data, returning
+// the original one if the unescaping fails
+func unescapeJSON(in []byte) []byte {
+	// This will only allocate if there are actual escape sequences in the data
+	unescaped, err := jsonparser.Unescape(in, nil)
```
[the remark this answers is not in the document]
fiam: That should be fine because at the end of `encodeQueryParams()` we encode the whole query string

pkg/customhttpclient/customhttpclient.go
fiam: I've changed it to an early return

pkg/customhttpclient/customhttpclient.go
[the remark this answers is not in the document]
fiam: I don't think we can use a pool effectively here. The number of keys is variable and we would need the buffers inside the `jsonparser` callbacks, so we need to allocate to keep track of the buffers that we need to put back. On the other hand, I've realized that the JSON we receive at that point is double encoded, which means we can safely unescape it into itself, hence avoiding any allocations. PTAL

pkg/customhttpclient/customhttpclient.go
```diff
@@ -80,33 +96,11 @@ func DoWithStatus(client *http.Client, ctx context.Context, requestInput []byte,
...
-				if bytes.Equal(parameterValue[:1], literal.LBRACK) {
-					_, _ = jsonparser.ArrayEach(parameterValue, func(value []byte, dataType jsonparser.ValueType, offset int, err error) {
-						query.Add(string(parameterName), string(value))
-					})
-				} else {
-					query.Add(string(parameterName), string(parameterValue))
-				}
-			}
-		})
+		rawQuery, err := encodeQueryParams(queryParams)
```
jensneuse, line 99: If you're using unsafe inside encodeQueryParams, we have to get a buffer from the pool, copy the original queryParams into the buffer, encode it, and defer release the buffer. Otherwise we're messing up "requestInput", which is the source of queryParams. requestInput can be passed to many Load functions, re-using resources. We're not allowed to directly "touch" it. If we want to manipulate it, we either make a copy with unsafe OR we use a string, which implicitly makes a copy (string is an immutable byte slice). I've ran into this issue before with the Load function. So, either we use string, or we have to explicitly make a copy and keep the buffer around as long as we're using the bytes we've touched with unsafe.

pkg/customhttpclient/customhttpclient.go
```diff
@@ -27,6 +30,19 @@ var (
...
+// any allocations. This means that in case of an error the data will be
+// malformed.
+func unescapeJSON(in []byte) ([]byte, error) {
+	// All variables that can be fed into Do()/DoWithStatus() output must be
+	// first formatted as valid JSON. This means the only escaping that will
+	// be added by the second encoding run is prepending a second "layer" of
+	// escaping and this should all be "deescapable" onto the same source
+	// slice because we're always at least writing one byte behind where we
+	// read from.
+	return jsonparser.Unescape(in, in)
```
jensneuse, line 43: I'm not sure if this is correct. Why not just create another buffer outside this function (in Load) from a buffer pool, pass it to this func as out buffer, and release it with defer in Load. This should work fine and will be less error prone.

jensneuse (changes requested):
I suggest 2 small changes, then we're good.

pkg/customhttpclient/customhttpclient.go
fiam: Added a local copy of queryParams with a pooled buffer, PTAL

pkg/customhttpclient/customhttpclient.go
fiam: Added a local copy, as requested. PTAL

jensneuse (approved):

jensneuse (approved):

StarpTech (approved):
lgtm

PR #583, opened 2023-01-22
fix: use operationPath in client operation metadata

### LINKED ISSUE #581: X-CSRF-Token header is missing due to mismatch in operationName and OperationMetadata keys
### Bug description

In the newest @wundergraph/sdk@0.132.0, I noticed that the `X-CSRF-Token` header is no longer present in the requests, causing the WunderGraph backend to reject the request.

I think this is due to [this comparison](https://github.com/wundergraph/wundergraph/blob/d7e50cd91d4d18dda7331b348b737ad75a770380/packages/sdk/src/client/client.ts#L242), where the keys in `operationMetadata` no longer match the actual operation names, causing this if condition to always evaluate to false. For example, if I define an operation in `.wundergraph/operations/images/GetImages.graphql`, the `operationName` becomes `images/GetImages` while the key in operationMetadata is `ImagesGetImages`.

### How to reproduce

see above

### Expected behavior

_No response_

### WunderGraph information

@wundergraph/sdk@0.132.0

### Environment & setup

- OS: 
- Go version: 
- Database: 
- Node.js version: 

### WunderCtl Version

wunderctl 0.127.0

spetrunin:
fixes #581

#### Checklist

- [ ] run `make test`

### DIFF
--- a/packages/sdk/src/codegen/templates/typescript/client.template.ts
+++ b/packages/sdk/src/codegen/templates/typescript/client.template.ts
@@ -88,7 +88,7 @@ export const defaultClientConfig: ClientConfig = {
 
 export const operationMetadata: OperationMetadata = {
 {{#each allOperations}}
-    {{operationName}}: {
+    "{{operationPath}}": {
         requiresAuthentication: {{requiresAuthentication}}
 		}
     {{#unless @last}},{{/unless}}
--- a/packages/sdk/src/client/client.ts
+++ b/packages/sdk/src/client/client.ts
@@ -237,12 +237,7 @@ export class Client {
 
 		const headers: Headers = {};
 
-		if (
-			this.options.operationMetadata &&
-			this.options.operationMetadata[options.operationName] &&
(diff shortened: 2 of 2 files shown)


packages/sdk/src/codegen/templates/typescript/client.template.ts
```diff
@@ -88,7 +88,7 @@ export const defaultClientConfig: ClientConfig = {
 
 export const operationMetadata: OperationMetadata = {
 {{#each allOperations}}
-    {{operationName}}: {
+    "{{operationPath}}": {
```
fiam, line 91: This needs to be quoted because it can now contain /

jensneuse (approved):

packages/sdk/src/codegen/templates/typescript/client.template.ts
spetrunin: didn't get you, it is already quoted

fiam (approved):
LGTM

PR #584, opened 2023-01-23
fix: mark json-schema-to-typescript as dependency instead of devDependency

fiam:


Pagebakers (approved):
Let's get this released.

jensneuse (approved):

PR #595, opened 2023-01-24
chore: improve functions

jensneuse:


jensneuse:
TODO: @Pagebakers please fix the errors in the client test

packages/sdk/src/operations/operations.ts
```diff
@@ -34,7 +47,7 @@ interface LiveQueryConfig {
 }
 
 const createQuery =
-	<IC extends InternalClient, UserRole extends string>() =>
+	<IC extends InternalClient, UserRole extends string, Queries, Mutations, Subscriptions>() =>
```
Pagebakers, line 50: I'd prefer to see Queries/Mutations and Subscriptions be merged into one generic `Operations`, cleaner and consistent with all frontend clients.

packages/sdk/src/server/operations-client.ts
```diff
@@ -0,0 +1,134 @@
...
+				Accept: 'text/event-stream',
+				...(this.options.extraHeaders || {}),
+			}
+		);
+		const input = (args as OperationArgs<T, ExtractInput<Subscriptions[T]>>).input || undefined;
+		const body = JSON.stringify({ input, __wg: { clientRequest: this.options.clientRequest } });
+		const abort = new AbortController();
+		const generator = async function* () {
+			try {
+				const res = await fetch(url, {
```
Pagebakers, line 101: fetch in nodejs is still experimental

fiam:
I've run this against all the test in examples locally and they pass

Pagebakers (changes requested):
Overall nice improvement, but I'm not sure why we couldn't re-use the Typescript client here with simply setting a custom baseURL.

packages/sdk/src/operations/operations.ts
fiam: It can't be merged into one generic `Operations` because otherwise then we can't constrain on them as template parameters (not exactly in this spot, but somewhere else along the chain).

You can constrain a generic as `<..., TQueries extends Operations, TMutations extends Operations, TSubscriptions extends Operations>` and then replace then with subtypes.

On the other hand, if you use `<... Operations extends Operations<Operation, Operation, Operation>` then `Operations<TQueries, Mutation, TSubscription>` doesn't extend `Operations<Operation, Operation, Operation>`

packages/sdk/src/server/operations-client.ts
fiam: I've added support for using a custom fetch function. PTAL

packages/sdk/src/server/operations-client.ts
```diff
@@ -31,6 +31,7 @@ export interface Options {
 	baseURL: string;
 	extraHeaders?: { [key: string]: string };
 	clientRequest: any;
+	customFetch?: (input: URL | RequestInfo, init?: RequestInit) => Promise<globalThis.Response>;
```
Pagebakers, line 34: Sorry but we have to revert this, this will still result in a broken state if fetch is not enabled in node (and older node versions). There is no way to configure a custom fetch anywhere, the operations client is initiated in the fastify server.

https://github.com/wundergraph/wundergraph/pull/595/files#diff-2f9c5d63c84e8681f889caeb22b494c16f93545d9368d37cf2d15829be1db0b5R38

I'd suggest to use axios like in the internalClient for now, so we don't need to introduce new dependencies and we can replace it once node fetch is GA.

Pagebakers (changes requested):

packages/sdk/src/server/operations-client.ts
fiam: Thanks for pointing it out. As discussed, we've used `cross-fetch` instead for easier forwards compatibility

Pagebakers (approved):

packages/sdk/src/server/operations-client.ts
```diff
@@ -0,0 +1,136 @@
...
+				'Content-Type': 'application/json',
+				Accept: 'text/event-stream',
+				...(this.options.extraHeaders || {}),
+			}
+		);
+		const input = (args as OperationArgs<T, ExtractInput<Subscriptions[T]>>).input || undefined;
+		const body = JSON.stringify({ input, __wg: { clientRequest: this.options.clientRequest } });
+		const abort = new AbortController();
+		const generator = async function* () {
+			try {
```
StarpTech, line 102: How does error handling work when try has no catch?

StarpTech (changes requested):

packages/sdk/src/server/operations-client.ts
fiam: I think this `try` is only for the `finally` on line 128

packages/sdk/src/server/operations-client.ts
```diff
@@ -0,0 +1,136 @@
...
+					const data = decoder.decode(value);
+					buffer += data;
+					if (buffer.endsWith('\n\n')) {
+						const json = JSON.parse(buffer.substring(0, buffer.length - 2));
+						yield json as Subscriptions[T] extends { response: any } ? Subscriptions[T]['response'] : never;
+						buffer = '';
+					}
+				}
+			} finally {
+				abort.abort();
```
StarpTech, line 129: @fiam I mean, in my opinion, the error must be rethrown here, so the user can react and log errors.

StarpTech (changes requested):

packages/sdk/src/server/operations-client.ts
StarpTech: Nice, thank you.

StarpTech (approved):
LGTM

fiam (approved):

PR #599, opened 2023-01-25
feat: load telemetry anonymous id from WG_TELEMETRY_ANONYMOUS_ID

jensneuse:


Pagebakers (approved):

pkg/config/config.go
```diff
@@ -62,11 +62,17 @@ func initViper(isTelemetryEnabled bool) {
...
-		viper.Set("anonymousID", ksuid.New().String())
-		allSettings := viper.AllSettings()
-		if err := SaveConfig(allSettings); err != nil {
-			log.Printf("Could not save config: %v\n", err)
+	if isTelemetryEnabled {
+		telemetryID, exists := os.LookupEnv("WG_TELEMETRY_ANONYMOUS_ID")
+		if exists {
+			viper.Set("anonymousID", telemetryID)
+		}
+		if !viper.InConfig("anonymousID") {
```
StarpTech, line 70: 1. Use `else if` ?
2. Intentionally to not save it to disk?

pkg/telemetry/client.go
```diff
@@ -61,6 +63,8 @@ func NewClient(address string, clientInfo MetricClientInfo, opts ...ClientOption
 		clientInfo: clientInfo,
 	}
 
+	c.authToken = os.Getenv("WG_TELEMETRY_AUTH_TOKEN")
```
StarpTech, line 66: The config should come from the options. Please create an option and set it at creation.

pkg/telemetry/client.go
```diff
@@ -137,6 +141,9 @@ func (c *client) Send(metrics []Metric) error {
 
 	req.Header.Set("User-Agent", userAgent)
 	req.Header.Add("Content-Type", "application/json")
+	if c.authToken != "" {
+		req.Header.Add("X-WG-TELEMETRY-AUTHORIZATION", fmt.Sprintf("Bearer %s", c.authToken))
```
StarpTech, line 145: Why not using `Authorization`?

StarpTech (changes requested):

pkg/telemetry/client.go
jensneuse: we might later want to use the Authorization header differently, so I'd rather not block this use case

pkg/config/config.go
jensneuse: I've tested if it needs to be saved before being able to read it back. It works correctly without saving.
I don't want to save it because in this particular use case, saving doesn't have any effect as the storage is ephemeral.
With non ephemeral storage, e.g. on premises, this env var should never be set.
So really, the use case only happens in our cloud and saving it there doesn't work.

pkg/telemetry/client.go
StarpTech: ok

pkg/telemetry/client.go
jensneuse: changed

pkg/config/config.go
StarpTech: ok

StarpTech (approved):
LGTM

PR #602, opened 2023-01-26
chore: fix check for prettier formatted files in CI

fiam:
Tell prettier to ignore unknown extensions, and also exclude .html
files since we do not run those by prettier

### DIFF
--- a/.github/workflows/packages-ci.yaml
+++ b/.github/workflows/packages-ci.yaml
@@ -93,7 +93,7 @@ jobs:
           main-branch-name: ${{ steps.branch-name.outputs.current_branch }} # Get the last successful commit for the current branch.
 
       - name: Check for formatting errors
-        run: eval "pnpx prettier --check $(git diff --name-only --diff-filter=ACMRT -r ${{ steps.last_successful_commit_push.outputs.base }} | xargs)"
+        run: eval "pnpx prettier --check -u $(git diff --name-only --diff-filter=ACMRT -r ${{ steps.last_successful_commit_push.outputs.base }} | xargs)"
 
       - name: Check if git is not dirty after generating files
         run: git diff --no-ext-diff --exit-code
--- a/.prettierignore
+++ b/.prettierignore
@@ -22,6 +22,9 @@ logs
 
 CHANGELOG.md
 
+# Ignore html files
+*.html
+
 # Ignore go files
 *.go
 


.github/workflows/packages-ci.yaml
```diff
@@ -93,7 +93,9 @@ jobs:
           main-branch-name: ${{ steps.branch-name.outputs.current_branch }} # Get the last successful commit for the current branch.
 
       - name: Check for formatting errors
-        run: eval "pnpx prettier --check $(git diff --name-only --diff-filter=ACMRT -r ${{ steps.last_successful_commit_push.outputs.base }} | xargs)"
+        # Ignore unknown file extensions for prettier as well as files with .html extension, since
+        # we don't format those with prettier.
+        run: eval "pnpx prettier --check -u $(git diff --name-only --diff-filter=ACMRT -r ${{ steps.last_successful_commit_push.outputs.base }} | grep -v '\.html$' | xargs)"
```
StarpTech, line 98: IMO this is the wrong approach. Prettier has an ignore file. We need to exclude patterns here https://github.com/wundergraph/wundergraph/blob/main/.prettierignore

StarpTech (changes requested):

.github/workflows/packages-ci.yaml
StarpTech: Prettier has also an https://prettier.io/docs/en/cli.html#--ignore-unknown option

fiam: It seems to ignore `.prettierignore` for command line arguments. Let me give it another try. `--ignore-unknown` is in the PR, `-u` is its short form.

StarpTech: Or try this https://github.com/azz/pretty-quick, which should work.

fiam: @StarpTech I'd rather not take a dependency when we're 95% there. It should be good now, PTAL.

StarpTech: If it works fine! :smile:

StarpTech (approved):
LGTM

.github/workflows/packages-ci.yaml
fiam: Thanks 👍

jensneuse (approved):

jensneuse:
Thank you @fiam !

PR #604, opened 2023-01-26
fix: examples, simple

YuriBuerov:
Upstream issue. 

#### Checklist

- [ ] run `make test`

### DIFF
--- a/examples/simple/.wundergraph/operations/Countries.graphql
+++ b/examples/simple/.wundergraph/operations/Countries.graphql
@@ -1,4 +1,4 @@
-query Countries($filter: countries_CountryFilterInput) {
+query Countries($filter: countries_CountryFilterInput) @removeNullVariables {
 	countries_countries(filter: $filter) {
 		code
 		name
--- a/examples/simple/.wundergraph/operations/Countries.graphql
+++ b/examples/simple/.wundergraph/operations/Countries.graphql
@@ -1,7 +1,0 @@
-query Countries($filter: countries_CountryFilterInput) @removeNullVariables {
-	countries_countries(filter: $filter) {
-		code
-		name
-		capital
-	}
-}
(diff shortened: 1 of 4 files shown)


jensneuse:
> Upstream issue.
> 

Is the problem with the origin? Is it not possible to fix it in the engine?

YuriBuerov:
> > Upstream issue.
> 
> Is the problem with the origin? Is it not possible to fix it in the engine?

correct

examples/simple/.wundergraph/operations/Countries.graphql
```diff
@@ -1,4 +1,4 @@
-query Countries($filter: countries_CountryFilterInput) {
+query Countries($filter: countries_CountryFilterInput) @removeNullVariables {
```
jensneuse, line 1: Can we just switch it to `countries_country(code: "US")`?
It might be confusing if you have to use such a directive in the first example.

jensneuse (changes requested):

examples/simple/.wundergraph/operations/Countries.graphql
fiam: I agree, I think it would be better if we could avoid introducing `@removeNullVariables` in an example called `simple`. What about giving the filter a default value of `{code:{ne:""}`?

YuriBuerov: for me, it looks even more unclear.

examples/simple/README.md
```diff
@@ -15,9 +15,7 @@ curl http://localhost:9991/operations/Continents
 #### Get Andorra
```
fiam, line 15: This subtitle needs to be updated too

jensneuse (approved):

fiam (approved):
LGTM

PR #605, opened 2023-01-26
feat: add prisma datasource

jensneuse:
This PR adds the "prisma" data source.
What's different between this ds and others, like PostgreSQL or SQLite, is that it doesn't introspect the database.
It only looks at the prisma schema and generates a data-source & GraphQL Schema from it.
This is very useful in case you don't want to build a GraphQL schema from the database, but rather from the prisma Schema.
E.g. Planetscale doesn't support relationships, so a database-introspected schema will not contain relationships.
However, you can add those in the prisma schema and then generate the GraphQL layer from the prisma schema.

tldr with this PR, we can properly support Planetscale, MongoDB and other use cases where the prisma schema differs from an database-introspected prisma schema.


Makefile
```diff
@@ -49,6 +49,10 @@ codegen: install-proto codegen-go
 build: codegen
 	cd cmd/wunderctl && go build -o ../../wunderctl -ldflags "-X 'main.commit=$(shell git rev-parse --short HEAD)' -X 'main.builtBy=dev' -X 'main.version=dev' -X 'main.date=$(shell date)'" -trimpath
 
+build-wrapper: build
```
spetrunin, line 52: do we need this command? is the existing one not enough?

```
wunderctl: build
	pnpm -r run --filter="./packages/wunderctl" build
	rm -f packages/wunderctl/download/wunderctl
	cp -f wunderctl packages/wunderctl/download/wunderctl
	rm wunderctl
```

packages/sdk/src/definition/index.ts
```diff
@@ -435,6 +446,8 @@ const databaseSchemaToKind = (schema: DatabaseSchema): DataSourceKind => {
 			return DataSourceKind.SQLSERVER;
 		case 'mongodb':
 			return DataSourceKind.MONGODB;
+		case 'prisma':
```
Aenimus, line 449: Could these database types not be a const somewhere? I've seen the same string literals at least 3 times

cli/commands/prisma.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"io/ioutil"
+	"path/filepath"
+
+	"github.com/spf13/cobra"
+
+	"github.com/wundergraph/wundergraph/pkg/files"
+)
+
+// postgresCmd represents the postgres command
```
YuriBuerov, line 12: typo?

cli/commands/mongodb.go
```diff
@@ -19,7 +19,7 @@ var mongoDbCmd = &cobra.Command{
   url      = "%s"
 }
 `, databaseURL)
-		return introspectDatabase(introspectionSchema)
+		return introspectDatabase(introspectionSchema, true)
```
spetrunin, line 22: Having separate files for db-specific cmds seems obsolete for me, we could move all cmd into `introspect.go`  which could be also renamed to `introspectdb.go`

and have some helper like this
```go
func cmdFor(provider string) func(cmd *cobra.Command, args []string) error {
  return func(cmd *cobra.Command, args []string) error {
		  databaseURL := args[0]
		  introspectionSchema := fmt.Sprintf(`datasource db {
    provider = "%s"
    url      = "%s"
  }
  `, provider, databaseURL)
		  return introspectDatabase(introspectionSchema, true)
	  },
}
```

than we just will have it like this

```go
var mongoDbCmd = &cobra.Command{
	Use:     "mongodb",
	Short:   "Introspects a mongodb database",
	Example: `wunderctl mongodb mongodb+srv://test:test@cluster0.ns1yp.mongodb.net/myFirstDatabase`,
	Args:    cobra.ExactArgs(1),
	RunE: cmdFor("mongodb"),
}

func init() {
	introspectCmd.AddCommand(mongoDbCmd)
  introspectCmd.AddCommand(mysqlCmd)
  // etc
}
```

spetrunin: also `Example:` is not correct for mongodb and sqlserver

cli/commands/prisma.go
```diff
@@ -0,0 +1,37 @@
...
+	RunE: func(cmd *cobra.Command, args []string) error {
+		schemaFilePath := args[0]
+		wunderGraphDir, err := files.FindWunderGraphDir(_wunderGraphDirConfig)
+		if err != nil {
+			return err
+		}
+		if !filepath.IsAbs(schemaFilePath) {
+			schemaFilePath = filepath.Join(wunderGraphDir, schemaFilePath)
+		}
+		content, err := ioutil.ReadFile(schemaFilePath)
```
YuriBuerov, line 27: `ioutil` is deprecated now it's: `os.ReadFile`

YuriBuerov (approved):
just a few minor comments

cli/commands/mongodb.go
jensneuse: yes, good idea for next refactor =)

cli/commands/prisma.go
fiam: I second this. No reason to to rewrite the ones that we have in the codebase, but it would be nice to avoid introducing new instances.

packages/sdk/src/definition/index.ts
jensneuse: I've checked, it's exactly once.

cli/commands/prisma.go
jensneuse: yes

Makefile
jensneuse: you're right

cli/commands/prisma.go
jensneuse: replaced

packages/sdk/src/definition/index.ts
jensneuse: I've moved the types to a const declaration.

Aenimus (approved):

PR #607, opened 2023-01-27
fix: paths and uri composition on Windows when using nested directories for operations

fiam:
- Use filepath.Join() to manipulate file paths in the Go side
- In the typescript side, avoid path.join() to manipulate URL
- Fix file path to module import path conversion on Windows

### DIFF
--- a/packages/sdk/src/server/plugins/functions.ts
+++ b/packages/sdk/src/server/plugins/functions.ts
@@ -15,7 +15,7 @@ const FastifyFunctionsPlugin: FastifyPluginAsync<FastifyFunctionsOptions> = asyn
 	for (const operation of config.operations) {
 		try {
 			const filePath = path.join(process.env.WG_DIR_ABS!, operation.module_path);
-			const routeUrl = '/functions/' + operation.api_mount_path;
+			const routeUrl = `/functions/${operation.api_mount_path}`;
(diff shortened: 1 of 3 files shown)


YuriBuerov (approved):

packages/sdk/src/server/plugins/functions.ts
```diff
@@ -15,7 +15,7 @@ const FastifyFunctionsPlugin: FastifyPluginAsync<FastifyFunctionsOptions> = asyn
 	for (const operation of config.operations) {
 		try {
 			const filePath = path.join(process.env.WG_DIR_ABS!, operation.module_path);
-			const routeUrl = path.join('/functions', operation.api_mount_path);
+			const routeUrl = '/functions/' + operation.api_mount_path;
```
Aenimus, line 18: Do we prefer concatenation or interpolation? Just asking because I am not sure.

fiam: Interpolation looks better, I’ll change it

jensneuse (approved):

packages/sdk/src/server/plugins/functions.ts
fiam: Changed to interpolation

PR #609, opened 2023-01-27
fix: datasource, database

YuriBuerov:
Addressed issue: https://github.com/wundergraph/wundergraph/pull/546

#### Checklist

- [ ] run `make test`

### DIFF
--- a/pkg/datasources/database/datasource.go
+++ b/pkg/datasources/database/datasource.go
@@ -7,8 +7,8 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"regexp"
 	"runtime"
-	"strings"
 	"sync"
 	"time"
 
@@ -49,8 +49,9 @@ type Planner struct {
 
 	engineFactory *LazyEngineFactory
 
-	debug bool
-	log   *zap.Logger
+	debug           bool
+	testsSkipEngine bool
+	log             *zap.Logger
 
 	insideJsonField bool
 	jsonFieldRef    int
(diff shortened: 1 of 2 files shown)


pkg/datasources/database/datasource.go
```diff
@@ -173,18 +174,29 @@ func (p *Planner) ConfigureFetch() plan.FetchConfiguration {
 			Renderer: renderer,
 		}
 		replacement, _ := p.variables.AddVariable(variable)
-		input.Query = strings.Replace(input.Query, currentName, replacement, -1)
+
+		re, err := regexp.Compile(fmt.Sprintf(`%s\b`, regexp.QuoteMeta(currentName)))
```
spetrunin, line 178: do u think we cover the case with leading characters?

like `theprovider` ?

pkg/datasources/database/datasource_test.go
```diff
@@ -0,0 +1,145 @@
...
+	operationWithSimilarAgrNames = `
+		mutation CreateUser($provider: String!, $providerId: Int!) {
+			createOneusers(data: {provider: $provider, providerId: $providerId, provider2: $provider}) {
+				id
+			}
+		}
+	`
+)
+
+func changeJsonRootType(f *resolve.GraphQLVariableRenderer, v resolve.JsonRootType) {
```
spetrunin, line 45: will it be enough to have a such constructor of the variable renderer to be able to get rid of reflection?
https://github.com/wundergraph/graphql-go-tools/pull/493/files

YuriBuerov: approved your PR

spetrunin (approved):

fiam (approved):
I do not like the fact that we're ignoring errors in `ConfigureFetch()`, but we need to make changes in graphql-go-tools to address that. Let's merge this now and find a better solution later.

PR #618, opened 2023-01-31
chore: typecheck examples during CI

fiam:
Don't just execute them, always validate the types


scripts/test-examples/single.sh
```diff
@@ -103,6 +99,17 @@ if grep -q '"test:playwright"' package.json; then
     WG_NODE_URL=${default_node_url} npm run test:playwright
 fi
 
+# If the example uses Next.js, compile it
+if grep -q '"build:next"' package.json; then
+    # This example doesn't build under a pnpm workspace
+    if !grep -q '"nextjs-react-query"' package.json; then
+        npm run build:next
+    fi
+elif grep -q '"check"' package.json; then
```
StarpTech, line 108: `check` should be part of the build script. We should split it.

StarpTech (changes requested):

scripts/test-examples/single.sh
fiam: All examples have `build` call `wunderctl generate`, while `check` calls `tsc --noEmit`. We've had that for both examples and testapps for quite long. Are you suggesting we change all those `package.json` to make `build` also call `check?

fiam: I was not aware of those plans. So should use the following commands in `package.json`: `"generate": "wunderctl generate"`, `"build": "npm run generate && npm run check"`?

fiam: I was not aware of those plans. So should we add `"build:wundergraph": "wunderctl generate"` and `"build": "build:wundergraph && npm run check`

fiam: No idea how the double comment happened, apologies for the inconvenience!

StarpTech: Not set in stone. I'd make tsc check part of `build`.

fiam: As discussed, we're splitting that into a separate issue. Let's concentrate on fixing type-checking in the examples for this one. PTAL.

StarpTech (approved):
LGTM

Aenimus (approved):

PR #621, opened 2023-01-31
feat: new open api introspection

spetrunin:
#### Checklist

- [ ] run `make test`


packages/sdk/src/rest2graphql/introspection.ts
```diff
@@ -0,0 +1,92 @@
...
+
+export const openApiSpecificationToGraphQLApi = async (
+	oas: string,
+	introspection: OpenAPIIntrospection
+): Promise<GraphQLApi> => {
+	const spec = readSpec(oas, introspection.source);
+	let apiID: string;
+
+	if (introspection.id) {
+		if (!introspection.id.match(/^[_\-0-9a-zA-z]+$/)) {
```
Aenimus, line 21: I believe it should be`^[a-zA-Z_][a-zA-Z0-9_]+$`

Also, your second `z` is lowercase.

packages/sdk/src/configure/index.ts
```diff
@@ -317,6 +317,18 @@ const resolveConfig = async (config: WunderGraphConfigApplicationConfig): Promis
 
 	const resolved = await resolveApplication(roles, config.apis, cors, config.s3UploadProvider || []);
 
+	for (const ds of resolved.EngineConfiguration.DataSources) {
+		if (ds.Kind === DataSourceKind.GRAPHQL) {
+			const custom = ds.Custom as unknown as GraphQLApiCustom;
+			if (
+				custom.Fetch.baseUrl?.kind === ConfigurationVariableKind.STATIC_CONFIGURATION_VARIABLE &&
+				custom.Fetch.baseUrl.staticVariableContent.includes('<SERVER_URL>')
```
jensneuse, line 325: Can you add a comment to explain this behavior?

packages/sdk/src/rest2graphql/execution.ts
```diff
@@ -0,0 +1,82 @@
...
+
+export const openApiSpecsLocation = path.join('generated', 'openapi');
+
+interface OpenApiSpec {
+	content: OasSpec;
+	baseURL: InputVariable | undefined;
+	headers: string[];
+}
+
+export const createExecutableSchema = async (specName: string): Promise<GraphQLSchema> => {
```
jensneuse, line 20: what is a specName?

packages/sdk/src/rest2graphql/execution.ts
```diff
@@ -0,0 +1,82 @@
...
+
+			spec.headers.reduce((acc, header: string) => {
+				const value = ctx.wundergraph?.headers[header];
+				if (value) {
+					acc[header] = value;
+				}
+				return acc;
+			}, headers);
+
+			ctx.wundergraph?.log?.info(`created headers ${JSON.stringify(headers)}`);
```
jensneuse, line 47: debug log?

packages/sdk/src/rest2graphql/introspection.ts
```diff
@@ -0,0 +1,92 @@
...
+	writeApiInfo(apiID, spec, introspection);
+
+	const { schema } = await createGraphQLSchema(spec, {
+		fillEmptyResponses: true,
+		viewer: false,
+	});
+
+	return introspect.graphql({
+		url: '',
+		baseUrl: '<SERVER_URL>',
```
jensneuse, line 39: What is `<SERVER_URL>`?

jensneuse (changes requested):
Looks great. My only concern is that we're generated GraphQL engines for all OAS files at startup, which will affect cold starts. One consideration I'd make is to lazy load the schemas on the fastify server. E.g. create handlers for all schemas, etc. but only initialize the schema when the first request hits this particular endpoint. E.g. you can create a factory that returns a promise. The first request will trigger the promise to resolve the executable schema. All subsequent requests will also wait for this promise. Once it's resolved, the promise will immediately return the schema. This way, we're only really initializing the GraphQL JS Engine/Executable Schema when it's needed.

packages/sdk/src/rest2graphql/introspection.ts
spetrunin: It is a server url placeholder

We know what it is only at configure stage, but the introspection promise is prepared earlier

So once introspection is resolved we are replacing this placeholder at configure stage

open to ideas on how to do it better

packages/sdk/src/rest2graphql/execution.ts
spetrunin: it is a name of a file in `generated/openapi` folder

packages/sdk/src/configure/index.ts
spetrunin: explained it in other comment

as this is a POC - this behaviour could be changed

packages/sdk/src/rest2graphql/execution.ts
spetrunin: yeah, in case anyone wants to play with POC - we will see that headers are picked up properly

packages/sdk/src/rest2graphql/introspection.ts
```diff
@@ -0,0 +1,92 @@
...
+		if (!introspection.id.match(/^[_\-0-9a-z]+$/)) {
+			throw new Error('Invalid characters in api id - please use only alphanumeric characters, dashes and underscores');
+		}
+		apiID = introspection.id;
+	} else {
+		apiID = objectHash(oas);
+	}
+
+	// TODO: in case apiID is changed the old one file still will exists. what could be a better place to cache the file?
+	writeApiInfo(apiID, spec, introspection);
```
spetrunin, line 30: it could be a problem, cause server didn't know what is actual version of a cached spec

And we will rewrite a file each time id or hash is changed, so we could have a couple of files in this case
And server will load each file as an api

For the POC - this OAS data source becomes just graphql API, so we don't have any additional information about it, cause engine don't need that info and for our api class Api<T> - T is only related to datasources

testapps/simple/.wundergraph/wundergraph.config.ts
```diff
@@ -2,9 +2,23 @@ import { configureWunderGraphApplication, cors, EnvironmentVariable, introspect,
 import server from './wundergraph.server';
 import operations from './wundergraph.operations';
 
+const stripe = introspect.openApiV2({
```
spetrunin, line 5: this is temp code, as stripe spec is very big and should not be stored in the repo

packages/sdk/src/server/plugins/graphql.ts
```diff
@@ -47,8 +51,19 @@ const FastifyGraphQLPlugin: FastifyPluginAsync<GraphQLServerConfig> = async (fas
 				query: req.query,
 			};
 
+			const headers = Object.entries(request.headers).reduce((acc, [key, value]) => {
```
spetrunin, line 54: not so clean solution, open to any suggestion how to do it better

packages/sdk/src/rest2graphql/execution.ts
```diff
@@ -0,0 +1,82 @@
...
+};
+
+const forwardHeaders = (introspection: OpenAPIIntrospection): string[] => {
+	const headersBuilder = new HeadersBuilder();
+
+	if (introspection.headers !== undefined) {
+		introspection.headers(headersBuilder);
+	}
+
+	return headersBuilder.build().map((value) => value.key.toLowerCase());
```
spetrunin, line 81: header name converted to lowercase, seems in fastify headers are always lowercased

packages/sdk/src/rest2graphql/introspection.ts
```diff
@@ -0,0 +1,92 @@
...
+	});
+
+	return introspect.graphql({
+		url: '',
+		baseUrl: '<SERVER_URL>',
+		path: `/openapis/${apiID}/graphql`,
+		apiNamespace: introspection.apiNamespace,
+		internal: true,
+		loadSchemaFromString: () => printSchema(schema),
+		headers: introspection.headers,
```
spetrunin, line 44: How headers works:

Headers builder is used by graphql datasource and datasource will process headers and forward proper headers from request to wundergraph server, the only thing that we need is to read these headers instead of client request headers and add them to graphql context

packages/sdk/src/rest2graphql/execution.ts
Pagebakers: Not very safe, we would output secrets in info now.

Pagebakers:
It's actually pretty fast on my dev environment using the stripe.yaml spec.

There are some quirky field names though. Why does it append 2 after these queries?
Why do some have v1 and others not.

Screenshot 2023-02-09 at 12 10 32
Screenshot 2023-02-09 at 12 10 11

Pagebakers:
Introspection keycloak works as well out of the box.

Screenshot 2023-02-09 at 14 03 40

packages/sdk/src/rest2graphql/execution.ts
StarpTech: Good point, don't log all headers by default.

packages/sdk/src/rest2graphql/execution.ts
```diff
@@ -0,0 +1,82 @@
...
+			return headers;
+		},
+	});
+
+	return schema;
+};
+
+export const writeApiInfo = (name: string, spec: OasSpec, introspection: OpenAPIIntrospection) => {
+	const specsFolderPath = path.join(process.env.WG_DIR_ABS!, openApiSpecsLocation);
+	fs.mkdir(specsFolderPath, { recursive: true }, (err) => {
```
StarpTech, line 58: Use async variant.

packages/sdk/src/rest2graphql/execution.ts
```diff
@@ -0,0 +1,82 @@
...
+	const fileName = `${name}.json`;
+	const filePath = path.join(specsFolderPath, fileName);
+
+	const item: OpenApiSpec = {
+		content: spec,
+		baseURL: introspection.baseURL,
+		headers: forwardHeaders(introspection),
+	};
+
+	fs.writeFileSync(filePath, JSON.stringify(item, null, 2));
```
StarpTech, line 71: Here too

packages/sdk/src/rest2graphql/introspection.ts
```diff
@@ -0,0 +1,92 @@
...
+
+const tryReadSpec = (spec: string): OasSpec => {
+	try {
+		return JSON.parse(spec);
+	} catch (e) {
+		const obj = yaml.load(spec) as OasSpec;
+		if (obj) {
+			return obj;
+		}
+		throw new Error('cannot read OAS');
```
StarpTech, line 90: embed the original error

packages/sdk/src/server/plugins/graphql.ts
```diff
@@ -28,27 +28,53 @@ export interface GraphQLServerConfig {
...
+					acc[key] = value.join(',');
+				} else if (value !== undefined) {
+					acc[key] = value;
+				}
+				return acc;
+			}, {} as Record<string, string>);
+
 			const pluginLogger = req.ctx.log.child({ server: config.serverName, plugin: 'graphql' });
 
+			pluginLogger.info(`GraphQL request  ${JSON.stringify(request)}`);
```
StarpTech, line 76: I think it was for debugging purpose. Fastify already logs requests.

packages/sdk/src/server/plugins/graphql.ts
StarpTech: Why is it necessary at all? Can't we pass the object as it is?

packages/sdk/src/server/server.ts
```diff
@@ -230,6 +232,29 @@ export const createServer = async ({
 				fastify.log.info(`Graphql server '${server.serverName}' listening at ${routeUrl}`);
 			}
 		}
+
+		const openApisSpecsPath = path.join(wundergraphDir, openApiSpecsLocation);
+		const openApisExists = fs.existsSync(openApisSpecsPath);
+		if (openApisExists) {
+			const specPaths: string[] = [];
+			fs.readdirSync(openApisSpecsPath).forEach((file) => {
```
StarpTech, line 240: Use async versions

packages/sdk/src/server/server.ts
```diff
@@ -230,6 +232,29 @@ export const createServer = async ({
...
+				const ext = path.extname(specPath);
+				const apiName = path.basename(specPath, ext);
+				const routeUrl = openApiServerMountPath(apiName);
+
+				fastify.register(require('./plugins/graphql') as FastifyPluginAsync<LazyGraphQLServerConfig>, {
+					serverName: apiName,
+					schema: () => createExecutableSchema(specPath),
+					routeUrl: routeUrl,
+				});
+				fastify.log.info('GraphQL plugin registered');
```
StarpTech, line 254: Registering plugins is not sync. You need to put the logs in after `fastify.after(() => )`

packages/sdk/src/server/server.ts
```diff
@@ -230,6 +232,29 @@ export const createServer = async ({
...
+			fs.readdirSync(openApisSpecsPath).forEach((file) => {
+				specPaths.push(file);
+			});
+
+			for await (const specPath of specPaths) {
+				const ext = path.extname(specPath);
+				const apiName = path.basename(specPath, ext);
+				const routeUrl = openApiServerMountPath(apiName);
+
+				fastify.register(require('./plugins/graphql') as FastifyPluginAsync<LazyGraphQLServerConfig>, {
```
StarpTech, line 249: `require` once before the loop. It is cached but will still make several calls under the hood.

packages/sdk/src/rest2graphql/execution.ts
```diff
@@ -0,0 +1,82 @@
...
+	const { schema } = await createGraphQLSchema(spec.content, {
+		fillEmptyResponses: true,
+		baseUrl: baseUrl,
+		viewer: false,
+		headers: function (method, path, title, resolverParams) {
+			const ctx = resolverParams?.context as ExecutionContext;
+
+			const headers: Record<string, string> = {};
+
+			spec.headers.reduce((acc, header: string) => {
```
StarpTech, line 39: I think this can be simplified to 

```ts
	const headers = {};
		for (const [key, value] of req.body.__wg.clientRequest?.headers.entries()) {
			headers[key] = value;
		}
```

StarpTech (changes requested):

packages/sdk/src/server/plugins/graphql.ts
```diff
@@ -28,27 +28,53 @@ export interface GraphQLServerConfig {
...
+const FastifyGraphQLPlugin: FastifyPluginAsync<LazyGraphQLServerConfig> = async (fastify, config) => {
 	const baseContext = await config.baseContext;
+	let schema: GraphQLSchema;
 
 	fastify.route({
 		method: ['GET', 'POST'],
 		url: config.routeUrl,
 		async handler(req, reply) {
+			if (!schema) {
+				schema = await lazyLoadSchema(config.schema);
```
StarpTech, line 55: Just for completeness. This will still introduce cold starts when the user uses this endpoint. The good thing is will not affect others.

packages/sdk/src/rest2graphql/execution.ts
StarpTech: Can't we generate the schema at build time and load it from disk as JSON? I saw some libraries that serialize the schema to and from JSON.

packages/sdk/src/rest2graphql/execution.ts
```diff
@@ -0,0 +1,82 @@
...
+	const fullSpecPath = path.join(process.env.WG_DIR_ABS!, openApiSpecsLocation, specName);
+	const specContent = fs.readFileSync(fullSpecPath).toString();
+	const spec: OpenApiSpec = JSON.parse(specContent);
+
+	let baseUrl: string | undefined;
+	if (spec.baseURL) {
+		baseUrl = resolveConfigurationVariable(mapInputVariable(spec.baseURL));
+	}
+
+	const { schema } = await createGraphQLSchema(spec.content, {
```
StarpTech, line 30: Can't we get rid of creating the schema at runtime by serializing the final schema to JSON or GraphQL at build time?
https://www.apollographql.com/blog/backend/schema-design/three-ways-to-represent-your-graphql-schema/

StarpTech (changes requested):

packages/sdk/src/rest2graphql/execution.ts
StarpTech: Nvm. openapi-to-graphql do much more. It implements the resolver. I forgot it.

packages/sdk/src/server/plugins/graphql.ts
Aenimus: Can't we do this? Or am I missing something?

`ts
const headers = Object.entries(request.headers).reduce((acc, [key, value]) => {
    if (value !== undefined) {
        acc[key] = value.toString();
    }
    return acc;
    }, {} as Record<string, string>);`

spetrunin: @Aenimus yes we could do that

packages/sdk/src/rest2graphql/execution.ts
spetrunin: I need headers which was forwarded by wundernode

spetrunin:
I consider it more or less ready, cleaned up as much as I can

@StarpTech please play with headers, maybe you will get how to simplify it

Important things:
- I need headers coming from node to server (not client request headers)
- at the end, they should be just an object string:string

packages/sdk/src/definition/index.ts
```diff
@@ -433,213 +428,78 @@ export interface GraphQLServerConfiguration extends Omit<GraphQLIntrospection, '
...
+			return introspectSQLServer(introspection);
+		};
+	},
+	mongodb: (introspection: DatabaseIntrospection): ILazyIntrospection<MongoDBApi> => {
+		return async (): Promise<MongoDBApi> => {
+			return introspectMongoDB(introspection);
+		};
+	},
+	prisma: (introspection: PrismaIntrospection): ILazyIntrospection<PrismaApi> => {
+		return async (): Promise<PrismaApi> => {
```
StarpTech, line 485: Remove `async` here it's not needed. It creates another promise.

StarpTech (changes requested):

StarpTech:
> I consider it more or less ready, cleaned up as much as I can

@spetrunin were you able to identify a degradation of the server start in general?

packages/sdk/src/server/plugins/graphql.ts
Pagebakers: Yep the load time is postponed. We have a few solutions in mind to speed it up. But good enough for now.

packages/sdk/src/definition/index.ts
spetrunin: done

spetrunin:
> > I consider it more or less ready, cleaned up as much as I can
> 
> @spetrunin were you able to identify a degradation of the server start in general?

for simple test app:
in main with custom graphql server: `29ms`
in this branch with custom graphql server and stripe api: `49ms`

StarpTech:
This is negligible. Looks good!

packages/sdk/src/server/server.ts
```diff
@@ -222,14 +225,42 @@ export const createServer = async ({
...
+
+				fastify
+					.register(graphqlPlugin as FastifyPluginAsync<LazyGraphQLServerConfig>, {
+						serverName: apiName,
+						schema: () => createExecutableSchema(specPath),
+						routeUrl: routeUrl,
+					})
+					.ready(() => {
+						fastify.log.info('GraphQL plugin registered');
+						fastify.log.info(`OpenAPi Graphql server '${apiName}' listening at ${routeUrl}`);
```
Pagebakers, line 260: OpenAPI, GraphQL

Pagebakers (changes requested):
A lot to digest, found some minor textual changes

Pagebakers (approved):

jensneuse (approved):

PR #626, opened 2023-02-01
feat: when OIDC authentication is refused, redirect back with the error

fiam:
Instead of returning a 400, if we have a redirect_uri go to it but pass
a parameter with the error message. This allows the frontend to react
accordingly.

### DIFF
--- a/pkg/authentication/oidc.go
+++ b/pkg/authentication/oidc.go
@@ -25,6 +25,11 @@ const (
 	OpenIDConnectFlavorAuth0
 )
 
+const (
+	oidcRedirectURICookieName             = "redirect_uri"
+	oidcRedirectOnCompletionURICookieName = "completion_redirect_uri"
+)
+
 type OpenIDConnectCookieHandler struct {
 	log    *zap.Logger
 	claims ClaimsInfo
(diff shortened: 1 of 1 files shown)


YuriBuerov (approved):

pkg/authentication/oidc.go
```diff
@@ -93,209 +98,236 @@ func (h *OpenIDConnectCookieHandler) Register(authorizeRouter, callbackRouter *m
...
+		Endpoint:     provider.Endpoint(),
+		RedirectURL:  redirectURI,
+		Scopes:       []string{oidc.ScopeOpenID, "profile", "email"},
+	}
 
-		cookiePath := fmt.Sprintf("/auth/cookie/callback/%s", config.ProviderID)
-		cookieDomain := sanitizeDomain(r.Host)
+	state, err := generateState()
+	if err != nil {
+		return
```
Aenimus, line 135: err is swallowed?

fiam: Ooops, good catch. I only moved that block around and I didn't realize. Let me take a deeper look.

pkg/authentication/oidc.go
```diff
@@ -93,209 +98,236 @@ func (h *OpenIDConnectCookieHandler) Register(authorizeRouter, callbackRouter *m
...
+	callbackRouter.Path(fmt.Sprintf("/%s", config.ProviderID)).Methods(http.MethodGet).HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		h.authorizationCallback(ctx, w, r, provider, &config, &hooks)
+	})
+}
 
-		redirectOnSuccessURL := r.URL.Query().Get("redirect_uri")
+func (h *OpenIDConnectCookieHandler) authorize(ctx context.Context, w http.ResponseWriter, r *http.Request, provider *oidc.Provider, config *OpenIDConnectConfig) {
+	if provider == nil {
+		http.Error(w, "oidc provider configuration error", http.StatusMethodNotAllowed)
+		return
```
Aenimus, line 112: It's on http but we should we return an err here also?

pkg/authentication/oidc.go
```diff
@@ -93,209 +98,236 @@ func (h *OpenIDConnectCookieHandler) Register(authorizeRouter, callbackRouter *m
...
-			http.Error(w, "oidc provider configuration error", http.StatusMethodNotAllowed)
-			return
-		}
+	state, err := r.Cookie("state")
+	if err != nil {
+		h.log.Error("OIDCCookieHandler state missing",
+			zap.Error(err),
+		)
+		w.WriteHeader(http.StatusBadRequest)
+		return
```
Aenimus, line 201: swallowed err here

pkg/authentication/oidc.go
```diff
@@ -93,209 +98,236 @@ func (h *OpenIDConnectCookieHandler) Register(authorizeRouter, callbackRouter *m
...
-			w.WriteHeader(http.StatusBadRequest)
-			return
-		}
+	redirectURI, err := r.Cookie(oidcRedirectURICookieName)
+	if err != nil {
+		h.log.Error("OIDCCookieHandler redirect uri missing",
+			zap.Error(err),
+		)
+		w.WriteHeader(http.StatusBadRequest)
+		return
```
Aenimus, line 218: swallowed err here

pkg/authentication/oidc.go
```diff
@@ -93,209 +98,236 @@ func (h *OpenIDConnectCookieHandler) Register(authorizeRouter, callbackRouter *m
...
-		if err != nil {
-			h.log.Error("OIDCCookieHandler redirect uri missing",
-				zap.Error(err),
-			)
+	// This is optional
+	var redirectOnCompletionURI string
+	if redirectOnCompletionCookie, err := r.Cookie(oidcRedirectOnCompletionURICookieName); err == nil {
+		redirectOnCompletionURI = redirectOnCompletionCookie.Value
+	}
+
```
Aenimus, line 226: what if err is not nil?

pkg/authentication/oidc.go
```diff
@@ -93,209 +98,236 @@ func (h *OpenIDConnectCookieHandler) Register(authorizeRouter, callbackRouter *m
...
+	oauth2Config := &oauth2.Config{
+		ClientID:     config.ClientID,
+		ClientSecret: config.ClientSecret,
+		Endpoint:     provider.Endpoint(),
+		RedirectURL:  redirectURI.Value,
+		Scopes:       []string{oidc.ScopeOpenID, "profile", "email"},
+	}
+	if err := h.exchangeToken(ctx, w, r, oauth2Config, provider, config, hooks); err != nil {
+		// Error is already logged by exchangeToken
```
Aenimus, line 235: I prefer happy path first... err is nil, write the header and return

pkg/authentication/oidc.go
fiam: That's not swallowed. We print an error to the log and we also return a 400.

pkg/authentication/oidc.go
fiam: Then we take the fallback path later in the file, used when `redirectOnCompletionURI` is not available

pkg/authentication/oidc.go
```diff
@@ -93,209 +98,236 @@ func (h *OpenIDConnectCookieHandler) Register(authorizeRouter, callbackRouter *m
...
-			)
-			w.WriteHeader(http.StatusBadRequest)
-			return
-		}
+	if r.URL.Query().Get("state") != state.Value {
+		h.log.Error("OIDCCookieHandler state mismatch",
+			zap.Error(err),
+		)
+		w.WriteHeader(http.StatusBadRequest)
+		return
```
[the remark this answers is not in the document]
fiam: We log it and we return 400, it should be good enough

pkg/authentication/oidc.go
Aenimus: Fair point, but why would we only sometimes return the err?

pkg/authentication/oidc.go
fiam: In Go the pattern `if something, err := function(); err != nil { handle error }` is more common

Aenimus: fair enough.

pkg/authentication/oidc.go
```diff
@@ -93,209 +98,236 @@ func (h *OpenIDConnectCookieHandler) Register(authorizeRouter, callbackRouter *m
...
+				qs.Add("error", url.QueryEscape(err.Error()))
+				redirURL.RawQuery = qs.Encode()
+				redirectOnCompletionURI = redirURL.String()
+			}
+			http.Redirect(w, r, redirectOnCompletionURI, http.StatusFound)
+		} else {
 			w.WriteHeader(http.StatusBadRequest)
-			return
 		}
+		return
```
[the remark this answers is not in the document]
Aenimus: Sure. What strikes me as weird is that we can get errors, and sometimes they're handled one way, and other times they're returned...

pkg/authentication/oidc.go
fiam: That error should never happen, I'm thinking of just removing the check (notice there's an earlier check for `provider == nil` in `Register()`

pkg/authentication/oidc.go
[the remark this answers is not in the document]
fiam: The idea returning `err` in `h.exchangeToken()` is that we need to do something with any error that comes from it because it can be caused by a user action and we might want to show a message, so we need to carry it around.

For errors that are either configuration errors or either non recoverable, we log them on the spot we get them because we have more context to display in the log message.

pkg/authentication/oidc.go
fiam: I've removed the entire check, thanks for pointing it out, otherwise I wouldn't realised that it wasn't needed.

pkg/authentication/oidc.go
fiam: Fixed

pkg/authentication/oidc.go
fiam: Check this, please https://github.com/wundergraph/wundergraph/pull/626#discussion_r1093374275

Aenimus (approved):

StarpTech (approved):
LGTM

PR #627, opened 2023-02-01
chore: examples, fix issues with origin

YuriBuerov:
The issue is with the origin. We use these examples as templates, this fix makes them stable.


fiam (approved):
LGTM

Aenimus (approved):

PR #632, opened 2023-02-02
feat: add support for uploads from anonymous users

fiam:
- Add a new per-profile field, which defaults to requiring authentication
- Add end to end tests to nodejs-file-upload example, which also validates
that the upload hooks run.


Pagebakers (changes requested):
I think the parameter name we use with functions is `requiresAuthentication` let's make sure we use the same (I couldn't find it in the docs btw, please verify)

I've checked the source code, that's the parameter name. I'll update it.

fiam:
CI should pass once this is rebased on top of #631

fiam:
> I think the parameter name we use with functions is `requiresAuthentication` let's make sure we use the same (I couldn't find it in the docs btw, please verify)
> 
> I've checked the source code, that's the parameter name. I'll update it.

Changed, PTAL

YuriBuerov (approved):

pkg/s3uploadclient/s3uploadclient.go
```diff
@@ -50,6 +50,8 @@ type preparedProfile struct {
 // UploadProfile specifies options like maximum file size and allowed
 // extensions for a given upload profile
 type UploadProfile struct {
+	// Wether the profile requires authentication to upload a file
```
StarpTech, line 53: Typo

pkg/s3uploadclient/s3uploadclient.go
```diff
@@ -408,6 +415,18 @@ func (s *S3UploadClient) handlePart(ctx context.Context, r *http.Request, part *
 	return info, nil
 }
 
+func (s *S3UploadClient) hasRequiredAuthentication(w http.ResponseWriter, r *http.Request) bool {
+	// Don't check for a valid profile name here. Since the default is requiring users
+	// to be authenticated, we can just check wether there's a profile. If the name
+	// doesn't exist we'll get an error later.
+	_, profile, _ := s.uploadProfile(r)
+	if (profile == nil || profile.RequiresAuthentication) && authentication.UserFromContext(r.Context()) == nil {
```
StarpTech, line 423: That contains many error cases. I prefer to handle each of them individually with some logs.

StarpTech (changes requested):

pkg/s3uploadclient/s3uploadclient.go
fiam: Thanks, fixed

pkg/s3uploadclient/s3uploadclient.go
fiam: I've separated it into two checks and added a log message. PTAL

pkg/s3uploadclient/s3uploadclient.go
```diff
@@ -408,6 +417,23 @@ func (s *S3UploadClient) handlePart(ctx context.Context, r *http.Request, part *
...
 
+func (s *S3UploadClient) hasRequiredAuthentication(w http.ResponseWriter, r *http.Request) bool {
+	// Don't check for a valid profile name here. Since the default is requiring users
+	// to be authenticated, we can just check if there's a profile. If the name
+	// doesn't exist we'll get an error later.
+	profileName, profile, _ := s.uploadProfile(r)
+	if profile == nil || profile.RequiresAuthentication {
+		if authentication.UserFromContext(r.Context()) == nil {
+			if s.logger != nil {
+				s.logger.Info("refusing upload from anonymous user", zap.String("provider", s.name), zap.String("profile", profileName))
```
StarpTech, line 428: I'd prefer debug.

StarpTech (changes requested):

pkg/s3uploadclient/s3uploadclient.go
fiam: Changed to Debug

Pagebakers (approved):

StarpTech (approved):
LGTM

PR #636, opened 2023-02-02
fix: avoid feeding paths into applicationHash calculation

fiam:
Store ResolveApplicationConfig in a field in CodeGenerationConfig
rather than extending it. This allows us to safely hash ResolveApplicationConfig
without including the paths in CodeGenerationConfig.

### DIFF
--- a/packages/golang-client/src/index.ts
+++ b/packages/golang-client/src/index.ts
@@ -1,4 +1,4 @@
-import { BaseTypeScriptDataModel, ResolvedWunderGraphConfig, Template, TemplateOutputFile } from '@wundergraph/sdk';
+import { BaseTypeScriptDataModel, Template, TemplateOutputFile } from '@wundergraph/sdk';
 import { hasInput, visitJSONSchema } from '@wundergraph/sdk/internal';
 import { JSONSchema7 as JSONSchema, JSONSchema7 } from 'json-schema';
 import execa from 'execa';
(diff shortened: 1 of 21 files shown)


packages/golang-client/src/index.ts
```diff
@@ -7,6 +7,7 @@ import Handlebars from 'handlebars';
 import { clientTemplate } from './client-template';
 import { OperationType } from '@wundergraph/protobuf';
 import Logger from '@wundergraph/sdk/internal/logger';
+import { CodeGenerationConfig } from '@wundergraph/sdk/dist/configure';
```
StarpTech, line 10: Why point to dist? if is not exported, please do it.

packages/sdk/src/configure/index.ts
```diff
@@ -259,7 +259,8 @@ export interface ResolvedWunderGraphConfig {
 	serverOptions?: ResolvedServerOptions;
 }
 
-export interface CodeGenerationConfig extends ResolvedWunderGraphConfig {
```
StarpTech, line 262: Could you explain why it makes a difference for hashing?

StarpTech (changes requested):

packages/golang-client/src/index.ts
Pagebakers: Yep /dist was for backwards compat, we should remove it at some point

fiam: It was not exported. The editor filled the import automatically and I didn't think about it. I'll update it.

packages/sdk/src/configure/index.ts
fiam: I've added comments in `CodeGenerationConfig` and `ResolvedWunderGraphConfig`, PTAL

StarpTech: ah now, I got it. Great fix

StarpTech (approved):
LGTM

thisisnithin (approved):

PR #648, opened 2023-02-06
feat: add support for custom claims (#648

fiam:
- Parse custom claims when authenticating users
- Allow defining custom claims on a per-app basis, including their type
- Declare a GraphQL enum for the available custom claims
- Add @fromCustomClaim directive, which injects as claim into a variable


fiam:
Docs are missing, but code should be ready for review

packages/sdk/src/definition/merge.ts
```diff
@@ -84,10 +84,25 @@ const mergeTypeConfigurations = <T extends {} = {}>(apis: Api<T>[]): TypeConfigu
...
+"""
+The @fromClaim directive sets the variable to the value retrieved from the given well-known Claim.
+Adding this directive makes the operation require authentication. For custom claims, see @fromCustomClaim.
+"""
+
 directive @fromClaim(
   name: Claim
 ) on VARIABLE_DEFINITION
 
+enum Claim {
```
jensneuse, line 96: I'd rename this to WG_WELL_KNOWN_CLAIM

packages/sdk/src/definition/merge.ts
```diff
@@ -84,10 +84,25 @@ const mergeTypeConfigurations = <T extends {} = {}>(apis: Api<T>[]): TypeConfigu
...
+	USERID
+	EMAIL
+	EMAIL_VERIFIED
+	NAME
+	NICKNAME
+	LOCATION
+	PROVIDER
+}
```
jensneuse, line 104: As you're touching this already, do you like to add the missing claims?

pkg/apihandler/apihandler.go
```diff
@@ -998,7 +1005,64 @@ func injectClaims(operation *wgpb.Operation, r *http.Request, variables []byte)
...
+	for _, claim := range customClaims {
+		value, valueType, _, _ := jsonparser.Get(customUserClaimsData, claim.Claim.JsonPathComponents...)
+		if claim.Claim.Required && (valueType == jsonparser.NotExist || valueType == jsonparser.Null) {
+			return nil, &inputvariables.ValidationError{
+				Message: fmt.Sprintf("required customClaim %s not found", claim.Claim.Name),
+			}
+		}
+		var replacement []byte
+		switch valueType {
+		case jsonparser.String:
```
jensneuse, line 1021: can we move the mapper somewhere else to keep the for loop readable?

jensneuse (changes requested):
only minor change requests
fantastic work!

packages/sdk/src/definition/merge.ts
fiam: Sure, will do

packages/sdk/src/definition/merge.ts
fiam:
```enum WellKnownClaim {
	EMAIL = 0; // email
	EMAIL_VERIFIED = 1; // emailVerified
	NAME = 2; // name
	NICKNAME = 3; // nickName
	LOCATION = 4; // location
	PROVIDER = 5; // providerId - notice this is usually set by us, not from claims
	USERID = 6; // userId
	FIRST_NAME = 7; // firstName
	LAST_NAME = 8; // lastName 
	DESCRIPTION = 9; // description
	AVATAR_URL = 10; // avatarUrl ```

Anything else that comes to mind?

File: packages/sdk/src/definition/merge.ts (hunk 1 above)
Comment (fiam) [reply packages/sdk/src/definition/merge.ts]: Done

File: packages/sdk/src/definition/merge.ts (hunk 2 above)
Comment (fiam) [reply packages/sdk/src/definition/merge.ts:176]: Done

File: docs-website/src/pages/docs/directives-reference/from-custom-claim-directive.md [hunk 4]
```diff
@@ -0,0 +1,65 @@
...
+```ts
+// configureWunderGraph emits the configuration
+configureWunderGraphApplication({
+  ...
+	authentication: {
+		customClaims: {
+      // Implicit: required
+			SHOPID: {
+				jsonPath: 'shop.id', // Nested 'id' field inside a 'shop' object
+				type: ValueType.INT, // Must be an integer
```
StarpTech, line 25: Can we adapt here the same zod schema as in function, or is it too much effort?

pkg/apihandler/apihandler.go
```diff
@@ -962,43 +962,174 @@ func (h *GraphQLHandler) preparePlan(operationHash uint64, requestOperationName
...
+			return item
+		}
+		return lookupJsonPath(item, keys, keyIndex+1)
+	}
+	return nil
+}
+
+func injectCustomClaims(customClaims []*wgpb.CustomClaimConfig, user *authentication.User, variables []byte) ([]byte, error) {
+	for _, claim := range customClaims {
+		value := lookupJsonPath(user.CustomClaims, claim.Claim.JsonPathComponents, 0)
```
StarpTech, line 1054: You can use `gjson` https://github.com/tidwall/gjson#path-syntax which we already use.

testapps/auth/test/index.test.ts
```diff
@@ -0,0 +1,262 @@
+import { afterAll, beforeAll, describe, expect, it, test } from '@jest/globals';
```
StarpTech, line 1: Great tests but a bit odd to have them in the testapps we will see as a toy playground. I think we also don't run them on go code change. It would be great if we were able to write similar tests in the go codebase.

docs-website/src/pages/docs/directives-reference/from-custom-claim-directive.md
StarpTech, line 25: - Can we use string types instead enums?
- Should we summarize Float, Int to `Number` ?
- Can a claim be optional? E.g. in Auth0 you add custom claims conditionally.

StarpTech (changes requested):
Awesome work!

testapps/auth/test/index.test.ts
fiam: There's no other way to write a comprehensive test that validates both the TS and Go sides, as well as their interaction. We can add unit tests for the each side, but they're more code to maintain and the integration tests make the unit test almost redundant.

pkg/apihandler/apihandler.go
fiam: That's what the first iteration of the code did, but using `jsonparser` (which we also use). The problem with that is we would need to turn the custom claims into JSON on every call to `injectCustomClaims()`, which allocates. This way we avoid allocations at the cost of implementing the path lookup ourselves.

fiam: One way I considered to get around the allocation and still use a JSON parser for the lookup is to keep the custom claims as a `json.RawMessage`, but the API around that is such a pain in the Go side that I decided to not go for it.

docs-website/src/pages/docs/directives-reference/from-custom-claim-directive.md
fiam: Zod is a bit too much for this because we don't support objects, just plain values. I think the confusion it could introduce is not worth the familiarity it brings.

docs-website/config/tags.js
```diff
@@ -107,6 +107,7 @@ const tags = {
 	'@directive': '/docs/directives-reference',
 	'@export': '/docs/directives-reference/export-directive',
 	'@fromClaim': '/docs/directives-reference/from-claim-directive',
+	'@fromCustomClaim': '/docs/directives-reference/from-custom-claim-directive',
```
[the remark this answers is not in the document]
fiam: We can handle it as a single directive, but it was requested as a separate one. We're still in time to change that, let's discuss it and decide.

docs-website/src/pages/docs/directives-reference/from-custom-claim-directive.md
fiam: Sorry, what do you mean by a string type? `"string|int|float"`? We could do that and make the enum internal for communicating between the TS and Go sides. I wouldn't group together integer and float as number, otherwise it becomes quite the chore to validate it in every operation. As for validating for a specific pattern, we can add support for matching it to a regular expression. Claims are optional using `required: false`, it's documented in the last paragraph of the document.

StarpTech: ok

pkg/apihandler/apihandler.go
StarpTech: ok, but please document the syntax in the docs. Also whats, not possible e.g arrays etc.

testapps/auth/test/index.test.ts
StarpTech: How do we ensure that the tests are running on code change?

docs-website/src/pages/docs/directives-reference/from-custom-claim-directive.md
StarpTech: > Sorry, what do you mean by a string type? "string|int|float"? We could do that and make the enum internal for communicating between the TS and Go sides. I wouldn't group together integer and float as number, ....

Looks good. Can we support String, Int, Float, and Boolean? Then we are compliant with GraphQL base types.

> Claims are optional using required: false, it's documented in the last paragraph of the document.

Great.

docs-website/config/tags.js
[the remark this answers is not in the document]
StarpTech: I'm in favor of covering it with a single directive.

testapps/auth/test/index.test.ts
fiam: The NPM CI workflow runs `pnpm test` in these directories and they have a `package.json` that runs `jest`. Any PR that changes the SDK will have to pass these tests.

pkg/apihandler/apihandler.go
fiam: Good idea, I'll document it.

docs-website/src/pages/docs/directives-reference/from-custom-claim-directive.md
StarpTech: Thanks

jensneuse (approved):
fantastic work!

PR #661, opened 2023-02-10
fix: internal client should use correct mappings for nested operations

### LINKED ISSUE #645: Internal operation API paths are no longer correct in SDK
### Bug description

#595 updated the `apiPath` of internal operations in the WunderGraph server (see [here](https://github.com/wundergraph/wundergraph/blob/f2e1ec10f4ab89f6caf4cc713eab18c345aeaddf/pkg/apihandler/apihandler.go#L447)). E.g. if we define a query in `.wundergraph/operations/images/GetImages.graphql`, then the path is now `http://localhost:9991/internal/operations/images/GetImages` instead of `http://localhost:9991/internal/operations/ImagesGetImages`).

However, the SDK is still using the old path (see [here](https://github.com/wundergraph/wundergraph/blob/f2e1ec10f4ab89f6caf4cc713eab18c345aeaddf/packages/sdk/src/server/internal-client.ts#L96)). This means that all calls to internal operations in subdirectories of `operations` now return a 404 error.

### Expected behavior

_No response_

### Environment & setup

- OS: 
- Go version: 
- Database: 
- Node.js version: 

### WunderCtl Version

```

```

jensneuse:
fixes #645


thisisnithin (approved):

fiam:
Can we add a test to make sure the operation works? I can add it for you if you want me to.

spetrunin (approved):

PR #662, opened 2023-02-10
chore: update graphql-go-tools

### LINKED ISSUE #634: Undefined optional parameters in POST operation are passed to downstream operation as nulls.
### Bug description

Related to #114, but the opposite direction...

I had the below operation defined that I'm pretty sure was working when I wrote it (July 2022) but now it is not. The reason is that any parameters that are not passed to the operation (such as if I omitted "description") are passed through to the downstream operation with the value null.

This breaks the semantic that a null value clears a value, but not sending a value leaves it unchanged.

The current behaviour of Wundergraph changes the meaning of the request between client and downstream. Undefined means I have nothing to say about it, null means I definitely don't want a value.

```
# Whichever arguments to the mutation are missing in the body will not be used.
# Any arguments set to null will clear an existing value from the system object.
mutation updateObject($id: Int!, $name: String, $description: String, $icon: URL) {
    object: downstreamUpdateObject(id: $id, name: $name, description: $description, icon: $icon)
    {
        id
    }
}
```
Log Output
```

--- ClientRequest start ---

POST /operations/UpdateObject HTTP/1.1
Host: localhost:9991
Accept: */*
Content-Length: 25
Content-Type: application/json
User-Agent: curl/7.68.0

{"id": 0, "name": "root"}



--- ClientRequest end ---



--- DebugTransport ---

Request:

POST / HTTP/1.1
Host: localhost:8000
Accept: application/json
Content-Type: application/json
X-Request-Id: 3c5a8ca8-cab3-7d7b-513f-28fdc058694e

{"variables":{"description":null,"icon":null,"id":0,"name":"root"},"query":"mutation($description: String, $icon: URL, $id: Int!, $name: String){object: updateObject(description: $description, icon: $icon, id: $id, name: $name){id}}"}

Duration: 38 ms

Response:

HTTP/1.1 200 OK
Content-Length: 28
Content-Type: application/json
Date: Thu, 02 Feb 2023 11:47:52 GMT
Server: uvicorn

{"data":{"object":{"id":0}}}

--- DebugTransport
```

### How to reproduce

- Create a mutation operation on Wundergraph with multiple optional parameters that is a simple forwarding operation to an downstream graphql api with the same signature.
- call the Wundergraph operation with some parameters undefined
- the downstream API will receive the call with the undefined parameters defined with the value 'null'

### Expected behavior

When a Wundergraph operation with optional parameters is called, and one or more of the optional parameters is excluded from the call, then the call that Wundergraph makes to the downstream operation should not have the optional parameters included. They should be absent from the call to the downstream operation.

### WunderGraph information

```
"dependencies": {
    "@types/node": "^14.14.37",
    "@wundergraph/sdk": "^0.133.0",
    "graphql": "^16.3.0",
    "node-fetch": "2",
    "tslint": "^6.1.3",
    "typescript": "^4.7.4"
  },
  "devDependencies": {
    "@types/json-schema": "^7.0.11",
    "@types/node-fetch": "^2.6.1",
    "@wundergraph/wunderctl": "^0.128.0"
  },
```

### Environment & setup

- OS: Mint 20.3
- Go version: 1.13.8
- Node.js version: v18.12.1

### WunderCtl Version

```
Version: 0.128.0
Commit: f136eb833ed444af70a3dd0f6e5d5b4d1b959335
Date: 2023-01-30T12:55:12Z
BuiltBy: ci
```

spetrunin:
fixes #634 

#### Checklist

- [ ] run `make test`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -46,7 +46,7 @@ require (
 	github.com/tidwall/gjson v1.11.0
 	github.com/tidwall/sjson v1.1.5
 	github.com/valyala/fasthttp v1.44.0
-	github.com/wundergraph/graphql-go-tools v1.61.0
+	github.com/wundergraph/graphql-go-tools v1.61.1
 	go.uber.org/zap v1.18.1
 	golang.org/x/net v0.5.0
 	golang.org/x/oauth2 v0.0.0-20211104180415-d3ed0bb246c8
@@ -72,7 +72,7 @@ require (
 	github.com/go-playground/validator/v10 v10.11.2 // indirect
 	github.com/goccy/go-json v0.10.0 // indirect
 	github.com/golang/protobuf v1.5.2 // indirect
-	github.com/google/go-cmp v0.5.6 // indirect
+	github.com/google/go-cmp v0.5.8 // indirect
 	github.com/google/go-querystring v1.0.0 // indirect
 	github.com/google/uuid v1.1.1 // indirect
 	github.com/hashicorp/errwrap v1.0.0 // indirect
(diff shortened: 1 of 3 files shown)


jensneuse (approved):

fiam (approved):
LGTM

PR #666, opened 2023-02-11
feat: store generated authentication secrets in secrets manager on deploy

Aenimus:
#### Checklist

- [ ] run `make test`


pkg/apihandler/api.go
```diff
@@ -1,13 +1,23 @@
...
 	"go.uber.org/zap/zapcore"
 
 	"github.com/wundergraph/wundergraph/pkg/wgpb"
 )
 
+const (
+	WgEnvCsrfSecret       = "WUNDERGRAPH_CSRF_TOKEN_SECRET"
+	WgEnvHashKey          = "WUNDERGRAPH_SECURE_COOKIE_HASH_KEY"
+	WgEnvBlockKey         = "WUNDERGRAPH_SECURE_COOKIE_BLOCK_KEY"
+	validSecretCharacters = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890"
```
Aenimus, line 18: Are other characters valid?

pkg/apihandler/api.go
```diff
@@ -41,6 +113,7 @@ type Api struct {
 	S3UploadConfiguration []*wgpb.S3UploadConfiguration
 	Webhooks              []*wgpb.WebhookConfiguration
 	Options               *Options
+	CookieBasedSecrets    *CookieBasedSecrets
```
Aenimus, line 116: Should I move CookieBasedSecrets into AuthenticationConfig > CookieBased? Or leave it as a top level field?

packages/sdk/src/configure/index.ts
```diff
@@ -253,10 +253,13 @@ export interface ResolvedWunderGraphConfig {
 			revalidateAuthentication: boolean;
 			postLogout: boolean;
 		};
-		cookieSecurity: {
-			secureCookieHashKey: ConfigurationVariable;
-			secureCookieBlockKey: ConfigurationVariable;
-			csrfTokenSecret: ConfigurationVariable;
+		cookieSecurity?: {
+			/** @deprecated: Not used anymore */
```
jensneuse, line 280: Add links to docs for replacement/alternative?

packages/sdk/src/configure/index.ts
```diff
@@ -253,10 +253,13 @@ export interface ResolvedWunderGraphConfig {
...
+			secureCookieHashKey?: any;
+			/** @deprecated: Not used anymore */
```
jensneuse, line 282: same

packages/sdk/src/configure/index.ts
```diff
@@ -253,10 +253,13 @@ export interface ResolvedWunderGraphConfig {
...
+			secureCookieBlockKey?: any;
+			/** @deprecated: Not used anymore */
```
jensneuse, line 284: same

pkg/apihandler/api.go
jensneuse: I'm not sure. I think anything is valid?

jensneuse (changes requested):
A few minor issues.
One thing I was missing is that my understanding was that in dev mode (wunderctl up) we always want to set the same static value. Am I missing this?

pkg/apihandler/api.go
Aenimus: Let me rephrase. What would be reasonable to include?

pkg/apihandler/api.go
```diff
@@ -25,6 +35,68 @@ type Options struct {
...
+	case WgEnvCsrfSecret:
+		cookieBasedSecrets.CsrfSecret = secretValue
+	case WgEnvBlockKey:
+		cookieBasedSecrets.BlockKey = secretValue
+	case WgEnvHashKey:
+		cookieBasedSecrets.HashKey = secretValue
+	}
+}
+
+func generateRandomStringOfLength(length int) []byte {
```
StarpTech, line 66: `rand.Intn` should not be used for security-senstive stuff. 

> This package's outputs might be easily predictable regardless of how it's seeded. For random numbers suitable for security-sensitive work, see the crypto/rand package.
https://pkg.go.dev/math/rand

Use this

```go
// GenerateRandomBytes returns securely generated random bytes.
// It will return an error if the system's secure random
// number generator fails to function correctly, in which
// case the caller should not continue.
func GenerateRandomBytes(n int) ([]byte, error) {
	b := make([]byte, n)
	_, err := rand.Read(b)
	// Note that err == nil only if we read len(b) bytes.
	if err != nil {
		return nil, err
	}

	return b, nil
}

// GenerateRandomString generated a random string with the length as the first argument
func GenerateRandomString(n int) (string, error) {
	b, err := GenerateRandomBytes(n)
	return base64.URLEncoding.EncodeToString(b), err
}
```

packages/protobuf/wgpb/wundernode_config.ts
```diff
@@ -771,9 +771,6 @@ export interface CookieBasedAuthentication {
   providers: AuthProvider[];
   authorizedRedirectUris: ConfigurationVariable[];
   authorizedRedirectUriRegexes: ConfigurationVariable[];
-  hashKey: ConfigurationVariable | undefined;
```
StarpTech, line 774: Why do we remove this? IMO the user should still be able to configure it. The only difference is that we generate the value. This will also be a breaking change.

StarpTech (changes requested):

pkg/apihandler/api.go
```diff
@@ -25,6 +35,68 @@ type Options struct {
...
+	bytes := make([]byte, length)
+	for i := range bytes {
+		bytes[i] = validSecretCharacters[rand.Intn(len(validSecretCharacters))]
```
StarpTech, line 69: `rand.Intn` should not be used for security-sensitive work. Please use crypto rand e.g.

```go
func GenerateRandomString(n int) (string, error) {
	const letters = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz-"
	ret := make([]byte, n)
	for i := 0; i < n; i++ {
		num, err := rand.Int(rand.Reader, big.NewInt(int64(len(letters))))
		if err != nil {
			return "", err
		}
		ret[i] = letters[num.Int64()]
	}

	return string(ret), nil
}
```

StarpTech (changes requested):

packages/protobuf/wgpb/wundernode_config.ts
Aenimus: My mistake. I didn't intend to remove from interfaces.

Aenimus: This file is generated from proto, and I removed the proto values, so that's why they're gone here.

Is this definitely a breaking change?

jensneuse (approved):

StarpTech (changes requested):

docs-website/src/pages/docs/self-hosted/security.md
```diff
@@ -0,0 +1,53 @@
+---
+title: 'Security'
+pageTitle: WunderGraph - Security
+description: Securing your self-hosted WunderGraph deployment
+---
+
+WunderGraph uses three secrets to secure your deployment.
```
StarpTech, line 7: The secrets are only about authentication, nothing more.

docs-website/src/pages/docs/self-hosted/security.md
```diff
@@ -0,0 +1,53 @@
...
+Cryptographically-secure (pseudo-)randomly generated strings should be added to your `.env` file:
+
+```js
+// these strings are insecure examples with the intention to display length
+WUNDERGRAPH_CSRF_TOKEN_SECRET = aaaaaaaaaaa; // 11 bytes
+WUNDERGRAPH_SECURE_COOKIE_HASH_KEY = aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa; // 32 bytes
+WUNDERGRAPH_SECURE_COOKIE_BLOCK_KEY = aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa; // 32 bytes
+```
+
+Although values will be generated for you in the case that they have not been supplied, these secrets are your
```
StarpTech, line 17: We should mention that in case of you don't use fix values, we regenerate them and you users will be logged out after every restart

StarpTech (changes requested):

pkg/apihandler/api.go
```diff
@@ -25,6 +36,80 @@ type Options struct {
...
+			return nil, err
+		}
+		bytes[i] = validSecretCharacters[num.Int64()]
+	}
+
+	return bytes, nil
+}
+
+func NewCookieBasedSecrets(isDevMode bool) (cookieBasedSecrets *CookieBasedSecrets, errorMessages []string) {
+	if isDevMode {
```
StarpTech, line 83: I'd handle this condition outside of this method, so the only responsibility of this method is to generate "new" ones.

pkg/apihandler/apihandler.go
```diff
@@ -2120,18 +2120,24 @@ func (r *Builder) registerAuth(insecureCookies bool) error {
 
 	if h := loadvariable.String(r.api.AuthenticationConfig.CookieBased.HashKey); h != "" {
 		hashKey = []byte(h)
+	} else if fallBackValue := r.api.CookieBasedSecrets.HashKey; fallBackValue != nil {
```
StarpTech, line 2123: I'll try :smile: in IT it is called `fallback`

StarpTech (changes requested):

pkg/validate/apiconfig_test.go
```diff
@@ -44,142 +53,278 @@ func TestValidateApiConfig(t *testing.T) {
...
+				},
+			},
+			CookieBasedSecrets: &apihandler.CookieBasedSecrets{
+				CsrfSecret: elevenBytes,
+				BlockKey:   elevenBytes,
+			},
+		})
+
+		assert.False(t, valid)
+		assert.Equal(t, "the required environment variable WUNDERGRAPH_SECURE_COOKIE_BLOCK_KEY must have a length of 32", messages[0])
```
StarpTech, line 293: Let's go with `WG_*` for consistency.

pkg/validate/apiconfig.go
```diff
@@ -1,32 +1,51 @@
...
 
+// ApiConfig validates the Api
+func ApiConfig(api *apihandler.Api) (bool, []string) {
 	if !api.HasCookieAuthEnabled() {
 		// skip following tests if cookie auth is not enabled
 		// in this case, we don't have to check it
 		return true, nil
 	}
 
+	requiredCsrfLength := 11
```
StarpTech, line 22: Where did you implement the logic for the cloud?

StarpTech (changes requested):

pkg/validate/apiconfig.go
StarpTech: Clarified

StarpTech (approved):
LGTM

PR #671, opened 2023-02-14
fix: fix docs nev/prev navigation

JivusAyrus:
#### Checklist

- [ ] run `make test`

image


docs-website/src/components/Layout.jsx
```diff
@@ -120,7 +120,15 @@ function useTableOfContents(tableOfContents) {
 
 export function Layout({ children, title, tableOfContents, frontmatter }) {
 	let router = useRouter();
-	let allLinks = navigation.flatMap((section) => section.links);
+	let links = navigation.flatMap((section) => section.links);
```
Aenimus, line 123: Why do we use `let` instead of `const`. Where are we reassigning?

docs-website/src/components/Layout.jsx
```diff
@@ -120,7 +120,15 @@ function useTableOfContents(tableOfContents) {
...
+	let allLinks = links.flatMap((section) => {
```
Aenimus, line 124: All on one line without curly braces if we use a ternary? :D

docs-website/src/components/Layout.jsx
Aenimus: For all these cases, I don't see why we would not use `const`

docs-website/src/components/Layout.jsx
Aenimus:
```ts
const allLinks = links.flatMap((section) => section?.links ?? section
```

JivusAyrus: Got it

JivusAyrus: Done

docs-website/src/components/Layout.jsx
```diff
@@ -119,13 +119,15 @@ function useTableOfContents(tableOfContents) {
...
-	let router = useRouter();
-	let allLinks = navigation.flatMap((section) => section.links);
-	let linkIndex = allLinks.findIndex((link) => link?.href === router.pathname);
-	let previousPage = allLinks[linkIndex - 1];
-	let nextPage = allLinks[linkIndex + 1];
-	let section = navigation.find((section) => section.links?.find((link) => link?.href === router.pathname));
-	let currentSection = useTableOfContents(tableOfContents);
+	const router = useRouter();
+	const links = navigation.flatMap((section) => section.links);
+	const allLinks = links.flatMap((section) => (section?.links ? section?.links : section));
```
Aenimus, line 124: nitpick, but the inner parentheses are unnecessary, and the second `sections?` doesn't need `?` because it cannot be null at this point

Aenimus (approved):

StarpTech (approved):
LGTM

docs-website/src/components/Layout.jsx
JivusAyrus: Prettier adds the paranthesis

docs-website/src/components/Layout.jsx
```diff
@@ -119,13 +119,15 @@ function useTableOfContents(tableOfContents) {
...
 export function Layout({ children, title, tableOfContents, frontmatter }) {
-	let router = useRouter();
-	let allLinks = navigation.flatMap((section) => section.links);
-	let linkIndex = allLinks.findIndex((link) => link?.href === router.pathname);
-	let previousPage = allLinks[linkIndex - 1];
-	let nextPage = allLinks[linkIndex + 1];
-	let section = navigation.find((section) => section.links?.find((link) => link?.href === router.pathname));
-	let currentSection = useTableOfContents(tableOfContents);
+	const router = useRouter();
+	const links = navigation.flatMap((section) => section.links);
```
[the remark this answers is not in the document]
JivusAyrus: on the root level, every object has links, so shouldn't be a problem

docs-website/src/components/Layout.jsx
JivusAyrus: Removed the "?"

PR #673, opened 2023-02-14
feat: add solid-js client

Pagebakers:
#### Checklist

- [ ] run `make test`


docs-website/src/pages/docs/clients-reference/solid-query.md
```diff
@@ -0,0 +1,156 @@
...
+export const { createQuery, createMutation, createSubscription, createFileUpload, useUser, useAuth, queryKey } =
+  createHooks<Operations>(client);
+```
+
+In your `App.tsx` add QueryClientProvider:
+
+```tsx
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+
+const queryClient = new QueryClient();
```
Aenimus, line 51: They could run into the same problems as we did if they don't use useQueryClient(). For instance, the queries will not invalidate.

docs-website/src/pages/docs/clients-reference/solid-query.md
```diff
@@ -0,0 +1,156 @@
...
+
+```ts
+const user = useUser();
+
+user.data;
+```
+
+### queryKey
+
+You can use the `queryKey` helper function to create a unique key for the query in a typesafe way. This is useful if you want to invalidate the query after mutating.
```
Aenimus, line 139: can't use `new QueryClient` for this

Aenimus (approved):

examples/vite-solidjs/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "dev:vite": "vite",
+    "build:vite": "vite build",
+    "build:wundergraph": "wunderctl generate",
+    "build": "npm run build:wundergraph && npm run build:vite",
+    "preview": "vite preview",
+    "test": "jest"
+  },
+  "license": "MIT",
+  "devDependencies": {
+    "autoprefixer": "^10.4.13",
```
fiam, line 20: We need `jest`, `@jest/globals` and `ts-jest` in `devDependencies`, otherwise the tests won't run.

packages/solid-query/CHANGELOG.md
```diff
@@ -0,0 +1,4 @@
+# Change Log
```
fiam, line 1: Do we need this file here?

packages/solid-query/README.md
```diff
@@ -0,0 +1,157 @@
+# WunderGraph React Query Integration
```
fiam, line 1: This file should be about Solid Query, not React Query, right?

fiam (changes requested):
Great job. I've just pointed a couple of things in the code. We also need to add the new package to `lerna.json` at the repo root, so it will be picked up by the release action.

packages/solid-query/CHANGELOG.md
Pagebakers: Yeah it's auto generated, the changelog will be appended here.

packages/solid-query/README.md
Pagebakers: Whoops indeed

examples/vite-solidjs/package.json
Pagebakers: Added

Pagebakers: Ok this works now.

docs-website/src/pages/docs/clients-reference/solid-query.md
Pagebakers: No you actually need to initialize it and then pass it to the context provider. So this is correct. Inside components you use useQueryClient()

fiam (approved):
LGTM, not sure about the CI error 🤔

PR #678, opened 2023-02-15
feat: add support for injecting variables into nested paths

fiam:
Add an optional `on:` argument to:

- `@fromClaim`
- `@injectGeneratedUUID`
- `@injectCurrentDateTime`
- `@injectEnvironmentVariable`
- `@jsonSchema`

This new argument allows specifying a path inside the object to set the value. Update these directives to make them
repeatable and add tests as well as update docs to reflect these changes.


fiam:
Pending docs update

docs-website/src/pages/docs/directives-reference/from-claim-directive.md
```diff
@@ -111,4 +111,26 @@ mutation ($shopID: Int! @fromClaim(name: SHOPID), $productID: Int!) {
...
+We can use `@fromClaim` to set the value of `code` using:
+
+```graphql
+mutation ($filter: countries_CountryFilterInput! @fromClaim(name: COUNTRY_CODE, on: "code")) {
+  countries_Countries(filter: $filter) {
+    id
+    name
+  }
+}
+```
```
Pagebakers, line 136: Will input.filter.code be available in the generated client? or is it removed just like regular claims?

What happens when I use a non existing field name in `on`? Some more explanation would be great.

docs-website/src/pages/docs/directives-reference/inject-current-datetime-directive.md
```diff
@@ -169,13 +153,36 @@ mutation (
...
+  email: String!
+  name: String!
+  createdAt: DateTime!
+}
+```
+
+We can use `@injectCurrentDateTime` to set the value of `createdAt` using:
+
+```graphql
+mutation ($input: createUserInput! @injectCurrentDateTime(on: "createdAt")) {
```
Pagebakers, line 181: Can I inject multiple values into the input?

For example?
```
$input: createUserInput! @injectCurrentDateTime(on: "createdAt") @injectEnvironmentVariable(name: "ID", on: "shopId")
```

docs-website/src/pages/docs/directives-reference/json-schema-directive.md
```diff
@@ -152,3 +152,28 @@ successfully. If it has boolean value true, the instance validates
...
+  id: String!
+  email: String!
+  name: String!
+}
+```
+
+We can use `@jsonSchema` to set the JSON schema of `id` using:
+
+```graphql
+mutation ($input: createUserInput! @jsonSchema(pattern: "[0-9]+", on: "id")) {
```
Pagebakers, line 172: Nice! Like my previous comment, can I add patterns to multiple fields in $input?

Pagebakers (changes requested):
Good stuff, I'd like to clarify the documentation a bit.

It seems that the directives are now 'repeatable' but I'm not sure, if so please add some examples to the docs as well to make it clear to the dev.

docs-website/src/pages/docs/directives-reference/json-schema-directive.md
jensneuse: This was the intended design, but we need to ask @fiam for clarification.

fiam: All these directives can be used multiple times, targeting different fields. I'll update the documentation to explain that better.

docs-website/src/pages/docs/directives-reference/from-claim-directive.md
fiam: Targeting an invalid field in the on argument will throw an error during generation, but not a very obvious one. I'll update it to provide a better error message.

Injected paths are removed from the operation input's schema, but I have not tested how that affects the generate client. I'll check it.

spetrunin:
seems at the moment prevent infinite recursive calls logic somehow is not working

so recursive hooks tests are commented out in the main branch
https://github.com/wundergraph/wundergraph/blob/main/testapps/hooks/test/index.test.ts#L10

fiam:
> seems at the moment prevent infinite recursive calls logic somehow is not working
> 
> so recursive hooks tests are commented out in the main branch [`main`/testapps/hooks/test/index.test.ts#L10](https://github.com/wundergraph/wundergraph/blob/main/testapps/hooks/test/index.test.ts?rgh-link-date=2023-02-22T21%3A12%3A16Z#L10)

That's weird, they were working for me and they still are. I'll submit a PR.

docs-website/src/pages/docs/directives-reference/json-schema-directive.md
fiam: I've updated the docs to clarify that all these directives can be used multiple times and combined.

docs-website/src/pages/docs/directives-reference/inject-current-datetime-directive.md
fiam: Docs updated to reflect that you can

docs-website/src/pages/docs/directives-reference/from-claim-directive.md
fiam: Injected values are removed from the generated client inputs, but we're not removing them from the internal operations yet (tracked in ENG-1350). I've added a better error message when a referenced field doesn't exist.

image

Pagebakers (approved):

devsergiy (approved):

PR #682, opened 2023-02-17
fix: should skip file when file in diff dir

javacode123:
i do not think we should filter operation file under dir. when file in different dir, it will be skiped, such as:
AB.graphql
A/B.graphql
both operation name will be AB, so it will skip A/B. So. i del filer logic. is there any other design to slove this problem?

### DIFF
--- a/pkg/loadoperations/loader.go
+++ b/pkg/loadoperations/loader.go
@@ -160,15 +160,6 @@ func (l *Loader) readGraphQLOperation(filePath string) {
 	}
 
 	operationName := normalizeOperationName(fileName)
-
-	for _, file := range l.out.GraphQLOperationFiles {
-		if file.OperationName == operationName {
-			l.out.Info = append(l.out.Info, fmt.Sprintf(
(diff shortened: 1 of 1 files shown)


spetrunin:
@javacode123 as both such operations will be mounted as `/AB` we don't know how to fix such conflict so we will show a warning

and you could ether rename directory or rename your operation

this check is required to been able to provide a warning

javacode123:
hi, i do not know why we must normalizeOperationName operation name, mybe we can directly use {dir}+'/'+{operations_name}

devsergiy:
@javacode123 operation name in this case should be readable and uniq as it is used as identifier for the generated models definitions for the typescript client 

while operation route will be mounted following the folders structure

What problems do you have with this approach?

javacode123:
>  just confused. now i got your mean. i think the main problem is that we need give every operation a name which should be readable and uniq as it is used as identifier for the generated models definitions for the typescript client. but we can not use {foldername/filename} directly, because it contain file separator which should not as a model name in ts

PR #701, opened 2023-02-27
refactor: pass a bytes.Buffer to hooks.EncodeData and reuse it

fiam:


fiam:
Converted to draft until we add parallel tests

jensneuse (approved):

testapps/default/test/index.test.ts
```diff
@@ -1,25 +1,28 @@
 import { afterAll, beforeAll, describe, expect, test } from '@jest/globals';
-import { InputValidationError } from '@wundergraph/sdk/client';
+import { ClientResponse, InputValidationError } from '@wundergraph/sdk/client';
+import { NestedInternalResponse } from '../.wundergraph/generated/models';
 import { createTestServer } from '../.wundergraph/generated/testing';
 
 const wg = createTestServer();
 
-// TODO: Remove this once the secret refactor is done
-process.env['WG_SECURE_COOKIE_HASH_KEY'] = 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa';
```
StarpTech, line 8: Any other places where we can get rid of it?

fiam: I think we can get rid of it everywhere, but let's do it in a separate PR.

testapps/default/test/index.test.ts
```diff
@@ -1,25 +1,28 @@
...
 describe('functions', () => {
 	test('internal operation call from  function', async () => {
 		const client = wg.client();
-		const result = await client.query({
-			operationName: 'nested/InternalWrapper',
-		});
-		expect(result.error).toBeUndefined();
-		expect(result.data?.data?.chinook_findFirstAlbum?.AlbumId).toBe(1);
+		const promises: Promise<ClientResponse<NestedInternalResponse>>[] = [];
+		for (let ii = 0; ii < 50; ii++) {
```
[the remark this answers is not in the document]
fiam: It's to exercise a path with multiple requests in parallel. I'll add a comment to make sure it's not removed by accident.

testapps/default/test/index.test.ts
fiam: Tracked in ENG-1684

testapps/default/test/index.test.ts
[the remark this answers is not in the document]
fiam: Comment added. PTAL

PR #702, opened 2023-02-28
chore: fix prettier formatting check when filenames contain $

fiam:
$ needs to be escaped by sed and not passed through xargs (which swallows
the escape sequence)


StarpTech (approved):
LGTM

JivusAyrus (approved):

PR #703, opened 2023-02-28
fix: operation name normalization on Windows

fiam:
Code for loading .graphql operations was different than for .ts operations.
Unify loading for both types of operations and fix Unix-isms. Fix also a
couple of typos in user visible log messages.


Aenimus (approved):

pkg/loadoperations/loader.go
```diff
@@ -115,64 +115,89 @@ func (l *Loader) readOperations() error {
...
+	return unixLikeRelativeFilePathNonExt
+}
+
+// ensureUniqueOperationName checks that the given normalized operationName is not already used
+// by another operation. If the name is already taken, it appends an INFO log entry to
+// the Loader's output and returns false. Otherwise it returns true.
+func (l *Loader) normalizedUniqueOperationName(relativeFilePath string) (mountPath string, operationName string, ok bool) {
+	mountPath = l.operationMountPath(relativeFilePath)
+	if mountPath == "" {
+		return "", "", false
```
devsergiy, line 169: could we extend the test to include the case of conflicting names between ts and graphql operations

devsergiy (changes requested):

pkg/loadoperations/loader.go
fiam: It's checked just below this: [line 173](https://github.com/wundergraph/wundergraph/pull/703/files#diff-e12ef71dec9cf88b0a4001dc6b3004a5ab5058f4bd5bdf190936a614893d920bR173) and [line 181](https://github.com/wundergraph/wundergraph/pull/703/files#diff-e12ef71dec9cf88b0a4001dc6b3004a5ab5058f4bd5bdf190936a614893d920bR181)

jensneuse:
@fiam I think we load the path from the json and use it to load the modules, e.g. to generate custom config
Will this work as well with your changes? I have never tried it on windows?

fiam:
> @fiam I think we load the path from the json and use it to load the modules, e.g. to generate custom config
> Will this work as well with your changes? I have never tried it on windows?

That part was fixed in #607, but we missed that its GraphQL counterpart had a similar problem. This PR unifies the code for loading operations regardless of their type, so a problem in there should be easier to catch moving forward (we have nested TS operations in the examples CI). I'll trigger examples CI just to be safe.

fiam:
@jensneuse @StarpTech NPM CI is failing because we’ve been enforcing alphanumeric operation names, but only for GraphQL operations, and we also have operations in the nextjs testapp that use underscores (e.g. users/subscribe_no_input.ts) - I think we need to be consistent in this, we either allow underscores/dashes in all types or in none. Question is, should we rename the operations in the nextjs testapp or change the accepted operation names to allow dashes and underscores?

jensneuse:
> @jensneuse @StarpTech NPM CI is failing because we’ve been enforcing alphanumeric operation names, but only for GraphQL operations, and we also have operations in the nextjs testapp that use underscores (e.g. users/subscribe_no_input.ts) - I think we need to be consistent in this, we either allow underscores/dashes in all types or in none. Question is, should we rename the operations in the nextjs testapp or change the accepted operation names to allow dashes and underscores?

I think it's fine to allow it.

fiam:
> I think it's fine to allow it.

Wouldn't that be problematic if we expose those operations to GraphQL?

fiam:
> > I think it's fine to allow it.
> 
> Wouldn't that be problematic if we expose those operations to GraphQL?
> 
> 

Now that I think about it, underscores should be fine. Dashes would be the problematic ones.

fiam:
@devsergiy PTAL - Thanks!

pkg/loadoperations/loader.go
devsergiy: yeah, but under extend test I mean add a test case here https://github.com/wundergraph/wundergraph/blob/main/pkg/loadoperations/loader_test.go

fiam: I've added a conflicting `.ts` operation to the test and updated the golden file

devsergiy (approved):

PR #706, opened 2023-02-28
feat: add support for declaring a list of public claims

fiam:
If non-empty, all other claims are hidden from the frontend.

- Add PublicCustomClaims type, which picks fields from CustomClaims
- Declare a PublicUser type in client.ts, which optionally picks the
selected fields.
- Add required machinery in the Go side to handle these fields
- Validate the selected claims during code generation


pkg/authentication/authentication.go
```diff
@@ -732,6 +797,7 @@ type UserHandler struct {
 	Host              string
 	InsecureCookies   bool
 	Cookie            *securecookie.SecureCookie
+	PublicClaims      []string
```
jensneuse, line 800: We don't need to do this. We have this information in the API configuration. Cookie space is limited.

jensneuse (changes requested):

packages/sdk/src/configure/index.ts
```diff
@@ -115,7 +116,26 @@ export interface WunderGraphConfigApplicationConfig {
...
+		/**
+		 * Claims to be publicly available (i.e. served by the API to the frontend), referenced by
+		 * their shorthand name for custom claims (i.e. keys in the customClaims attribute) or by their
+		 * enum value for well known ones. If the list is empty, all claims are made public.
+		 *
+		 * @default empty
+		 *
+		 * @see WunderGraphConfigApplicationConfig.customClaims
+		 */
+		publicClaims?: string[];
```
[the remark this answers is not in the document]
fiam: We can't generate an enum for that, since the user needs to write a `WunderGraphConfigApplicationConfig` and call `configureWunderGraphApplication()` (where that input field is) before code generation is run. However, the strings that are provided there are currently validated at generation time, causing a failure if any of them is incorrect. Validation is [here](https://github.com/wundergraph/wundergraph/pull/706/files#diff-944359769ce5d07540eab5d54fc4358ff90dd67b5223918bed68053b8edce9b7R312)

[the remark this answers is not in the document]
fiam: Something that we can do is create a type and provide predefined values for the well known claims, so the list can look like `publicClaims: [WellKnownClaim.EMAIL, CustomClaim('shopID')]`. That gives us type safety at "typing time" for the well known claims (as well as allowing us to move some logic to there), but I don't think there's much else we can do.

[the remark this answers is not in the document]
jensneuse: I think it should be possible. The well-known claims are known ahead of time. There should be a way to infer names of the custom claims and combine both into a list of possible names. Not sure if this is exactly the solution, but it might be close: https://stackoverflow.com/a/66723158/20693155

pkg/authentication/authentication.go
jensneuse: ok

packages/sdk/src/configure/index.ts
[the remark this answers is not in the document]
fiam: Turns out this was possible, but I had to use a different approach because I couldn't use `this` in that context. PTAL

jensneuse (approved):

packages/sdk/src/configure/index.ts
```diff
@@ -286,6 +312,22 @@ export interface CodeGenerationConfig {
 	wunderGraphDir: string;
 }
 
+const resolvePublicClaims = (config: WunderGraphConfigApplicationConfig) => {
+	let publicClaims: string[] = [];
+	for (const claim of config.authentication?.publicClaims ?? []) {
+		let customClaim = config.authentication?.customClaims?.[claim];
```
Aenimus, line 318: Why `let`?

packages/sdk/src/configure/index.ts
```diff
@@ -1169,9 +1218,13 @@ const resolveOperationsConfigurations = async (
 			default:
 				throw new Error(`customClaim ${key} has invalid type ${claim.type}`);
 		}
+		const jsonPathComponents = claim.jsonPath.split('.');
+		if (jsonPathComponents.length == 0) {
```
Aenimus, line 1222: If you want to coerce, `!jsonPathComponents.length`, otherwise `===`

packages/sdk/src/definition/merge.ts
```diff
@@ -254,7 +255,34 @@ enum WG_ROLE {
...
+		GENDER: 'gender',
+		BIRTH_DATE: 'birthdate',
+		ZONE_INFO: 'zoneinfo',
+		LOCALE: 'locale',
+		LOCATION: 'location',
+	};
+	if (Object.keys(docs).length !== WellKnownClaimValues.length) {
+		throw new Error('unhandled claims in claimsSchema()');
+	}
+	const wellKnownClaims = Object.keys(docs).map((key) => `"""${docs[key as WellKnownClaim]}"""\n${key}`);
```
Aenimus, line 284: Entries

Aenimus: I think this would allow you to remove the `as`

packages/sdk/src/graphql/operations.ts
```diff
@@ -458,11 +486,27 @@ const parseWellKnownClaim = (name: string, operation?: GraphQLOperation) => {
 		LOCALE: ClaimType.LOCALE,
 		LOCATION: ClaimType.LOCATION,
 	};
+	if (Object.keys(claims).length !== WellKnownClaimValues.length) {
+		throw new Error('unhandled claims in parseWellKnownClaim()');
+	}
 	if (name in claims) {
```
Aenimus, line 492: Just curious, is this O(1)?

pkg/authentication/authentication.go
```diff
@@ -167,23 +173,82 @@ type User struct {
...
+func (u *User) ToPublic(publicClaims []string) *User {
+	if len(publicClaims) == 0 {
+		return u
+	}
+	cpy := &User{
+		CustomAttributes: u.CustomAttributes,
+		Roles:            u.Roles,
+	}
+	for _, claim := range publicClaims {
+		if !cpy.copyWellKnownClaim(claim, u) {
```
Aenimus, line 198: I prefer
```go
if cpy.copyWellKnownClaim(claim, u) {
    continue
}
...```

File: pkg/authentication/authentication.go [hunk 8]
```diff
@@ -799,19 +865,18 @@ func (u *UserHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-	_ = json.NewEncoder(w).Encode(user)
+	if err := encoder.Encode(user.ToPublic(u.PublicClaims)); err != nil {
+		u.Log.Error("encoding user", zap.Error(err))
+	}
 }
 
 type CSRFTokenHandler struct{}
 
-func (_ *CSRFTokenHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+func (*CSRFTokenHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
```
Aenimus, line 879: Huh?

pkg/authentication/oidc.go
```diff
@@ -108,7 +108,7 @@ func (c *Claims) ToUser() User {
 	}
 }
 
-// isCustomClaim true if claim represents a token claim that we're not explicitely
+// isCustomClaim true if claim represents a token claim that we're not explicitly
```
Aenimus, line 111: lol

pkg/jsonpath/jsonpath.go
```diff
@@ -0,0 +1,44 @@
...
+package jsonpath
+
+// GetKeys retrieves the value at the given path by following each key in data. If
+// data is not a map or if a key is not present, it returns nil.
+func GetKeys(data interface{}, jsonKeys ...string) interface{} {
+	return getKeys(data, jsonKeys, 0)
+}
+
+func getKeys(data interface{}, keys []string, keyIndex int) interface{} {
+	if keyIndex < len(keys) {
```
Aenimus, line 11: Invert the logic to reduce nesting? :'(

pkg/jsonpath/jsonpath.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+	return nil
+}
+
+func SetKeys(data interface{}, value interface{}, jsonKeys ...string) map[string]interface{} {
+	return setKeys(data, value, jsonKeys, 0)
+}
+
+func setKeys(data interface{}, value interface{}, jsonKeys []string, keyIndex int) map[string]interface{} {
+	if keyIndex < len(jsonKeys) {
```
Aenimus, line 29: Same here

Aenimus (changes requested):
Comments for changes

pkg/authentication/authentication.go
fiam: Receiver is not used, so it should be omitted

packages/sdk/src/configure/index.ts
fiam: Oversight, I'll change it

packages/sdk/src/configure/index.ts
fiam: `jsonPathComponents` can't be null nor undefined at that point, but I'll change it to `===`

packages/sdk/src/graphql/operations.ts
fiam: `Object.keys()` is O(n) but `claims` has a fixed size

pkg/jsonpath/jsonpath.go
fiam: I've moved the length check to the callers, since it only makes a difference when the keys slice is empty. This way we remove a level of indirection as well as making it slightly faster.

packages/sdk/src/codegen/templates/typescript/claims.ts
```diff
@@ -1,14 +1,15 @@
...
 import { handlebarTemplate } from './claims.template';
-
 export class TypeScriptClaims implements Template {
 	generate(generationConfig: CodeGenerationConfig): Promise<TemplateOutputFile[]> {
 		const tmpl = Handlebars.compile(handlebarTemplate);
-		const customClaims = Object.keys(generationConfig.config.authentication.customClaims ?? []).map((key) => {
-			const claim = generationConfig.config.authentication.customClaims[key];
+		const claims = generationConfig.config.authentication.customClaims;
+		const customClaims = Object.keys(claims).map((key) => {
+			const claim = claims[key];
```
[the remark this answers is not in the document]
fiam: Changed to Object.entries

packages/sdk/src/configure/index.ts
fiam: Changed

packages/sdk/src/configure/index.ts
fiam: Changed

packages/sdk/src/graphql/operations.ts
Aenimus: No, I mean "in" for the records.

pkg/authentication/authentication.go
Aenimus: I didn't know you could just not have a ref

pkg/jsonpath/jsonpath.go
fiam: Moved the check to the caller

fiam:
> Comments for changes

Thanks for the review. I think I've addressed all requested changes. PTAL

packages/sdk/src/configure/index.ts
```diff
@@ -286,6 +312,22 @@ export interface CodeGenerationConfig {
 	wunderGraphDir: string;
 }
 
+const resolvePublicClaims = (config: WunderGraphConfigApplicationConfig) => {
+	let publicClaims: string[] = [];
```
Aenimus, line 316: Should be a const.

pkg/authentication/authentication.go
```diff
@@ -77,7 +79,7 @@ func (u *UserLoader) parseClaims(r io.Reader) (*Claims, error) {
 }
 
 // Load user from token. token is always non nil and contains at least a non-empty token.Raw
-// but it might be unvalidated if we have no key functions
+// but it might be not validated if we have no key functions
```
Aenimus, line 82: might not be

fiam: Thanks! Reworded.

packages/sdk/src/configure/index.ts
Aenimus: Will you address this @fiam

fiam: Sorry, I missed this one. It should be good now.

Aenimus (approved):

fiam:
Thanks for the thorough review @Aenimus 👍

fiam:
I've just realized we're missing some docs for this feature. Moving into draft to prevent merge.

fiam:
Docs added

PR #709, opened 2023-03-01
feat: improve webhooks API and add support for the new operations client

Pagebakers:
#### Checklist

- [ ] run `make test`


StarpTech (approved):
Great PR! A good next step could be to allow the same nested structure as with operations.

Aenimus (approved):

jensneuse:
@Pagebakers I have an idea for further improvement of the operations client.
When you instantiate the operations client, we can pass the "webhook request" as client request into the operations client.
This way, subsequent operations will have access to the original webhook request, which is exactly what we want.

PR #712, opened 2023-03-01
feat: otel, add tracer

YuriBuerov:
## Motivation and Context

Open Telemetry instrumentation for WunderGraph node, and hooks server

#### Checklist

- [x] run `make test`


pkg/telemetry/otel/trace/trace.go
```diff
@@ -0,0 +1,111 @@
...
+
+	p := sdktrace.NewTracerProvider(
+		sdktrace.WithBatcher(exporter),
+		sdktrace.WithResource(sdkresource.NewWithAttributes(
+			semconv.SchemaURL,
+			semconv.ServiceNameKey.String(config.ServiceName),
+		)),
+	)
+
+	otel.SetTracerProvider(p)
```
StarpTech, line 77: This will override the global provider. We should only work with local ones.

pkg/apihandler/apihandler.go
```diff
@@ -794,6 +794,8 @@ var (
 
 func (h *GraphQLHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	requestLogger := h.log.With(logging.WithRequestIDFromContext(r.Context()))
+	span := trace.SpanFromContext(r.Context())
+	defer span.End()
```
StarpTech, line 798: With this approach how do we track errors? I think it is very cumbersome to call `span.RecordError` on all possible places. I think we need another abstraction (middleware?) to start, end, and record errors in a request/response lifecycle.

pkg/hooks/hooks.go
```diff
@@ -185,6 +186,7 @@ func buildClient(requestTimeout time.Duration) *retryablehttp.Client {
 		}
 		return retryablehttp.DefaultRetryPolicy(ctx, resp, err)
 	}
+	httpClient.HTTPClient.Transport = trace.HTTPClientTransporter(httpClient.HTTPClient.Transport)
```
StarpTech, line 189: - Remember that this does not replace hotel instrumentation in Node.js.
- Did you already think of how can instrument calls to the origins?

StarpTech (changes requested):

pkg/telemetry/otel/trace/trace.go
```diff
@@ -0,0 +1,98 @@
...
+func NewNoopTracerProvider() *TracerProvider {
+	return &TracerProvider{
+		Provider: trace.NewNoopTracerProvider(),
+	}
+}
+
+func configureExporter(ctx context.Context, config *TracerProviderConfig) (sdktrace.SpanExporter, error) {
+	switch {
+	case config.Endpoint != "":
+		client := otlptracehttp.NewClient(
```
StarpTech, line 67: This must be configurable. In prod, we will use the secure variant.

pkg/telemetry/otel/trace/trace.go
```diff
@@ -0,0 +1,98 @@
...
+			otlptracehttp.WithInsecure(),
+			otlptracehttp.WithEndpoint(config.Endpoint),
+		)
+
+		return otlptrace.New(ctx, client)
+	case config.JaegerEndpoint != "":
```
StarpTech, line 73: Can we make it configurable with a proper options API?
E.g `NewTracerProvider(withJaeger(...))` In that way, you can also validate if the right fields are set.

pkg/telemetry/otel/trace/middleware.go
```diff
@@ -0,0 +1,186 @@
...
+	status  int
+}
+
+var rrwPool = &sync.Pool{
+	New: func() interface{} {
+		return &recordingResponseWriter{}
+	},
+}
+
+func getRRW(writer http.ResponseWriter) *recordingResponseWriter {
```
StarpTech, line 62: What means RRW?

pkg/telemetry/otel/trace/middleware.go
```diff
@@ -0,0 +1,186 @@
...
+type transport struct {
+	rt http.RoundTripper
+}
+
+func (t *transport) RoundTrip(r *http.Request) (*http.Response, error) {
+	span := SpanFromContext(r.Context())
+
+	// otelhttp client transport does not provide a path to the span name formatter, so we have to do it manually
+	// redefine the span name
+	span.SetName(fmt.Sprintf("HTTP %s %s", r.Method, r.URL.Path))
```
StarpTech, line 170: Please ensure that the span name does not change per request. It's more used as a category like operationName, RPC call. See https://opentelemetry.io/docs/reference/specification/trace/api/#span

pkg/telemetry/otel/trace/middleware_test.go
```diff
@@ -0,0 +1,81 @@
...
+		assert.Equal(t, sc, got)
+		w.WriteHeader(http.StatusOK)
+	})
+
+	r := httptest.NewRequest("GET", "/operation/Image", nil)
+	r = r.WithContext(oteltrace.ContextWithRemoteSpanContext(context.Background(), sc))
+	w := httptest.NewRecorder()
+
+	router.ServeHTTP(w, r)
+	assert.True(t, called)
```
StarpTech, line 80: You tested that the handler is called, not that tracer worked adequately.

pkg/apihandler/internalapihandler.go
```diff
@@ -334,6 +345,16 @@ type InternalSubscriptionApiHandler struct {
 }
 
 func (h *InternalSubscriptionApiHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	span := trace.SpanFromContext(r.Context())
+
+	trace.AddSpanTags(
+		span,
+		map[string]string{
+			"operation":     h.operation.Name,
```
StarpTech, line 353: We manage a namespace for our attributes. E.g.

- wg.operation.name
- wg.webhook.name
- wg.subscription (bool)

Check cloud shared/otel. We should follow them as closely as possible.

StarpTech (changes requested):

packages/sdk/src/server/plugins/hooks.ts
```diff
@@ -126,7 +127,15 @@ const FastifyHooksPlugin: FastifyPluginAsync<FastifyHooksOptions> = async (fasti
 		Body: { request: WunderGraphRequest; operationName: string; operationType: 'query' | 'mutation' | 'subscription' };
 	}>('/global/httpTransport/onOriginRequest', async (request, reply) => {
 		reply.type('application/json').code(200);
+
 		try {
+			const { span } = request.telemetry();
```
StarpTech, line 132: Put this in a pre-handler hook or middleware. You also forgot to instrument auth hooks.

packages/sdk/src/server/trace.ts
```diff
@@ -0,0 +1,67 @@
+import { Resource } from '@opentelemetry/resources';
+import { NodeTracerProvider, NoopSpanProcessor, SpanProcessor } from '@opentelemetry/sdk-trace-node';
+import { SemanticResourceAttributes } from '@opentelemetry/semantic-conventions';
+import { AlwaysOnSampler, BatchSpanProcessor, SpanExporter, ConsoleSpanExporter } from '@opentelemetry/sdk-trace-base';
+import { JaegerExporter } from '@opentelemetry/exporter-jaeger';
+import { OTLPTraceExporter } from '@opentelemetry/exporter-trace-otlp-http';
+import { TelemetryOptions, WunderGraphConfiguration } from '@wundergraph/protobuf';
+import { resolveConfigurationVariable } from '../configure/variables';
+// import { diag, DiagConsoleLogger, DiagLogLevel } from "@opentelemetry/api";
```
StarpTech, line 9: Delete or good for debugging? If so, make a comment

types/protos/wundernode_config.proto
```diff
@@ -522,12 +522,19 @@ message NodeLogging {
 	ConfigurationVariable level = 1;
 }
 
+message TelemetryOptions {
+	ConfigurationVariable otelEnabled = 1;
```
StarpTech, line 526: JWT secret is missing.

StarpTech (changes requested):
Great work! Few questions:

1. Did you ensure that OTEL can be enforced in cloud?
2. I think you forgot to instrument the OperationClient https://github.com/wundergraph/wundergraph/blob/yuri/eng-1434-add-otel-opentelemetry-to-wundernode/packages/sdk/src/server/operations-client.ts

docs-website/src/pages/docs/examples/open-telemetry.md
```diff
@@ -0,0 +1,51 @@
...
+---
+
+[Check the example](https://github.com/wundergraph/wundergraph/tree/main/examples/open-telemetry)
+
+## Configuration
+
+Check the `wundergraph.config.ts` confile:
+
+```typescript
+configureWunderGraphApplication({
```
jensneuse, line 14: How about this alternative configuration:


```typescript
configureWunderGraphApplication({
  apis: [spacex],
  options: {
    openTelemetry: {
      enable: true,
      exporterHTTPEndpoint: '',
      exporterJaegerEndpoint: 'http://localhost:14268/api/traces',
    },
  },
  // ...
});
```

jensneuse: Besides, what happens if I configure both endpoints? This should be documented.

docs-website/src/pages/docs/examples/open-telemetry.md
```diff
@@ -0,0 +1,51 @@
...
+Install the dependencies and run the complete example in one command:
+
+```shell
+npm install && npm start
+```
+
+## Get Dragons
+
+```shell
+curl -H "traceparent: 00-80e1afed08e019fc1110464cfa66635c-7a085853722dc6d2-01" http://localhost:9991/operations/Dragons
```
jensneuse, line 38: Explain this header in the docs.
Also explain what happens if we omit this header.

docs-website/src/pages/docs/examples/open-telemetry.md
```diff
@@ -0,0 +1,51 @@
...
+```
+
+## Test the tracing
+
+You can access Jaeger from http://localhost:16686/
+
+To obtain meaningful results, it is recommended to change the trace id in the request header every time you send a request.
+If you don't provide a traceparent header, a new trace id will be generated automatically.
```
jensneuse, line 46: You can add a reference to WG cloud that we do this automatically.

pkg/telemetry/otel/trace/trace.go
```diff
@@ -0,0 +1,98 @@
...
+	return &TracerProvider{
+		Provider: trace.NewNoopTracerProvider(),
+	}
+}
+
+func configureExporter(ctx context.Context, config *TracerProviderConfig) (sdktrace.SpanExporter, error) {
+	switch {
+	case config.Endpoint != "":
+		client := otlptracehttp.NewClient(
+			otlptracehttp.WithInsecure(),
```
jensneuse, line 68: There needs to be an option to set the JWT for auth against the public otel collector.

jensneuse (changes requested):

docs-website/src/pages/docs/examples/open-telemetry.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Environment variables configuration
+
+The Open Telemetry configuration can also be set via environment variables.
+
+```shell
+  WG_OTEL_ENABLED=true
+  WG_OTEL_EXPORTER_HTTP_ENDPOINT=''
+  WG_OTEL_EXPORTER_JAEGER_ENDPOINT='http://localhost:14268/api/traces'
+  WG_OTEL_JWT_TOKEN='' # production mode
```
StarpTech, line 63: Just `WG_OTEL_JWT`?

docs-website/src/pages/docs/examples/open-telemetry.md
```diff
@@ -0,0 +1,66 @@
...
+
+Configure hooks in the `wundergraph.server.ts` and check how the tracing data changes.
+
+## Environment variables configuration
+
+The Open Telemetry configuration can also be set via environment variables.
+
+```shell
+  WG_OTEL_ENABLED=true
+  WG_OTEL_EXPORTER_HTTP_ENDPOINT=''
```
StarpTech, line 61: This could be `http://localhost:4318`

docs-website/src/pages/docs/wundergraph-config-ts-reference/configure-open-telemetry.md
```diff
@@ -0,0 +1,68 @@
...
+If no exporter endpoint is configured, the default exporter is the `Stdout` which prints the traces to the console.
+If both exporter endpoints are configured, the `exporterHttpEndpoint` will be used.
+Configuration priority: `otelExporterHttpEndpoint` > `exporterJaegerEndpoint` > `Stdout`.
+
+```typescript
+// wundergraph.config.ts
+configureWunderGraphApplication({
+  apis: [spacex],
+  options: {
+    openTelemetry: {
```
StarpTech, line 23: OpenTelemetry is an implementation detail. Let's call it `telemetry`.

docs-website/src/pages/docs/wundergraph-config-ts-reference/configure-open-telemetry.md
```diff
@@ -0,0 +1,68 @@
...
+This section describes how to configure Open Telemetry in WunderGraph.
+
+OpenTelemetry is a collection of tools, APIs, and SDKs used to instrument, generate, collect, and export telemetry data (metrics, logs, and traces) for analysis in order to understand your software's performance and behavior.
+
+## Enable Open Telemetry
+
+To enable Open Telemetry, set the `enabled` option to `true` in the `telemetry` section of the `options` object in the `configureWunderGraphApplication` function.
+If no exporter endpoint is configured, the default exporter is the `Stdout` which prints the traces to the console.
+If both exporter endpoints are configured, the `exporterHttpEndpoint` will be used.
+Configuration priority: `otelExporterHttpEndpoint` > `exporterJaegerEndpoint` > `Stdout`.
```
StarpTech, line 16: I'd throw an error when both are set.

packages/sdk/src/server/plugins/telemetry.ts
```diff
@@ -0,0 +1,71 @@
...
+		}
+
+		span!.setAttributes({
+			'http.status_code': resp.statusCode,
+		});
+
+		span!.setStatus(spanStatus);
+		span!.end();
+
+		contextMap.delete(req);
```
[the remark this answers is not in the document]
StarpTech: What was the answer here?

docs-website/src/pages/docs/examples/open-telemetry.md
```diff
@@ -0,0 +1,66 @@
...
+
+```typescript
+// wundergraph.config.ts
+configureWunderGraphApplication({
+  apis: [spacex],
+  options: {
+    openTelemetry: {
+      enabled: true, // default: false
+      exporterHttpEndpoint: '', // standard otel http exporter endpoint
+      exporterJaegerEndpoint: '', // 'http://localhost:14268/api/traces' we recommed to use it for development
```
ggeijteman-wg, line 21: nit: recommend :)

docs-website/src/pages/docs/wundergraph-config-ts-reference/configure-open-telemetry.md
StarpTech: ok

jensneuse (changes requested):

docs-website/src/pages/docs/wundergraph-config-ts-reference/configure-open-telemetry.md
jensneuse: Good idea

StarpTech: @YuriBuerov thoughts?

docs-website/src/pages/docs/wundergraph-config-ts-reference/configure-open-telemetry.md
```diff
@@ -0,0 +1,71 @@
...
+  WG_OTEL_JWT='Bearer ...' # production mode
+```
+
+Please note that `wundergraph.config.ts` configuration has higher priority than environment variables.
+Empty values could be omitted.
+
+## Authentication
+
+OpenTelemetry collector can be secured with a JWT token.
+If `authToken` / `WG_OTEL_JWT` is set, the exporter will add the `Authorization` header to the request:
```
StarpTech, line 53: `Authorization Bearer` header

packages/sdk/src/server/trace.ts
```diff
@@ -0,0 +1,95 @@
...
+};
+
+// use for debugging
+// diag.setLogger(new DiagConsoleLogger(), DiagLogLevel.DEBUG);
+
+function getExporter(config?: TelemetryOptions): SpanExporter {
+	if (config?.exporterHttpEndpoint) {
+		const url = resolveConfigurationVariable(config.exporterHttpEndpoint);
+		if (url !== '') {
+			const [authHeader, hasAuthHeader] = getAuthHeader(config);
```
StarpTech, line 46: This looks wrong. The header must be in format `Authorization Bearer`. Please test against the public hotel collector.

StarpTech (changes requested):

docs-website/src/pages/docs/wundergraph-config-ts-reference/configure-open-telemetry.md
StarpTech: Clarified

docs-website/src/pages/docs/wundergraph-config-ts-reference/configure-open-telemetry.md
StarpTech: The configuration interface is fine. The reason for throwing an error is to catch misconfiguration. If the user can't run both, it makes no sense to allow both. It has a false positive impression.

packages/sdk/src/server/plugins/telemetry.test.ts
```diff
@@ -0,0 +1,180 @@
...
+			tracer = provider.getTracer('hook-server', '0.0.1');
+			span = tracer.startSpan('test');
+			tracerProvider = {
+				provider: provider,
+			};
+			app.register(FastifyTelemetryPlugin, tracerProvider);
+			app.post('/operation/Image/preResolve', async () => {
+				return { foo: 'bar' };
+			});
+			app.post('/operation/Image/postResolve', async (req, resp) => {
```
StarpTech, line 104: Could you make it clear in the path that this endpoint should error?

packages/sdk/src/server/plugins/telemetry.test.ts
```diff
@@ -0,0 +1,180 @@
...
+import { WunderGraphHooksAndServerConfig } from '../types';
+import { TelemetryTracerProvider } from '../trace';
+import { NodeTracerProvider, NoopSpanProcessor, Tracer } from '@opentelemetry/sdk-trace-node';
+import Fastify from 'fastify';
+import FastifyTelemetryPlugin from './telemetry';
+import { Span, SpanStatusCode } from '@opentelemetry/api';
+
+describe('FastifyTelemetryPlugin', () => {
+	describe('register plugins', () => {
+		it('should add telemetry data to request object', async () => {
```
StarpTech, line 13: You mean, "It should add telemetry request decorator..."?

packages/sdk/src/server/plugins/telemetry.test.ts
```diff
@@ -0,0 +1,180 @@
...
+			});
+			app.ready();
+		});
+
+		afterEach(async () => {
+			jest.clearAllMocks();
+		});
+
+		afterAll(async () => {
+			await app.close();
```
StarpTech, line 115: You can just return app.Close

packages/sdk/src/server/plugins/telemetry.test.ts
```diff
@@ -0,0 +1,180 @@
...
+				},
+			});
+
+			expect(app.hasRequestDecorator('telemetry')).toBe(true);
+			expect(tracer.startSpan).toBeCalledTimes(1);
+			expect(span.setAttributes).toBeCalledWith({
+				'http.method': 'POST',
+				'http.url': '/operation/Image/postResolve',
+			});
+			expect(span.setAttributes).toBeCalledWith({
```
StarpTech, line 167: Where are the wg attributes?

packages/sdk/src/server/plugins/telemetry.test.ts
```diff
@@ -0,0 +1,180 @@
...
+		afterAll(async () => {
+			await app.close();
+		});
+
+		it('should trace a request', async () => {
+			jest.spyOn(provider, 'getTracer').mockReturnValue(tracer);
+			jest.spyOn(tracer, 'startSpan').mockReturnValue(span);
+			jest.spyOn(span, 'setStatus');
+			jest.spyOn(span, 'setAttributes');
+			jest.spyOn(span, 'end');
```
StarpTech, line 123: Instead of spying, we should use **InMmeorySpanExporter** https://github.com/open-telemetry/opentelemetry-js/blob/93285585ce816a30df8db2b4bdd6d39fd78da806/experimental/packages/opentelemetry-instrumentation-http/test/functionals/http-enable.test.ts#L68 and call ` const spans = memoryExporter.getFinishedSpans();` at the end of each test. This gives us more confidence and is easier to handle.

pkg/apihandler/apihandler.go
```diff
@@ -1253,6 +1254,14 @@ type QueryHandler struct {
 func (h *QueryHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	requestLogger := h.log.With(logging.WithRequestIDFromContext(r.Context()))
 	r = setOperationMetaData(r, h.operation)
+	span := trace.SpanFromContext(r.Context())
+	trace.AddSpanTags(
+		span,
+		map[string]string{
```
StarpTech, line 1260: This is repeated. Can we build a function to enforce that structure e.g `GetOperationSpanTags`

StarpTech (changes requested):

packages/sdk/src/server/plugins/telemetry.test.ts
YuriBuerov: Could you explain how exactly it will improve these tests? We test the order of calls, returned values, and passed attributes. What additionally could be tested using `InMemoryExporter`?

packages/sdk/src/server/plugins/telemetry.test.ts
YuriBuerov: what's wrong with the `afterAll` hook? I'm not sure I understand your suggestion.

pkg/apihandler/apihandler.go
YuriBuerov: This is repeated until we have clear requirements on what unique attributes we need in each particular case

packages/sdk/src/server/plugins/telemetry.test.ts
YuriBuerov: Looking into the test you mentioned, I see no difference. It checks how many spans were created, attributes, and span names.

packages/sdk/src/server/plugins/telemetry.test.ts
StarpTech: Just to simplify it to `afterAll(() => app.close())`

packages/sdk/src/server/plugins/telemetry.test.ts
StarpTech: It's just about readability. It's just a suggestion. I have no strong opinion.

packages/sdk/src/server/plugins/telemetry.test.ts
StarpTech: Then I misunderstood the test, and we need to add tests for our actual implementation.

packages/sdk/src/server/plugins/telemetry.test.ts
StarpTech: It's more of a best practice. Spy relies on the actual implementation of the tracer. With the inMemoryExporter you could just test the result.

packages/sdk/src/server/trace.ts
```diff
@@ -0,0 +1,125 @@
...
+};
+
+function getExporter(config?: TelemetryOptions, logger?: pino.Logger): SpanExporter {
+	const httpEndpoint = config?.exporterHttpEndpoint ? resolveConfigurationVariable(config.exporterHttpEndpoint) : '';
+	const jaegerEndpoint = config?.exporterJaegerEndpoint
+		? resolveConfigurationVariable(config.exporterJaegerEndpoint)
+		: '';
+
+	if (httpEndpoint !== '' && jaegerEndpoint !== '') {
+		throw new Error('cannot configure both OTLP and Jaeger exporters');
```
StarpTech, line 52: Better would be `Cannot configure both OTLP and Jaeger endpoints` ?

packages/sdk/src/server/plugins/telemetry.test.ts
YuriBuerov: What do you mean by: for our actual implementation? We check that we register plugin successfully, this plugin contains the main logic, it propagates the span and the context everywhere. We check that request was decorated. We check what attributes we pass, the error we log, and when we end it. We don't check only custom attributes we pass in each particular hook, using this plugin.

packages/sdk/src/server/plugins/telemetry.test.ts
YuriBuerov: The result created span, and passed attributes? I'm ok if it's a personal preference. But it won't be delivered forever this way

packages/sdk/src/server/trace.ts
YuriBuerov: But we configure exporters not endpoints which is a configuration parameter passed by user?

packages/sdk/src/server/plugins/telemetry.test.ts
StarpTech: Here we have no tests for those places:

- https://github.com/wundergraph/wundergraph/pull/712/files#diff-2f9c5d63c84e8681f889caeb22b494c16f93545d9368d37cf2d15829be1db0b5R43
- https://github.com/wundergraph/wundergraph/pull/712/files#diff-107fe1c4d7590a60a471913fa23ac45ea5111116a722b7609f5b37c6dc22feadR62
- https://github.com/wundergraph/wundergraph/pull/712/files#diff-107fe1c4d7590a60a471913fa23ac45ea5111116a722b7609f5b37c6dc22feadR153

Go

- https://github.com/wundergraph/wundergraph/pull/712/files#diff-1a96a1da77525d4d78f508c5047d48be421b0da7ade77b887a69232107d31546R1258
- https://github.com/wundergraph/wundergraph/pull/712/files#diff-f070476f0b264aae4526fb1388c89665f28ec45309b13381fe38521512ebcb33R260
- https://github.com/wundergraph/wundergraph/pull/712/files#diff-f070476f0b264aae4526fb1388c89665f28ec45309b13381fe38521512ebcb33R352

We also don't test if the trace is properly passed to the internal client.

packages/sdk/src/server/trace.ts
StarpTech: My concern is: Does the user has enough context to fix the error when he reads this message?

YuriBuerov:
@StarpTech feel free to take it over. I improved tests, but it would be great to add tests to the functions, I was not sure how to do it properly. I resolved conflicts and adapted the changes which were made to the internal client.  I tested it, but it would be reasonable to double-check it. Apart from that, I think it's ready to be merged.

StarpTech:
Closed in favor of https://github.com/wundergraph/wundergraph/pull/1004

PR #715, opened 2023-03-02
fix: crash in ServerPortFromConfig when configuration is invalid

fiam:
Allows us to provide a better error message than crashing without an
explanation

### DIFF
--- a/cli/helpers/process.go
+++ b/cli/helpers/process.go
@@ -2,6 +2,7 @@ package helpers
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"os/exec"
@@ -21,7 +22,7 @@ func ServerPortFromConfig(configJsonPath string) (int, error) {
 	}
 
 	if len(data) == 0 {
-		return 0, fmt.Errorf("config file is empty")
+		return 0, errors.New("config file is empty")
 	}
 
 	var graphConfig struct {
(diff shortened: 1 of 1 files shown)


cli/helpers/process.go
```diff
@@ -33,8 +34,13 @@ func ServerPortFromConfig(configJsonPath string) (int, error) {
...
 
-	newPort := loadvariable.Int(graphConfig.Api.ServerOptions.Listen.Port)
-	return newPort, nil
+	if graphConfig.Api != nil && graphConfig.Api.ServerOptions != nil {
+		variable := graphConfig.Api.ServerOptions.GetListen().GetPort()
+		if variable != nil {
+			return loadvariable.Int(variable), nil
+		}
+	}
+	return 0, errors.New("configuration is invalid")
```
devsergiy, line 43: What are the circumstances when it happens?

devsergiy (approved):

cli/helpers/process.go
fiam: Someone reported it on Discord a few weeks ago. Their problem was that they updated the SDK and `npm start` crashed on startup because `ServerPortFromConfig` is called early, even before we update the configuration, during `wunderctl up` to potentially kill previous instances of the hooks server. For this reason, we need to make it resilient to an invalid configuration, otherwise the user needs to manually delete the generated `wundergraph.config.json` to proceed.

Aenimus (approved):

PR #716, opened 2023-03-02
feat: add support for generating OpenAPI specs for applications

fiam:
- feat: initial support for OpenAPI generation
- fix: lazily initialize definitions during op input schema generation
- fix: typo in user visible string in postman collection


docs-website/src/pages/docs/features/openapi-postman.md
```diff
@@ -0,0 +1,15 @@
...
+pageTitle: WunderGraph - OpenAPI definitions and Postman collection
+---
+
+## OpenAPI and Postman collection generation
+
+Out of the box, WunderGraph supports generating an OpenAPI Specification as well as a Postman collection to
+interact with your APIs. You can use them to test your API in Postman or to generate bindings to query your
+API from another language.
+
+After running `wunderctl generate`, (usually via `npm build` or `npm build:wundergraph`) both an OpenAPI
```
Aenimus, line 12: Parenthetical statement before the comma. Alternatively, remove the parentheses and replace the closing one with another comma.

packages/sdk/src/configure/index.ts
```diff
@@ -1192,6 +1214,86 @@ const trimTrailingSlash = (url: string): string => {
...
 
+// typeScriptOperationsResponseSchemas generates the response schemas for all TypeScript
+// operations at once, since it's several times faster than generating them one by one
+const typeScriptOperationsResponseSchemas = (operations: GraphQLOperation[]) => {
+	const functionTypeName = (op: GraphQLOperation) => `function_${op.Name}`;
+	const responseTypeName = (op: GraphQLOperation) => `${functionTypeName(op)}_Response`;
+
+	const programFile = 'typescript_schema_generator.ts';
+
+	let contents: string[] = ['import type { ExtractResponse } from "@wundergraph/sdk/operations";'];
```
Aenimus, line 1225: Const

packages/sdk/src/configure/index.ts
```diff
@@ -1192,6 +1214,86 @@ const trimTrailingSlash = (url: string): string => {
...
+		contents.push(`import type ${name} from "${relativePath}";`);
+		contents.push(`export type ${responseTypeName(op)} = ExtractResponse<typeof ${name}>`);
+	}
+
+	const cachePath = path.join('cache', `ts.operationTypes.${objectHash(contents)}.json`);
+	if (fs.existsSync(cachePath)) {
+		try {
+			const cached = fs.readFileSync(cachePath, { encoding: 'utf-8' });
+			return JSON.parse(cached) as Record<string, JSONSchema>;
+		} catch {}
```
Aenimus, line 1239: Empty catch?

packages/sdk/src/configure/index.ts
```diff
@@ -1192,6 +1214,86 @@ const trimTrailingSlash = (url: string): string => {
...
+
+	const compilerOptions: TJS.CompilerOptions = {
+		strictNullChecks: true,
+		noEmit: true,
+		ignoreErrors: true,
+	};
+
+	const program = TJS.getProgramFromFiles([programPath], compilerOptions, basePath);
+
+	let schemas: Record<string, JSONSchema> = {};
```
Aenimus, line 1255: const

packages/sdk/src/configure/index.ts
```diff
@@ -1192,6 +1214,86 @@ const trimTrailingSlash = (url: string): string => {
...
+	const settings: TJS.PartialArgs = {
+		required: true,
+	};
+	// XXX: There's no way to silence warnings from TJS, override console.warn
+	const warn = console.warn;
+	console.warn = (_message?: any, ..._optionalParams: any[]) => {};
+	let generator = TJS.buildGenerator(program, settings);
```
Aenimus, line 1262: const

packages/sdk/src/configure/index.ts
```diff
@@ -1192,6 +1214,86 @@ const trimTrailingSlash = (url: string): string => {
...
+	// generator can be null if the program can't be compiled
+	if (!generator) {
+		console.warn = warn;
+		throw new Error('could not parse .ts operation files');
+	}
+	for (const op of operations) {
+		let schema = generator!.getSchemaForSymbol(responseTypeName(op));
```
Aenimus, line 1269: const

packages/sdk/src/configure/index.ts
fiam: That's right. If the cache loading fails (maybe the file ended up corrupted somehow), we'd rather fail silently and try to regenerate everything as if there was no cache entry.

packages/sdk/src/configure/index.ts
```diff
@@ -1192,6 +1214,86 @@ const trimTrailingSlash = (url: string): string => {
...
+	}
+	console.warn = warn;
+	const cached = JSON.stringify(schemas);
+	try {
+		const cacheDir = path.dirname(cachePath);
+		if (!fs.existsSync(cacheDir)) {
+			fs.mkdirSync(cacheDir, { recursive: true });
+		}
+		fs.writeFileSync(cachePath, cached, { encoding: 'utf-8' });
+	} catch {}
```
Aenimus, line 1283: empty catch?

packages/sdk/src/configure/index.ts
```diff
@@ -1236,7 +1338,7 @@ const resolveOperationsConfigurations = async (
 		customClaims,
 	});
 	const nodeJSOperations: GraphQLOperation[] = [];
-	if (loadedOperations.typescript_operation_files)
+	if (loadedOperations.typescript_operation_files) {
```
Aenimus, line 1341: Could invert the if an return early. It would mean you have two identical returns, but it would reduce the huge nested block. Your call.

packages/sdk/src/graphql/operations.ts
```diff
@@ -998,6 +997,9 @@ const typeSchema = (
 							break;
 						case 'InputObjectTypeDefinition':
 							const typeName = namedType.name;
+							if (!root.definitions) {
```
Aenimus, line 1000: Why can't we set this where it's declared?

packages/sdk/src/configure/index.ts
fiam: Let's move the nested block to a function

packages/sdk/src/configure/index.ts
fiam: I guess we can log a warning in this one. After all, if the cache can't be written to there's probably something wrong in the user's setup.

packages/sdk/src/configure/index.ts
devsergiy: @fiam put this explanation as a comment please

packages/sdk/src/openapibuilder/operations.ts
```diff
@@ -0,0 +1,50 @@
...
+// path syntax follows https://github.com/tidwall/sjson#path-syntax
+export function buildPath(
+	path: string[],
+	required: boolean,
+	obj: JSONSchema7Definition,
+	paths: JSONSchemaParameterPath[]
+) {
+	if (typeof obj === 'boolean') return;
+
+	if (obj.type === 'array') {
```
Aenimus, line 19: why typeof above and `.type` here?

And on line 30 you coerce obj.type. Shouldn't this be the first thing you check?

packages/sdk/src/graphql/operations.ts
fiam: That is eventually serialized into the OpenAPI spec. If we always initialize it as `{}` then it clutters the generated schema when it's empty. Lazily initializing it removes it from the output when it has no entries.

packages/sdk/src/configure/index.ts
```diff
@@ -1192,6 +1214,86 @@ const trimTrailingSlash = (url: string): string => {
...
+	let contents: string[] = ['import type { ExtractResponse } from "@wundergraph/sdk/operations";'];
+
+	for (const op of operations) {
+		const relativePath = `../operations/${op.PathName}`;
+		const name = functionTypeName(op);
+		contents.push(`import type ${name} from "${relativePath}";`);
+		contents.push(`export type ${responseTypeName(op)} = ExtractResponse<typeof ${name}>`);
+	}
+
+	const cachePath = path.join('cache', `ts.operationTypes.${objectHash(contents)}.json`);
```
StarpTech, line 1234: We want to get rid of the current cache in the next cycle. Is it so slow, or can we remove this?

packages/sdk/src/openapibuilder/__snapshots__/index.test.ts.snap
```diff
@@ -0,0 +1,136 @@
...
+    },
+    "/operations/GetRootObj2": {
+      "get": {
+        "operationId": "GetRootObj2",
+        "parameters": [],
+        "responses": {
+          "200": {
+            "content": {
+              "application/json": {
+                "schema": undefined,
```
devsergiy, line 85: is it correct that schema is undefined?

packages/sdk/src/openapibuilder/index.test.ts
```diff
@@ -0,0 +1,52 @@
...
+		VariablesSchema: {
+			type: 'object',
+			properties: {},
+			additionalProperties: false,
+			definitions: {},
+		},
+	},
+] as GraphQLOperation[];
+
+test('OpenAPI Builder', async () => {
```
StarpTech, line 44: Can we have some sort of table tests to understand what each fixture should represent? I can imagine this place here will grow with different use cases.

StarpTech (changes requested):
Great stuff!

packages/sdk/src/configure/index.ts
fiam: This works kinda separately from the rest of the cache, the only shared detail is that they use the same directory. `objectHash` should also be quite fast here, since we're hashing just an array of strings. Generating the schemas for the default testapp from scratch takes around 4s in my machine, so I think a cache is a nice improvement for this feature. Once we get rid of the one for introspection, we can decide if we keep this or if we refactor it into something else.

packages/sdk/src/openapibuilder/index.test.ts
fiam: I'll add more fine grained tests

packages/sdk/src/openapibuilder/__snapshots__/index.test.ts.snap
fiam: It is, the operation for the test has no response schema. For real operations, we copy the field as is so this ends up as undefined in the test.

packages/sdk/src/openapibuilder/index.ts
```diff
@@ -0,0 +1,284 @@
...
+
+const errorRefSchema: JSONSchema = {
+	$ref: `#/components/schemas/${invalidInputErrorName}`,
+};
+
+interface OpenAPIServer {
+	url: string;
+}
+
+interface OpenAPIInfo {
```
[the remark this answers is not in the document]
fiam: I took a look at a couple of types for OAS and they seemed to bring way more problems than they fixed, because we only support a small subset instead of the whole thing. I'll take a look at this one.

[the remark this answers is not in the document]
devsergiy: this is types only, no processing

packages/sdk/src/openapibuilder/index.ts
```diff
@@ -0,0 +1,284 @@
...
+			}
+			if (opPath) {
+				paths[`/operations/${op.PathName}`] = opPath;
+			}
+		}
+
+		const schemas: Record<string, JSONSchema> = {};
+		schemas[invalidInputErrorName] = errorSchema;
+
+		let spec: OpenAPISpec = {
```
devsergiy, line 268: I propose to also validate the resulting object

https://github.com/kogosoftwarellc/open-api/tree/master/packages/openapi-schema-validator

maybe it will be enough to do that in tests

packages/sdk/src/graphql/operations.ts
Aenimus: I don't think I understand. If it's falsy, you set it to an empty object. If it's truthy, it's already some form of object. Is the schema already generated before this? If so, what's the point of setting it at all? I'm really confused.

packages/sdk/src/openapibuilder/index.ts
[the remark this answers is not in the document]
fiam: These types are incomplete, [SchemaObject is missing $ref](https://github.com/kogosoftwarellc/open-api/blob/master/packages/openapi-types/index.ts#L143)

packages/sdk/src/openapibuilder/index.ts
fiam: This validator doesn't support `$ref` in schemas nor `additionalProperties: false`

packages/sdk/src/openapibuilder/index.test.ts
fiam: I've added a couple of tests to check a few more properties. They're not very extensive for now because we're not doing that much in this department yet, it's basically a dump of the operations list.

packages/sdk/src/configure/index.ts
StarpTech: ok

packages/sdk/src/configure/index.ts
```diff
@@ -145,6 +150,14 @@ export interface WunderGraphConfigApplicationConfig<
 	links?: LinkConfiguration;
 	security?: SecurityConfig;
 
+	/**
+	 * OpenAPI generator configuration
+	 */
+	openAPI?: {
```
StarpTech, line 156: I'd prefer `openApi`. Uppercase words are not a thing in javascript.

packages/sdk/src/configure/index.ts
```diff
@@ -1192,6 +1224,151 @@ const trimTrailingSlash = (url: string): string => {
...
 
+// typeScriptOperationsResponseSchemas generates the response schemas for all TypeScript
+// operations at once, since it's several times faster than generating them one by one
+const typeScriptOperationsResponseSchemas = (operations: GraphQLOperation[]) => {
+	const functionTypeName = (op: GraphQLOperation) => `function_${op.Name}`;
+	const responseTypeName = (op: GraphQLOperation) => `${functionTypeName(op)}_Response`;
+
+	const programFile = 'typescript_schema_generator.ts';
+
+	const contents: string[] = ['import type { ExtractResponse } from "@wundergraph/sdk/operations";'];
```
StarpTech, line 1235: For code generation, we have typescript templates `packages/sdk/src/codegen`

packages/sdk/src/configure/index.ts
```diff
@@ -1192,6 +1224,151 @@ const trimTrailingSlash = (url: string): string => {
...
+		try {
+			const cached = fs.readFileSync(cachePath, { encoding: 'utf-8' });
+			return JSON.parse(cached) as Record<string, JSONSchema>;
+		} catch {
+			// If the cache loading fails (maybe the file ended up corrupted somehow), we'd rather
+			// fail silently and try to regenerate everything as if there was no cache entry.
+		}
+	}
+
+	const basePath = path.join(process.env.WG_DIR_ABS!, 'generated');
```
StarpTech, line 1255: Reference to environment variables in functions is an anti-pattern. The path should be part of the interface, so it is easier to test and reason about.

packages/sdk/src/configure/index.ts
```diff
@@ -1192,6 +1224,151 @@ const trimTrailingSlash = (url: string): string => {
...
+	for (const op of operations) {
+		const responseSchema = schemas[op.Name];
+		if (responseSchema) {
+			op.ResponseSchema = responseSchema;
+		}
+	}
+};
+
+const loadNodeJsOperation = async (file: TypeScriptOperationFile) => {
+	const filePath = path.join(process.env.WG_DIR_ABS!, file.module_path);
```
StarpTech, line 1313: Same here.

packages/sdk/src/configure/index.ts
```diff
@@ -1192,6 +1224,151 @@ const trimTrailingSlash = (url: string): string => {
...
+	};
+
+	const program = TJS.getProgramFromFiles([programPath], compilerOptions, basePath);
+
+	const schemas: Record<string, JSONSchema> = {};
+	const settings: TJS.PartialArgs = {
+		required: true,
+	};
+	// XXX: There's no way to silence warnings from TJS, override console.warn
+	const warn = console.warn;
```
StarpTech, line 1273: Could you call it `orginalWarn`?

packages/sdk/src/configure/index.ts
```diff
@@ -1192,6 +1224,151 @@ const trimTrailingSlash = (url: string): string => {
...
+	console.warn = warn;
+	const cached = JSON.stringify(schemas);
+	try {
+		const cacheDir = path.dirname(cachePath);
+		if (!fs.existsSync(cacheDir)) {
+			fs.mkdirSync(cacheDir, { recursive: true });
+		}
+		fs.writeFileSync(cachePath, cached, { encoding: 'utf-8' });
+	} catch (e: any) {
+		logger.debug(`error storing cache entry: ${e}`);
```
StarpTech, line 1297: I think this is an error worth it.

StarpTech (changes requested):

docs-website/src/pages/docs/features/openapi-postman.md
```diff
@@ -0,0 +1,29 @@
...
+---
+
+## OpenAPI and Postman collection generation
+
+Out of the box, WunderGraph supports generating an OpenAPI Specification as well as a Postman collection to
+interact with your APIs. You can use them to test your API in Postman or to generate bindings to query your
+API from another language.
+
+After running `wunderctl generate`, usually via `npm build` or `npm build:wundergraph`, both an OpenAPI
+specification and a Postman collection will be produced inside `.wundergraph/generated`.
```
StarpTech, line 13: How can the user get access to the openAPI spec file? I'd name the concrete paths.

StarpTech (changes requested):

packages/sdk/src/configure/index.ts
fiam: We cannot route this through `GenerateCode()`, because retrieving the schemas for the TS operation responses requires code generation to run first.

packages/sdk/src/configure/index.ts
fiam: Renamed, but I don't think that topic is settled https://github.com/airbnb/javascript#naming--Acronyms-and-Initialisms 😅

packages/sdk/src/configure/index.ts
fiam: Changed

packages/sdk/src/configure/index.ts
fiam: Renamed

packages/sdk/src/configure/index.ts
fiam: I just moved that around, but I'll refactor it to read the environment variable once in `configureWunderGraphApplication()` and then pass it around

docs-website/src/pages/docs/features/openapi-postman.md
fiam: Good idea, I've added the exact filenames to the doc

packages/sdk/src/configure/index.ts
fiam: Changed

packages/sdk/src/configure/index.ts
fiam: Refactored into an argument

packages/sdk/src/configure/index.ts
StarpTech: This is a styleguide from airbnb. Most modules don't follow it, at least from my experience.

packages/sdk/src/configure/index.ts
```diff
@@ -1232,78 +1416,29 @@ const resolveOperationsConfigurations = async (
...
 				logger.info(`Skipping operation ${file.file_path} due to error: ${e.message}`);
 			}
 		}
+	}
 	return {
 		operations: [...graphQLOperations.operations, ...nodeJSOperations],
 	};
 };
 
-const loadAndApplyNodeJsOperationOverrides = async (operation: GraphQLOperation): Promise<GraphQLOperation> => {
```
StarpTech, line 1302: Could you add some comments what the methods are doing? With few tests its hard to understand the implementation. A short description how we convert to openapi would also helpful.

StarpTech (changes requested):

packages/sdk/src/configure/index.ts
```diff
@@ -12,6 +11,10 @@ import {
 	print,
 	visit,
 } from 'graphql';
+import { JSONSchema7 as JSONSchema } from 'json-schema';
+import objectHash from 'object-hash';
+import _ from 'lodash';
+import * as TJS from 'typescript-json-schema';
```
StarpTech, line 17: Please differentiate between type and imports. `*` should be avoided if possible.

StarpTech (changes requested):

packages/sdk/src/configure/index.ts
StarpTech: Alright :+1:

packages/sdk/src/configure/index.ts
fiam: Thanks for the heads up. I copied the import from the repo README and then I didn't think of it again. I've updated it to import only the two functions we need.

packages/sdk/src/configure/index.ts
fiam: I also found other JS style guides that suggested using the naming convention you mention, like Google's. Just wanted to point out that there are popular guides (AirBnB's has like 130k stars) that suggest something else, so I think we should explicitly pick one to follow and document it.

StarpTech (approved):
LGTM

packages/sdk/src/configure/index.ts
```diff
@@ -1188,7 +1226,153 @@ const trimTrailingSlash = (url: string): string => {
...
+		ignoreErrors: true,
+	};
+
+	const program = getProgramFromFiles([programPath], compilerOptions, basePath);
+
+	const schemas: Record<string, JSONSchema> = {};
+	const settings = {
+		required: true,
+	};
+	// XXX: There's no way to silence warnings from TJS, override console.warn
```
Pagebakers, line 1274: 🤦‍♂️

Pagebakers (approved):

devsergiy (approved):

PR #717, opened 2023-03-02
chore: remove new api introspection

devsergiy:
This reverts commit 0f070bb28b37c4cfe30b356f2c11c1f69d1bd903.

#### Checklist

- [ ] run `make test`


jensneuse (approved):

fiam (approved):
LGTM

msaktor:
Hi, can you shed a bit of light why was the `openApiV2` removed?

I am evaluating Wundergraph for our upcoming project and I like it a lot but the new OpenAPI introspection kinda worked with our specs whereas the `openApi` fails to parse the json document.

Are there plans to bring the new introspection back?
Thanks!

msaktor:
Thank you for a quick reply Jens! That sounds great!
Another thing I've stumbled upon is a bit inflexible way how to add custom headers = if I need to use HMAC, common part of the signature is the [endpoint path](https://learn.microsoft.com/en-us/azure/communication-services/tutorials/hmac-header-tutorial?pivots=programming-language-csharp#sign-an-http-request-with-c)

So it would be nice to have access to the request object = `headers: (builder, request) => ...`
```
const fooService = introspect.openApi({
	apiNamespace: 'cs',
	source: {
		kind: 'file',
		filePath: '../foo-service.json',
	},
	headers: (builder) => {
		const headers = signRequestWithHmac();
		return builder
			.addStaticHeader('X-TS', `${headers['X-TS']}`)
			.addStaticHeader('X-Key', headers['X-Key'])
			.addStaticHeader('X-Sig', headers['X-Sig'])
	},
})
```

jensneuse:
You can use an onOriginRequestHook, check for the host in the URL and add a signature if needed: https://docs.wundergraph.com/docs/wundergraph-server-ts-reference/on-origin-request-hook#__next

PR #722, opened 2023-03-06
fix: don't require a clientRequest for hooks

fiam:
Upload hooks don't have a client request


Pagebakers (approved):
Approved if this doesn't break using operations client inside webhooks.

fiam:
> Approved if this doesn't break using operations client inside webhooks.

I don't think it does, but if you have a testcase I can verify it.

devsergiy (approved):

PR #727, opened 2023-03-08
feat: new local cache implementation

fiam:
New cache implementation, using different policies depending on the data origin. This allows us to cache data coming from the filesystem forever, while caching data coming from the local network less aggressively than data coming from remote sources.

Fix also cache key derivation when the introspection object contains functions.

Move cache to node_modules/.cache/.wundergraph

Fix also updating the OAS when the TypeScript operations change. Cache key calculation was not considering all inputs correctly.

Fixes ENG-1454
Fixes ENG-1453
Fixes ENG-1452
Fixes ENG-1428
Fixes ENG-1384
Fixes ENG-781
Fixes ENG-581

Fixes ENG-1519


cli/helpers/cache.go
```diff
@@ -14,3 +15,28 @@ func GlobalWunderGraphCacheDir() (string, error) {
 	}
 	return filepath.Join(cacheDir, "wundergraph"), nil
 }
+
+// LocalWunderGraphCacheDir returns the absolute path to the
+// cache directory for the current WunderGraph app
+func LocalWunderGraphCacheDir(wgDir string) (string, error) {
```
StarpTech, line 21: The `.cache` folder is always placed on the root node_modules of a repository. So e.g. in an npm workspace the cache folder is at the root not at the wundergraph dir. Ideally, this path is configurable. similar to https://docs.astro.build/en/guides/integrations-guide/image/#configcachedir
image

StarpTech: This might also be useful. https://github.com/withastro/astro/blob/c44aa15534007c3ac320fa6e632c760ddb2ce2f3/packages/astro/src/core/create-vite.ts#L101

I'd checkout astro and turborepo and checkout how they cache behave with pnpm,npm and monorepo setups.

StarpTech (changes requested):

jensneuse:
@fiam can you add a description of the cache behaviour, edge cases, etc.. to the code so that it's clear how caching works without reading all the code?

cli/helpers/cache.go
fiam: Thanks for the clarification, I was not aware of that behavior. I've changed it to use the topmost `node_modules` instead of the first one found. Do you think we also need to make it configurable at this point?

fiam:
> @fiam can you add a description of the cache behaviour, edge cases, etc.. to the code so that it's clear how caching works without reading all the code?


I've added a few comments explaining the principles behind this implementation. PTAL and let me know if there's something else we can improve. Thanks!

cli/helpers/cache.go
StarpTech: > Do you think we also need to make it configurable at this point?

No, we can ship the first version without.

jensneuse (changes requested):
I have a few minor comments.

One thing I'd like to as is about the caching behavior of GraphQL APIs.
I think the "default" behavior should be the following:

- when we run "up" for the first time, introspect the origin directly, don't use the cache, only as fallback, e.g. after 10s of timeout
- during subsequent "reconfigure" runs, always use the cache AND keep the cache updated via polling every 5s (by default)

I think this is the perfect tradeoff between keeping the cache fresh, and reconfigure fast. WDYT?

cli/commands/up.go
```diff
@@ -86,11 +84,9 @@ var upCmd = &cobra.Command{
 			AbsWorkingDir: wunderGraphDir,
 			ScriptArgs:    []string{configOutFile},
 			Logger:        log,
-			ScriptEnv: append(helpers.CliEnv(rootFlags),
+			ScriptEnv: append(append(helpers.CliEnv(rootFlags), commonScriptEnv(wunderGraphDir)...),
```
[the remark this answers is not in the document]
fiam: Refactored into a helper function

packages/sdk/src/definition/introspection-cache.ts
```diff
@@ -7,14 +7,65 @@ import {
...
+ *
+ * Note that there's no way to find out wether a remote API has changed, so we make some assumptions to
+ * walk the fine line between responsiveness and correctness.
+ *
+ * There are 3 different types of sources when it comes to API introspection, and we use different caching
+ * strategies for these:
+ *
+ * ## Introspection from data coming from the filesystem
+ *
+ * There are some API types (like OpenAPI) where all we use for introspection comes from the local filesystem.
```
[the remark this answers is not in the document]
fiam: Almost. The derived cache key depends on the contents of the file, so we generate a different key when the file changes.

cli/helpers/cache.go
```diff
@@ -14,3 +15,33 @@ func GlobalWunderGraphCacheDir() (string, error) {
...
+	// Find the top node_modules directory from the WunderGraph
+	// directory by walking up
+	abs, err := filepath.Abs(wgDir)
+	if err != nil {
+		return "", err
+	}
+	const nodeModulesDirName = "node_modules"
```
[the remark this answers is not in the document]
fiam: It should be fine. The SDK side is written to work without a cache directory, acting as if there's no cache.

cli/commands/generate.go
```diff
@@ -57,15 +57,13 @@ var generateCmd = &cobra.Command{
 			ScriptArgs:    []string{configOutFile},
 			AbsWorkingDir: wunderGraphDir,
 			Logger:        log,
-			ScriptEnv: append(
-				helpers.CliEnv(rootFlags),
+			ScriptEnv: append(append(
```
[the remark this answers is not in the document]
fiam: Refactored into a function

packages/sdk/src/definition/database-introspection.ts
```diff
@@ -143,72 +146,120 @@ const applyNameSpaceToTypeNames = (typeNames: string[], namespace?: string): str
...
-export const introspectMongoDB = async (introspection: DatabaseIntrospection): Promise<MongoDBApi> =>
-	introspectWithCache(introspection, async (introspection: DatabaseIntrospection): Promise<MongoDBApi> => {
-		const { schema, fields, types, dataSources, interpolateVariableDefinitionAsJSON } = await introspectDatabase(
-			introspection,
-			mongodb,
-			5
-		);
-		return new MongoDBApi(schema, dataSources, fields, types, interpolateVariableDefinitionAsJSON);
-	});
+export const introspectPostgresql = introspectDatabaseWithCache(
```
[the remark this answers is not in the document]
fiam: I tried passing the constructor to `introspectDatabaseWithCache`, but it became `undefined` at runtime. I'll see if I can find another approach.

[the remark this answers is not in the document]
fiam: Fixed, PTAL

StarpTech:
@fiam lets have a chat in the morning. I'd like to understand the new behavior and how can use it in our builder.

cli/helpers/cache.go
StarpTech, line 1: The logic is not deterministic. Based on cwd a different cache folder can be found. The cache folder should look at the wundergraph application. Therefore, the top "node_modules" can be next to the wundergraph.config.ts or on repo root (, depending on the package manager.) or even worse, outside of the repository if the user hasn't installed the app but use a global wunderctl version. We need a consistent approach.

StarpTech (changes requested):

cli/helpers/cache.go
fiam: Changed it to use the local `node_modules`, as requested.

jensneuse (approved):

cli/commands/up.go
```diff
@@ -328,6 +350,9 @@ var upCmd = &cobra.Command{
 
 func init() {
 	upCmd.PersistentFlags().BoolVar(&upCmdPrettyLogging, "pretty-logging", true, "switches the logging to human readable format")
+	upCmd.PersistentFlags().IntVar(&defaultDataSourcePollingIntervalSeconds, "default-polling-interval", 5, "default polling interval for data sources")
+	upCmd.PersistentFlags().BoolVar(&disableCache, "no-cache", false, "disables local caches")
+	upCmd.PersistentFlags().BoolVar(&clearCache, "clear-cache", false, "clears local caches during startup")
```
StarpTech, line 355: `before startup`?

StarpTech (changes requested):

StarpTech (changes requested):

cli/helpers/cache.go
```diff
@@ -14,3 +15,14 @@ func GlobalWunderGraphCacheDir() (string, error) {
 	}
 	return filepath.Join(cacheDir, "wundergraph"), nil
 }
+
+// LocalWunderGraphCacheDir returns the absolute path to the
+// cache directory for the current WunderGraph app
+func LocalWunderGraphCacheDir(wgDir string) (string, error) {
```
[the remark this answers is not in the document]
StarpTech: The nearest package.json starting from the wundergraph.config.ts is the directory level of the node_modules.

cli/commands/up.go
fiam: Fixed, thanks!

StarpTech (approved):
LGTM

fiam:
Thanks for the reviews and suggestions!

PR #745, opened 2023-03-22
feat: add raw sql support

jensneuse:
This PR adds support for writing custom SQL within GraphQL Operations.
It's the foundation to also expose SQL to TypeScript Operations.
In a later PR, we'll be able to mount raw SQL handlers on the internal handler, so we can inject it easily into the TS Operations ctx.

image


packages/sdk/src/db/introspection.ts
```diff
@@ -199,6 +200,24 @@ export const cleanupPrismaSchema = (
 						currentTypeName = node.name.value;
 					}
 				});
+				if (node.name.value === 'Mutation') {
+					// we don't like the prisma schema using just JSON, so we rewrite the fields
+					return {
+						...node,
+						fields: [
+							...(node.fields || []).filter((f) => f.name.value !== 'queryRaw' && f.name.value !== 'executeRaw'),
```
Aenimus, line 208: I think it's more readable to have a new const define above 203.
`const nodeFields = node.fields || [];`

examples/nextjs-postgres-prisma/prisma.ts
```diff
@@ -0,0 +1,49 @@
+import { PrismaClient } from '@prisma/client';
```
Pagebakers, line 1: Looks good, but confused why this file is here?

Pagebakers (changes requested):

examples/nextjs-postgres-prisma/prisma.ts
jensneuse: testing purposes

pkg/datasources/database/datasource.go
```diff
@@ -595,57 +711,15 @@ func (p *Planner) printOperation() []byte {
...
-
-	err = asttransform.MergeDefinitionWithBaseSchema(definition)
-	if err != nil {
-		p.stopWithError("merging upstream schema with base schema failed")
-		return nil
+	if p.isQueryRaw || p.isQueryRawRow {
+		// we've added rawQuery and rawQueryRow to the Query type for better ergonomics,
+		// but prisma expects them to be on the Mutation type
+		// so we simply rewrite the AST if we have a rawQuery root field
+		p.upstreamOperation.OperationDefinitions[0].OperationType = ast.OperationTypeMutation
```
devsergiy, line 718: could we use ref from `EnterOperationDefinition`  instead of the hardcoded `0` index?

pkg/datasources/database/datasource.go
```diff
@@ -595,57 +711,15 @@ func (p *Planner) printOperation() []byte {
...
 	}
 
-	// When datasource is nested and definition query type do not contain operation field
```
devsergiy, line 633: could we still add an external error? why was it removed?

devsergiy (changes requested):

pnpm-lock.yaml
```diff
@@ -548,6 +548,17 @@ importers:
       ts-jest: 29.0.5_p6ekqnroyms5nhqbfxosryz7rm
       ts-node: 10.9.1_awa2wsr5thmg3i7jqycphctjfq
 
+  testapps/cloud-starter:
```
fiam, line 551: Please, move this out of your local source tree, this is not in the repo https://github.com/wundergraph/wundergraph/tree/main/testapps. We're adding and removing these lines in a bunch of PRs 😅

testapps/default/.wundergraph/operations/ExecuteRaw.graphql
```diff
@@ -0,0 +1,6 @@
+mutation {
+	users_post_executeRaw(
+		query: "insert into \"User\" (id,name,email) values (?,?,?)"
```
fiam, line 3: Is there an elegant way to get rid of the escaped quotes? They're a bit annoying to type. I know SQL doesn't require them, but I'm not sure about prisma.

jensneuse: no, it's only used for tables like "User", because User is a special term.

fiam: That's only for PostgreSQL. For SQLite/MySQL you don't need to quote it (unless it changed recently)

packages/sdk/src/db/introspection.ts
```diff
@@ -350,6 +369,197 @@ export const cleanupPrismaSchema = (
 	return print(cleaned);
 };
 
+// the rawRowSchema allows us to define the shape of a SQL response via GraphQL fields
```
StarpTech, line 372: It would be great if we could start to structure new code somehow with separate files or modules etc... We always start to extend the logic. The files are already big, and the code becomes unmaintainable without tests and good docs.

Pagebakers (changes requested):

packages/sdk/src/db/introspection.ts
StarpTech: It would be great to get some feedback here.

pkg/datasources/database/datasource.go
```diff
@@ -146,6 +150,31 @@ type fetchInput struct {
...
+	if !r.parentIsJson {
+		// when the parent is already rendering as a JSON, we don't need to wrap the child in quotes
+		// this happens when using a variable inside the parameters list, e.g.
+		// query($id: String!){rawQuery(query: "select foo from bar where id = $1", parameters: [$id])}
+		// in this case, there will be a 2 variable renderers, one that renders the list (as JSON)
+		// and a second one to inline render the value of $id
+		// which doesn't need to be quoted again
+		_, _ = out.Write([]byte(`\"`))
+	}
+	_, _ = out.Write([]byte(strings.ReplaceAll(string(data), `"`, `\\\"`)))
```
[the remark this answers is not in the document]
StarpTech: ?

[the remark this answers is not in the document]
jensneuse: I'll extend this should we run into problems

packages/sdk/src/db/introspection.ts
jensneuse: once we have more feedback on the feature, yes

pkg/datasources/database/datasource.go
```diff
@@ -146,6 +150,31 @@ type fetchInput struct {
...
 
+type RawJsonVariableRenderer struct {
+	parentIsJson bool
+}
+
+func (r *RawJsonVariableRenderer) GetKind() string {
+	return "raw_json"
+}
+
+func (r *RawJsonVariableRenderer) RenderVariable(ctx context.Context, data []byte, out io.Writer) error {
```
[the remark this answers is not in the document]
jensneuse: tests are in typescript, covering multiple use cases (inline variable & context variable)
we can later add unit tests

Pagebakers (approved):

StarpTech (approved):
LGTM

PR #750, opened 2023-03-23
fix: initialize introspection cache directory from the Go side

fiam:
Otherwise, we fail to watch it and we never try again until we reload
the development server.


StarpTech (approved):
LGTM

Pagebakers (approved):

PR #753, opened 2023-03-23
fix: data corruption when polling is enabled for federation sources

fiam:
Federation was querying its graphql components by using an introspection
function that goes through the cache, which in turn launched an introspection
in an interval when polling was enabled. Since federation called each of its
graphql sources when polling, this resulted in a new scheduled loop on
every call.

To fix this, expose a graphql introspection function that doesn't use the
cache and use it for introspecting federated sources. This shouldn't be
a problem because we do cache the result of the federation.

Also, add a few debug messages and handle a few more error cases when
introspecting graphql (as a result of the debugging journey).

Additioanlly, there was also a race condition in the cache files, which
caused problems with concurrent writes. To fix this, use atomic writes
when creating the cache files.


jensneuse (approved):

packages/sdk/src/definition/graphql-introspection.ts
```diff
@@ -89,137 +90,135 @@ export const graphqlIntrospectionCacheConfiguration = async (
...
-			}
-			if (introspection.introspection?.headers !== undefined) {
-				introspection.introspection?.headers(introspectionHeadersBuilder);
-			}
+	const headersBuilder = new HeadersBuilder();
+	const introspectionHeadersBuilder = new HeadersBuilder();
 
-			const headers = mapHeaders(headersBuilder);
-			const introspectionHeaders = resolveGraphqlIntrospectionHeaders(mapHeaders(introspectionHeadersBuilder));
+	if (introspection.headers !== undefined) {
```
Aenimus, line 96: Could be null?

packages/sdk/src/definition/graphql-introspection.ts
```diff
@@ -89,137 +90,135 @@ export const graphqlIntrospectionCacheConfiguration = async (
...
-				generateTypeConfigurationsForNamespace(schemaSDL, introspection.apiNamespace),
-				[],
-				introspection.customJSONScalars
-			);
+			serviceSDL = await fetchFederationServiceSDL(resolveVariable(introspection.url), introspectionHeaders, {
+				apiNamespace: introspection.apiNamespace,
+			});
 		}
+	}
+	const serviceDocumentNode = serviceSDL !== undefined ? parse(serviceSDL) : undefined;
```
Aenimus, line 135: could be null?

packages/sdk/src/definition/graphql-introspection.ts
```diff
@@ -89,137 +90,135 @@ export const graphqlIntrospectionCacheConfiguration = async (
...
+	const schemaDocumentNode = parse(schemaSDL);
+	const graphQLSchema = buildSchema(schemaSDL);
+	const { RootNodes, ChildNodes, Fields } = configuration(
+		schemaDocumentNode,
+		introspection.customJSONScalars,
+		serviceDocumentNode,
+		argumentReplacements
 	);
+	const subscriptionsEnabled = hasSubscriptions(upstreamSchema);
+	if (introspection.internal === true) {
```
Aenimus, line 145: `=== true` seems oddly verbose to me :D

packages/sdk/src/definition/graphql-introspection.ts
```diff
@@ -89,137 +90,135 @@ export const graphqlIntrospectionCacheConfiguration = async (
...
+					staticVariableContent: 'true',
+					environmentVariableName: '',
+					environmentVariableDefaultValue: '',
+					placeholderVariableName: '',
+				},
+			],
+		};
+	}
+	return new GraphQLApi(
+		applyNameSpaceToGraphQLSchema(schemaSDL, introspection.skipRenameRootFields || [], introspection.apiNamespace),
```
Aenimus, line 159: How about a const above 158
`const skipRenameRootFields = introspection.skipRenameRootFields || [];`

And you can use it again on line 210.

Or just set the introspection field directly, but perhaps you don't want to alter it.

packages/sdk/src/definition/graphql-introspection.ts
```diff
@@ -281,32 +291,42 @@ const introspectGraphQLAPI = async (
...
+		res = await Fetcher().post(url, data, opts);
 	} catch (e: any) {
 		throw new Error(
-			`introspection failed (url: ${introspection.url}, namespace: ${introspection.apiNamespace || ''}), error: ${
-				e.message
-			}`
+			`introspection failed (url: ${url}, namespace: ${introspection.apiNamespace || ''}), error: ${e.message}`
 		);
 	}
 	if (res === undefined) {
```
Aenimus, line 306: Could it be null?

fiam: I only moved this around, but you make a good point. I'll make it check for both.

packages/sdk/src/definition/graphql-introspection.ts
fiam: I only moved this around, but you make a good point. I'll make it check for both.

packages/sdk/src/definition/graphql-introspection.ts
fiam: This can make sense if it can be something else that's truthy, but not true

packages/sdk/src/definition/graphql-introspection.ts
fiam: Can't be null

packages/sdk/src/definition/graphql-introspection.ts
```diff
@@ -89,137 +90,135 @@ export const graphqlIntrospectionCacheConfiguration = async (
...
+					},
+					Subscription: {
+						Enabled: subscriptionsEnabled,
+						URL:
+							introspection.subscriptionsURL !== undefined
+								? mapInputVariable(introspection.subscriptionsURL)
+								: typeof introspection.url === 'string'
+								? mapInputVariable(introspection.url)
+								: mapInputVariable(''),
+						UseSSE: introspection.subscriptionsUseSSE !== undefined ? introspection.subscriptionsUseSSE : false,
```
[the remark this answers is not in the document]
fiam: It can't be `null`, but it looks much better as you suggest. I've changed it.

packages/sdk/src/definition/graphql-introspection.ts
fiam: Great idea, I've changed it

PR #754, opened 2023-03-23
chore: re-enable vite-swr example

fiam:


packages/sdk/src/configure/index.ts
```diff
@@ -1297,6 +1297,7 @@ const typeScriptOperationsResponseSchemas = async (wgDirAbs: string, operations:
 	}
 	for (const op of operations) {
 		const schema = generator.getSchemaForSymbol(responseTypeName(op));
+		console.log(`schema for ${op} is ${schema}`);
```
jensneuse, line 1300: Should this be here?

jensneuse (approved):

packages/sdk/src/configure/index.ts
fiam: It's only for debugging. I can't get this reproduced locally, so I added that print to try to figure it out. I'll remove it before merging.

fiam: I’ve removed it

PR #759, opened 2023-03-26
feat: custom error support for typescript operations

StarpTech:
#### Checklist

- [ ] run `make test`


packages/sdk/src/codegen/templates/typescript/index.ts
```diff
@@ -26,6 +27,7 @@ export class TypeScriptInputModels implements Template {
 					op.VariablesSchema,
 					op.Name + 'Input',
 					false,
+					op.ExecutionEngine == OperationExecutionEngine.ENGINE_NODEJS ? { pathName: op.PathName } : undefined,
```
jensneuse, line 30: This line is used multiple times. How about extracting it into a function? Also, should we use `===` instead of `==`?

jensneuse (approved):
LGTM

packages/sdk/src/codegen/templates/typescript/index.ts
StarpTech: Based on the types, it makes no difference, but I will add it. I see no benefit in putting this into a function.  The usage differs in various cases.

packages/sdk/src/client/client.ts
```diff
@@ -142,21 +147,40 @@ export class Client {
 	}
 
 	// Determines whether the body is unparseable, plain text, or json (and assumes an invalid input if json)
-	private async handleClientResponseError(response: globalThis.Response): Promise<ClientResponseError> {
+	private async handleClientResponseError(response: globalThis.Response): Promise<ResponseError> {
+		// In some cases, the server does not return JSON to communicate errors.
+		// TODO: We should align it to always return JSON
```
Pagebakers, line 152: Did you create an issue for this?

Pagebakers (approved):
We should make sure to mention this type has changed

 ClientResponseError to ResponseError

packages/sdk/src/client/client.ts
StarpTech: No, but I will do :+1:

StarpTech: Done

Pagebakers (changes requested):
Sorry, retracting my approval.

We need to document this in the docs as well.

1. A Error handling section in the ts operations reference https://docs.wundergraph.com/docs/typescript-operations-reference
2. A guide about custom errors under the writing operations section https://docs.wundergraph.com/docs/guides

Should we update the RPC spec too?
https://docs.wundergraph.com/docs/architecture/wundergraph-rpc-protocol-explained#__next

Pagebakers:
A few questions after testing.

1. Should statusCode be required? why not fallback to 400 by default for example?
2. When I throw a custom 400 error, it's always returned as an input validation error.
3. Code is not inferred like this, i'd expect the generic type will now be `CUSTOM_ERROR`, but it's still string.
```ts
class CustomError extends ResponseError {
	code = 'CUSTOM_ERROR';
}
```
4. What's the reason for having to initialize the error when passing it to errors in the operation config?
```ts
	errors: [new CustomError({statusCode: 400})],
```

Pagebakers:
Solid.js client doesnt seem to work either.

StarpTech:
Good points, I will address the changes in the docs. We have no change in the RPC protocol.

> When no errors are defined, and for graphql operations the error type is GraphQLError, I suppose we can improve the types here as well? Eg autocomplete the code field with InputValidationError, AuthorizationError and ResponseError?

That's by design. A graphql error is not typed only the response of TS operations. The user should use unions to express errors in a type-safe way.

StarpTech:
Good points, I will address the changes in the docs. We have no change in the RPC protocol.

> When no errors are defined, and for graphql operations the error type is GraphQLError, I suppose we can improve the types here as well? Eg autocomplete the code field with InputValidationError, AuthorizationError and ResponseError?

That's by design. A graphql error is not typed only the response of TS operations. The user should use unions to express errors in a type-safe way.

StarpTech:
Good points, I will address the changes in the docs. We have no change in the RPC protocol.

> When no errors are defined, and for graphql operations the error type is GraphQLError, I suppose we can improve the types here as well? Eg autocomplete the code field with InputValidationError, AuthorizationError and ResponseError?

That's by design. A graphql error is not typed only the response of TS operations. The user should use unions to express errors in a type-safe way.

Pagebakers:
> 

Yeah I understand, but also a ts operation without predefined errors has type GraphQLError

As a developer who works on the backend you might know the difference, but if you work on the frontend, or you the generated api client from an npm registry, you will not know the difference between ts/gql operations. Hence it'd make sense to make both work the same. At least with the built-in error codes.

fiam (approved):
We should propagate user defined errors to the OAS definition, but I can take care of that after this is merged. LGTM

packages/sdk/src/client/client.ts
```diff
@@ -142,21 +153,43 @@ export class Client {
...
+		// In some cases, the server does not return JSON to communicate errors.
+		// TODO: We should align it to always return JSON and in a consistent format.
+
+		if (response.status === 401) {
+			return new AuthorizationError();
+		}
+
 		const text = await response.text();
+
+		console.log('text', text);
```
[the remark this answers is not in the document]
StarpTech: Good catch. We should setup a linter.

StarpTech:
> We should propagate user defined errors to the OAS definition, but I can take care of that after this is merged. LGTM

Yes, that was the goal. I didn't want to introduce more in this PR.

StarpTech:
> Yeah I understand, but also a ts operation without predefined errors has type GraphQLError

That is an issue. From what client? I can't reproduce it inside our testing framework tests. Errors from GraphQL operations are `GraphQLError` but all errors have the same base type `ResponseError`

docs-website/src/pages/docs/typescript-operations-reference/index.md
```diff
@@ -92,3 +89,106 @@ This will return the following JSON:
...
+import { createOperation, z } from '../generated/wundergraph.factory';
+
+export class DividedByZero extends OperationError {
+  statusCode = 400;
+  code = 'DividedByZero' as const;
+  message = 'Cannot divide by zero';
+}
+
+export default createOperation.query({
+  errors: [NotFoundErr],
```
Pagebakers, line 143: Wrong error name

Pagebakers (changes requested):

Pagebakers (approved):
Approved, but please also write a Guide (in the guides section) on how to use custom error handling. 🙄😘

PR #761, opened 2023-03-27
fix: print compiler error messages during TS operation response schema generation

fiam:
Previously, we were silencing error messages from typescript-json-schema because
we have some fallback logic to handle some errors. This is a problem when
we get an error that cannot be recovered, because we were not printing what the
error actually was. Now we collect the error messages, apply deduplication and
we print them using console.error() before returning the failure.

### DIFF
--- a/packages/sdk/src/configure/index.ts
+++ b/packages/sdk/src/configure/index.ts
@@ -1228,6 +1228,11 @@ const trimTrailingSlash = (url: string): string => {
 	return url.endsWith('/') ? url.slice(0, -1) : url;
 };
 
+interface consoleMessage {
+	message?: any;
+	optionalParams: any[];
+}
+
 // typeScriptOperationsResponseSchemas generates the response schemas for all TypeScript
 // operations at once, since it's several times faster than generating them one by one
 const typeScriptOperationsResponseSchemas = async (wgDirAbs: string, operations: GraphQLOperation[]) => {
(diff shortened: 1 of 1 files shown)


packages/sdk/src/configure/index.ts
```diff
@@ -1277,7 +1282,16 @@ const typeScriptOperationsResponseSchemas = async (wgDirAbs: string, operations:
 	const originalWarn = console.warn;
 	const originalError = console.error;
 	console.warn = (_message?: any, ..._optionalParams: any[]) => {};
-	console.error = (_message?: any, ..._optionalParams: any[]) => {};
+	const errorMessages: consoleMessage[] = [];
+	console.error = (message?: any, ...optionalParams: any[]) => {
+		// We might get duplicates because we try to call the generator twice, once
```
StarpTech, line 1287: Can we avoid calling the generator twice?

StarpTech (changes requested):

packages/sdk/src/configure/index.ts
fiam: We call it once with the default TS configuration, but for some frameworks that ship with a base TS config (like expo), we need to fallback to use the project's tsconfig.json. Do you think it would be better to change the logic to make the `tsconfig.json` take priority and fallback to using the default config if we can't find it?

Aenimus (approved):

packages/sdk/src/configure/index.ts
```diff
@@ -1291,42 +1291,50 @@ const typeScriptOperationsResponseSchemas = async (wgDirAbs: string, operations:
...
-	if (!generator) {
-		// Try to fallback to generating a program from tsconfig.json
-		const tsConfigPath = path.join(path.dirname(wgDirAbs), 'tsconfig.json');
-		if (fs.existsSync(tsConfigPath)) {
-			const tsConfigProgram = programFromConfig(tsConfigPath);
-			if (tsConfigProgram) {
-				generator = buildGenerator(tsConfigProgram, settings);
-			}
+	// If we can find a tsconfig.json, use it
+	let tsConfigPath: string | undefined;
```
StarpTech, line 1302: Please put this in a separate function so we can test, and there is no need for `while(loop)` loop if we can iterate over fixed path.

StarpTech (changes requested):

packages/sdk/src/configure/index.ts
fiam: Moved to a file and added tests. PTAL

StarpTech (approved):
LGTM

packages/sdk/src/configure/index.ts
```diff
@@ -1258,6 +1258,23 @@ const trimTrailingSlash = (url: string): string => {
...
 };
 
+// return the first file named filename starting from cwd and walking
+// up the filesystem hierarchy until the root
+const findUp = (filename: string, cwd: string) => {
+	let cur = path.resolve(cwd);
+	const { root } = path.parse(cur);
+	while (true) {
+		const potentialMatch = path.join(cur, filename);
+		if (fs.existsSync(potentialMatch)) {
```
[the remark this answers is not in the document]
fiam: Good idea. I've made it stat() the FS entry and check wether it's a dir. PTAL.

PR #762, opened 2023-03-27
fix: support TS operations that never return

fiam:
We didn't support operations that didn't return, which can be useful as e.g.
stubs during development (i.e. an operation that always throws an "unimplemented" error).
To support these, make ResponseSchema an optional and update callers using this field
to deal with it being possible undefined.

If an operation has no defined response schema, we use a schema that describes an empty
JSON object for its OpenAPI definition. Update tests to reflect these changes too.


jensneuse:
I've replicated this, here's my result of calling a TS Operation that doesn't return:

```shell
jens@MacBook-Pro-2 ~ % curl 'http://localhost:9991/operations/test' -v
*   Trying 127.0.0.1:9991...
* Connected to localhost (127.0.0.1) port 9991 (#0)
> GET /operations/test HTTP/1.1
> Host: localhost:9991
> User-Agent: curl/7.79.1
> Accept: */*
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Content-Type: application/json
< Vary: Origin
< Date: Tue, 28 Mar 2023 06:58:51 GMT
< Content-Length: 2
<
* Connection #0 to host localhost left intact
{}%
```

It returns 200 OK with an empty JSON Object.
I suggest that for TS Operations that never return, we add a 200 entry with application/json as the response, with an object as the root response type with no properties and no additional properties. WDYT?

packages/sdk/src/configure/index.ts
```diff
@@ -1341,9 +1343,16 @@ const typeScriptOperationsResponseSchemas = async (wgDirAbs: string, operations:
...
 	for (const op of operations) {
-		const responseSchema = schemas[op.Name];
-		if (responseSchema) {
-			op.ResponseSchema = responseSchema;
+		const schema = schemas[op.Name];
+		if (schema) {
+			op.ResponseSchema = schemas[schema];
+		} else {
+			// For functions that don't return anything, we return an empty JSON object
+			op.ResponseSchema = {
```
StarpTech, line 1351: Does it mean the API returns an empty object?

StarpTech (changes requested):

packages/sdk/src/configure/index.ts
fiam: Jens tested it here https://github.com/wundergraph/wundergraph/pull/762#issuecomment-1486322403 - But I did the same test and I got the same result.

image

Do you think we should return something else?

StarpTech: Looks fine. I didn't know that we returned an empty object.

StarpTech (approved):
LGTM

PR #767, opened 2023-03-28
feat: enable CORS headers for requests that result in 404/405

fiam:
gorilla/mux (the router we use) only runs middlewares when the path and method
match a registered handler. If we used a single router, we could circumvent
this limitation with a custom 404 handler, but since we use subrouters this
workaround doesn't work. Instead, wrap the whole mux.Router instance in the
CORS middleware, which will always run the middleware before mux.Router even
has a chance to dispatch the request. This causes all requests to include the CORS
header even if there's no matching handler.


jensneuse (approved):
Good solution!

Aenimus (approved):

PR #771, opened 2023-03-28
feat: add beta omnigraph backed open api data

devsergiy:
#### Checklist

- [ ] run `make test`


packages/sdk/src/definition/index.ts
```diff
@@ -22,7 +22,7 @@ import { InputVariable, mapInputVariable } from '../configure/variables';
 import { introspectGraphqlWithCache } from './graphql-introspection';
 import { introspectFederation } from './federation-introspection';
 import { IGraphqlIntrospectionHeadersBuilder, IHeadersBuilder } from './headers-builder';
-import { openApi } from './openapi-introspection';
+import { openApi, OpenAPIIntrospectionNew, openApiNew } from './openapi-introspection';
```
Pagebakers, line 25: Can we make this openApiV2?

Pagebakers (changes requested):

packages/sdk/src/definition/index.ts
Pagebakers: One other issue i found is that `id` is not required but it will throw an error if no id is supplied to the introspection.

devsergiy: I have made an ID required and renamed to v2

StarpTech (approved):
LGTM

jensneuse (approved):

PR #776, opened 2023-03-30
fix: make TS operation response schema generation more forgiving

fiam:
- Load only the .ts file with the response types instead of the whole
program
- Ignore non-fatal errors during schema generation
- Correctly set the schema in updateTypeScriptOperationsResponseSchemas()
before passing it to the OpenAPI builder (broken in #762)

### DIFF
--- a/packages/sdk/src/configure/index.ts
+++ b/packages/sdk/src/configure/index.ts
@@ -1284,7 +1284,7 @@ const typeScriptOperationsResponseSchemas = async (wgDirAbs: string, operations:
 	const cacheKey = `ts.operationTypes.${objectHash([contents, operationHashes])}`;
 	const cachedData = await cache.getJSON(cacheKey);
 	if (cachedData) {
-		return cachedData;
+		return cachedData as Record<string, JSONSchema>;
 	}
 
 	const basePath = path.join(wgDirAbs, 'generated');
@@ -1294,6 +1294,7 @@ const typeScriptOperationsResponseSchemas = async (wgDirAbs: string, operations:
 
 	const settings = {
 		required: true,
+		ignoreErrors: true,
 	};
 
 	// XXX: There's no way to silence warnings from TJS, override console.warn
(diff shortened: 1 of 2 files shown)


jensneuse:
I don't see any change in tests. Is this code change already covered?

fiam:
> I don't see any change in tests. Is this code change already covered?

I was thinking about adding a test that loads the generated OpenAPI self and test that it has the properties that it should have. That way we ensure that the response schema generation works correctly. WDYT?

packages/sdk/src/configure/index.ts
```diff
@@ -1303,7 +1304,7 @@ const typeScriptOperationsResponseSchemas = async (wgDirAbs: string, operations:
 	const tsConfigPath = await findUp('tsconfig.json', wgDirAbs);
 	let generator: JsonSchemaGenerator | null = null;
 	if (tsConfigPath) {
-		const tsConfigProgram = programFromConfig(tsConfigPath);
+		const tsConfigProgram = programFromConfig(tsConfigPath, [programPath]);
 		if (tsConfigProgram) {
```
StarpTech, line 1308: What's the exact difference from before? Does it also respect imported files?

StarpTech (changes requested):

packages/sdk/src/configure/index.ts
StarpTech: :+1:

StarpTech (approved):
LGTM

fiam:
@jensneuse I've managed to test this inside the builder by using the whole monorepo as the WG project, with a wunderctl binary built for Linux placed at the root, and then overriding the build command to call `pnpm -r build` filtering just the nextjs example. It does work.

Note that it's not exactly the same as what runs in cloud, since the template uses npm there, but I can't make it work with both npm and the local copy of the SDK inside the builder.

Slickstef11 (approved):

PR #782, opened 2023-04-01
fix: multiple fixes in authentication hooks

fiam:
- Don't ignore error on postAuthentication with tokens
- Fix revalidation from Client when using token based authentication
- Call revalidation hook appropriately
- Unify error messages returned to callers from hooks that can deny auth
- Add tests for authentication hooks

### DIFF
--- a/pkg/authentication/authentication.go
+++ b/pkg/authentication/authentication.go
@@ -132,10 +132,18 @@ func (u *UserLoader) userFromToken(token *jwt.Token, cfg *UserLoadConfig, user *
 	tempUser.ProviderName = "token"
 	tempUser.ProviderID = issuer
 	tempUser.AccessToken = tryParseJWT(token.Raw)
-	u.hooks.PostAuthentication(ctx, tempUser)
+	if err := u.hooks.PostAuthentication(ctx, tempUser); err != nil {
+		return err
+	}
 	tempUser, err := u.hooks.MutatingPostAuthentication(ctx, tempUser)
 	if err != nil {
-		return fmt.Errorf("access denied: %w", err)
+		return err
+	}
+	if revalidate {
+		tempUser, err = u.hooks.RevalidateAuthentication(ctx, tempUser)
+		if err != nil {
+			return err
+		}
 	}
 	*user = *tempUser
 	if cfg.cacheTtlSeconds > 0 {
@@ -316,7 +324,7 @@ func (u *User) Load(loader *UserLoader, r *http.Request) error {
 			return fmt.Errorf("invalid authorization Header")
 		}
 		tokenString := strings.TrimPrefix(authorizationHeader, "Bearer ")
-		revalidate := r.URL.Query().Get("revalidate") == "true"
+		revalidate := r.URL.Query().Has("revalidate")
 		for _, config := range loader.userLoadConfigs {
 			keyFunc := config.Keyfunc()
 			token, err := jwt.Parse(tokenString, keyFunc)
--- a/testapps/auth/.wundergraph/wundergraph.server.ts
+++ b/testapps/auth/.wundergraph/wundergraph.server.ts
@@ -8,7 +8,7 @@ import {
 	GraphQLInputObjectType,
 	GraphQLNonNull,
 } from 'graphql';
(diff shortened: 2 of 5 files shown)


pkg/authentication/authentication.go
```diff
@@ -132,10 +132,18 @@ func (u *UserLoader) userFromToken(token *jwt.Token, cfg *UserLoadConfig, user *
 	tempUser.ProviderName = "token"
 	tempUser.ProviderID = issuer
 	tempUser.AccessToken = tryParseJWT(token.Raw)
-	u.hooks.PostAuthentication(ctx, tempUser)
+	if err := u.hooks.PostAuthentication(ctx, tempUser); err != nil {
+		return err
```
StarpTech, line 136: Do we ensure 401?

pkg/authentication/authentication.go
```diff
@@ -316,7 +324,7 @@ func (u *User) Load(loader *UserLoader, r *http.Request) error {
 			return fmt.Errorf("invalid authorization Header")
 		}
 		tokenString := strings.TrimPrefix(authorizationHeader, "Bearer ")
-		revalidate := r.URL.Query().Get("revalidate") == "true"
+		revalidate := r.URL.Query().Has("revalidate")
```
StarpTech, line 327: Carefully, it means `revalidate=false` is considered true.

testapps/auth/.wundergraph/wundergraph.server.ts
```diff
@@ -8,14 +8,54 @@ import {
...
 
 export default configureWunderGraphServer<HooksConfig, InternalClient>(() => ({
 	hooks: {
 		queries: {},
 		mutations: {},
+		authentication: {
+			postAuthentication: async (hook) => {
+				console.log(`postAuthentication: ${JSON.stringify(hook.user)}`);
+			},
+			mutatingPostAuthentication: async (hook) => {
```
StarpTech, line 23: In the long term, we should not write integration tests against testapps that have been built to change. We need to manage a test suite. Maybe you have some ideas.

StarpTech (changes requested):

pkg/authentication/authentication.go
fiam: Not yet, we have an issue opened to change the returned response when sign-in fails or when there's no signed in user, but we haven't decided how to handle that yet.

testapps/auth/.wundergraph/wundergraph.server.ts
jensneuse: I agree. We should have a dedicated directory, I suggest "testsuite" that's actually used to properly test WG.

jensneuse (approved):

pkg/authentication/authentication.go
fiam: That's what we're doing already in a couple of places. See the [client side](https://github.com/wundergraph/wundergraph/blob/65c50fcf205b5a25de7df72eb559f6ad5b66f7c9/packages/sdk/src/client/client.ts#L338) and the [server side](https://github.com/wundergraph/wundergraph/blob/65c50fcf205b5a25de7df72eb559f6ad5b66f7c9/pkg/authentication/authentication.go#L722) - Since we had two places that considered the presence of the query parameter to mean `true`, I changed the one that was different.

testapps/auth/.wundergraph/wundergraph.server.ts
fiam: Besides a dedicated directory, it would be quite beneficial to allow overriding the configuration at runtime. That way we can pass configuration overrides to `createTestServer()` and test a lot of things without needing multiple applications. The main blocker we have right now is that a lot of things can't be toggled on and off (e.g. hooks).

pkg/authentication/authentication.go
StarpTech: ok

testapps/auth/.wundergraph/wundergraph.server.ts
StarpTech: For now, we can manage multiple fixtures (config files) to emulate it.

StarpTech (approved):
LGTM

PR #783, opened 2023-04-02
feat: implement typescript orm

timkendall:
## Motivation and Context

This PR is the first pass implementation of the TypeScript ORM proposed in  #597. It is still very much a work-in-progress but basic functionality and the integration with WunderGraph works. Opening primarily for the WunderGraph team to provide feedback on some of the WunderGraph-specific integration details that should be addressed before merge.

Left to do (before merge to allow this PR to mostly stand on it's own):
- [x] Add developer documentation (i.e architecture, contributing, etc.)
- [x] Utilize internal GraphQL handler (ensure subscriptions work)
- [x] Add some minimal docs for community
- [x] Fix array result type-inference bug (array results are getting inferred as `never[]` right now)
- [x] Union type support (e.g `on` method)
- [x] Selecting on interface implementations support
- [x] Subscription operation execution
- [x] Document default and filtered selections
- [x] Transpile generated orm with `wunderctl` 
- [x] Default all scalars selection on nested objects
- [x] Cancel subscription on client disconnect 
- [x] Only rebuild ORM on API schema changes

Potential work to be done in subsequent PRs:
- Configurable serializers/deserializers for custom scalars
- Selecting common interface fields by default
- Support non-namespaced APIs
- Additional refactoring (specifically the TypeScript API schema representation), tests, and bug fixes

#### Checklist

- [ ] run `make test`


packages/sdk/src/configure/index.ts
```diff
@@ -671,9 +672,10 @@ const resolveApplication = async (
 	hooks?: HooksConfiguration
 ): Promise<ResolvedApplication> => {
 	const resolvedApis = await Promise.all(apis);
-	const merged = mergeApis(roles, customClaims, ...resolvedApis);
+	const merged = mergeApis(roles, customClaims, undefined, ...resolvedApis);
 	const s3Configurations = s3?.map((config) => resolveUploadConfiguration(config, hooks)) || [];
 	return {
+		Apis: resolvedApis,
```
timkendall, line 678: I added this to get access to the individual namespaced APIs in-order-to simplify code-generation. We generate a separate un-namespaced TypeScript representation of each API's GraphQL schema (this is primarily in-consideration of TS performance but also more ergonomic in general).

packages/sdk/src/definition/index.ts
```diff
@@ -77,14 +77,16 @@ export class Api<T = ApiType> implements RenameTypes, RenameTypeFields {
...
+		customJsonScalars?: string[],
+		namespace?: string
 	) {
 		this.Schema = schema;
 		this.DataSources = dataSources;
 		this.Fields = fields;
 		this.Types = types;
 		this.interpolateVariableDefinitionAsJSON = interpolateVariableDefinitionAsJSON;
 		this.CustomJsonScalars = customJsonScalars;
+		this.Namespace = namespace
```
timkendall, line 89: See above comment. i.e need access to each API's namespace for efficient/simple code-generation.

packages/sdk/src/server/server.ts
```diff
@@ -268,6 +273,7 @@ export const createServer = async ({
 				operations: operationsConfig.typescript_operation_files,
 				internalClientFactory: clientFactory,
 				nodeURL,
+				orm: ormModule ? ormModule.orm : null,
```
timkendall, line 276: Not safe. Are you guys fine requiring the orm be generated (i.e we can throw above if it is not)? cc. @jensneuse

pkg/operations/operations.go
```diff
@@ -103,16 +103,19 @@ func EnsureWunderGraphFactoryTS(wunderGraphDir string) error {
 	return os.WriteFile(path, []byte(wunderGraphFactoryTSContent), os.ModePerm)
 }
 
+// @todo we need to ensure the ORM is generated for every
```
timkendall, line 106: Same question, are you guys fine making the orm a required component? @jensneuse

testapps/simple/.wundergraph/operations/users/getWeather.ts
```diff
@@ -1,9 +1,13 @@
 import { createOperation, z } from '../../generated/wundergraph.factory';
 import { NotFoundErr, RateLimitErr } from '../../../errors/errors';
 
+// @todo should we expose API utilties (e.g this enum) on the ORM 
+// client itself (to avoid the need for static imports)
+import { Weather } from '../../generated/orm'
```
timkendall, line 6: I think this primarily only matters for enums used in operation variables. Some options:

1. Rely on static imports (not very ergonomic)
2. Expose enum references on handler context (maybe on `graph`?)
3. Replace enum types w/string literal unions (still get editor auto-complete, maybe the best option?)

testapps/simple/.wundergraph/operations/users/getWeather.ts
```diff
@@ -12,14 +16,43 @@ export default createOperation.query({
 			},
 		});
 
+		// GraphQL ORM operation
+		// @note `select` will need to have a max parameters of "10"
```
timkendall, line 20: I think we should iterate on this API. The example below really stress-tests the current API. 

While tweaking `select` to support selecting all nested scalar fields by default will help this example (e.g `select('weather', 'weather.summary', 'weather.temperature', 'weather.wind')`) it still feels unnecessarily verbose.

Perhaps we can decorate the GraphQL API schema w/additional metadata to let the ORM infer what objects are values (automatically selected) vs. entities (require explicit selection)

testapps/simple/.wundergraph/operations/users/getWeather.ts
timkendall: Made the change to string literals as it seemed the obvious solution. Happy to re-visit if needed.

testapps/simple/.wundergraph/operations/users/getWeather.ts
jensneuse: Would it be possible to do `weather.summary.*` & `weather.temperature.*` to include all scalars on the current object, or `weather.**` to include all fields exhaustively? (except _join to prevent recursion)

packages/sdk/src/configure/index.ts
jensneuse: I think this makes sense.

packages/sdk/src/definition/index.ts
jensneuse: yup

packages/sdk/src/server/server.ts
jensneuse: To me it sounds like the server should not start if the orm is not generated. Opinions?

packages/sdk/src/server/server.ts
```diff
@@ -255,10 +256,14 @@ export const createServer = async ({
 
 	const operationsFilePath = path.join(wundergraphDir, 'generated', 'wundergraph.operations.json');
 	const operationsFileExists = fs.existsSync(operationsFilePath);
+
 	if (operationsFileExists) {
 		const operationsConfigFile = fs.readFileSync(operationsFilePath, 'utf-8');
 		const operationsConfig = JSON.parse(operationsConfigFile) as LoadOperationsOutput;
 
+		const ormModulePath = buildOrm(wundergraphDir)
```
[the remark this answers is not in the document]
jensneuse: @StarpTech is this the way to go? We already use esbuild internally from Go

testapps/simple/.wundergraph/operations/users/getWeather.ts
timkendall: Yeah I was thinking that could be the way to go. Nice and explicit. Exhaustive select-all may be tricky (and a nice-to-have IMO) but I think the former are do-able. Will spike.

StarpTech:
> Utilize internal GraphQL handler (doesn't exist yet, cc. @jensneuse )

Hi @timkendall we're currently working on the internal graphql handler. In Middle of the next week, we should have a working version.

> Add developer documentation (i.e architecture, contributing, etc.)

In the meantime, it would be valuable, before we merge this PR that we have some docs about the architecture.

timkendall:
@StarpTech Yep will do.

packages/sdk/src/server/server.ts
```diff
@@ -296,10 +297,14 @@ export const createServer = async ({
 
 	const operationsFilePath = path.join(wundergraphDir, 'generated', 'wundergraph.operations.json');
 	const operationsFileExists = fs.existsSync(operationsFilePath);
+
 	if (operationsFileExists) {
 		const operationsConfigFile = fs.readFileSync(operationsFilePath, 'utf-8');
 		const operationsConfig = JSON.parse(operationsConfigFile) as LoadOperationsOutput;
 
+		const ormModulePath = buildOrm(wundergraphDir);
```
timkendall, line 305: @StarpTech acceptable for now (see Jens original [comment](https://github.com/wundergraph/wundergraph/pull/783#discussion_r1157103488); outdated due to rebase)?

timkendall:
Added some minimal documentation here:
- [Architecture doc](https://github.com/timkendall/wundergraph/blob/feat/orm/packages/orm/docs/architecture.md)
- Community docs
   - [`/docs/features/typescript-orm`](https://github.com/timkendall/wundergraph/blob/feat/orm/docs-website/src/pages/docs/features/typescript-orm.md)
   - [`/docs/typescript-operations-reference/accessing-data-sources-from-operations`](https://github.com/timkendall/wundergraph/blob/feat/orm/docs-website/src/pages/docs/typescript-operations-reference/accessing-data-sources-from-operations.md)

StarpTech (changes requested):

StarpTech:
@timkendall please rebase :smile:

packages/sdk/src/codegen/templates/typescript/orm.template.ts
```diff
@@ -0,0 +1,46 @@
...
+  {{/each}}
+} from './schemas'
+
+export * from './schemas'
+
+// base URL of the WunderGraph GraphQL API
+const BASE_URL = "{{ baseUrl }}"
+
+const APIs = {
+  {{#each namespaces}}
```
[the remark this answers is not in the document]
timkendall: Yeah this was cruft, good callout. We can improve slightly right now (implemented, see changes) but TypeScript still requires us to statically reference the `Schema` interface and `SCHEMA` constant (so we can't do `schema.Schema`).

(this could be further simplified by sharing the schema representation type between comp time and runtime like is done with operations)

packages/sdk/src/server/server.ts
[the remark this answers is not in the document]
timkendall: Not sure if anything does, likely just my ignorance here. As long as we've ensured that the ORM code-generator has run we can transpire within the Go-implemented `generate` command.

pkg/operations/operations.go
timkendall: Yeah idk. My assumption is that it'd be fairly negligible impact in either context (could be wrong, haven't quantified).

Regardless, we need to ensure that we're generating a valid reference (to `orm`) here; always generating the ORM is just the first solution that came to mind for me. Let me know how you guys want to handle 👍

StarpTech:
@timkendall, the internal graphql handler has been merged.

timkendall:
@StarpTech Yep saw, update. Is [this](https://github.com/timkendall/wundergraph/blob/feat/orm/packages/sdk/src/codegen/templates/typescript/orm.ts#L64) the appropriate way to reference the new internal endpoint (tested locally and it works)?

StarpTech:
> Yep saw, updated. Is [this](https://github.com/timkendall/wundergraph/blob/feat/orm/packages/sdk/src/codegen/templates/typescript/orm.ts#L64) the appropriate way to reference the new internal endpoint (tested locally and it works)?

Looks good!

timkendall:
@StarpTech @jensneuse is moving the ORM transpilation to build-time a blocker for this being merged? If so I can focus on solutions for that. In the mean-time I will continue fixing bugs and implementing small improvements to the functionality.

Pagebakers:
Wanted to give this a test run in the nextjs testapp, but not all code generation seems to be up to date yet.

Updating `wundergraph.factory.ts` to the following works. 
```ts
// Code generated by wunderctl. DO NOT EDIT.

import type {InternalClient} from "./wundergraph.internal.client";
import {Queries,Mutations,Subscriptions} from "./wundergraph.internal.operations.client";
import type {Role} from "./wundergraph.server";
import type {CustomClaims} from "./claims";
import {createOperationFactory} from "@wundergraph/sdk/operations";
import {orm} from './orm'

export {z, AuthorizationError} from "@wundergraph/sdk/operations";
export const createOperation = createOperationFactory<InternalClient, Role, CustomClaims, Queries, Mutations, Subscriptions, typeof orm>();
```

@wundergraph/orm needs to be a dependency of the project right now, next to @wundergraph/sdk. Ideally sdk is the only requirement.

We use es5 target in a lot of examples, which is outdated and incompatible with the generated orm, setting the target to 2019 fixes (same target as the sdk) this.

The weather API doesn't work, but the only error message I get is `Client error` so not sure what's wrong.

Really great work! Can't wait to use this.

StarpTech:
> @StarpTech @jensneuse is moving the ORM transpilation to build-time a blocker for this being merged? 

I'd like to give it at least a spin before we give up on this. But yes, if the effort is too high, we can merge it as it is.

packages/orm/src/fragment-definition-builder.ts
```diff
@@ -43,12 +73,88 @@ export class FragmentDefinitionBuilder<Config extends FragmentDefinitionBuilderC
...
+				leave: () => {
+					selectionPath.pop();
+				},
+			},
+		});
+
+		this.selection = visit(this.selection, visitor);
+
+		// @note needed to prevent TypeScript from entering infinite type inference recursion
+		return this as any;
```
StarpTech, line 156: Curious if `unknown` works here too.

timkendall: It doesn't

timkendall:
@StarpTech I think we actually should be able to bundle/transpile the generated ORM after we finish executing the config in Node ([here](https://github.com/timkendall/wundergraph/blob/feat/orm/cli/commands/generate.go#L147)). I'll spike this out shortly.

(confused myself there for a minute ;)

timkendall:
@StarpTech alright there we go (scrappy but works, let me know what should be adjusted please 🙏 )

testapps/default/.wundergraph/operations/federation/CustomPriceUpdates.ts
```diff
@@ -1,21 +1,26 @@
...
-			operationName: 'federation/PriceUpdates',
-			subscribeOnce: true,
-		});
-		for await (const update of updates) {
-			const data = update.data?.federated_updatedPrice ?? undefined;
-			if (!data) {
-				continue;
-			}
+	handler: async function* ({ operations, graph }) {
+		// @bug this seems to cause the WunderNode (Go) server to crash
```
timkendall, line 5: I mentioned this to @Pagebakers but running this operation from `main` results in a segfault in the Go server it seems (cc. @StarpTech )

jensneuse: I've tried this using "go debug", no issues.
However, I found another problem.
We're not canceling the subscription when the client request disconnects.
On the operations client, we've exposed a cancelSubscriptions func to the fastify server to cancel all subscriptions on client disconnect. We could use a similar pattern if there's no alternative:
https://github.com/wundergraph/wundergraph/blob/1e3cf6094c23e1c5830cdef23db8399e5e259319/packages/sdk/src/server/operations-client.ts#L72

timkendall: Ah good catch, will address 👍

jensneuse:
@timkendall I found another issue

Given the following operation:

```typescript
const capsule = await graph.from('spacex')
			.query('capsule')
			.select('id','status','type','dragon.id','dragon.active')
			.exec()
```

The response object will have the path `dragon.id` and `draong.active`.

If I change the operation to this:

```typescript
const capsule = await graph.from('spacex')
			.query('capsule')
			.select('id','status','type','dragon','dragon.id','dragon.active')
			.exec()
```

Now dragon is an empty object on the response shape.

I think in both cases, the response shape should be the same.
Additionally, I'm not sure what's the value of selecting 'dragons' at all.

That said, I had an idea that we previously named `dragons.*`.
If a user wants to select all scalar fields on dragons, they could just use the following operation:

```typescript
const capsule = await graph.from('spacex')
			.query('capsule')
			.select('id','status','type','dragon')
			.exec()
```

WDYT`

StarpTech:
@timkendall I haven't forgotten you. I'll revisit the PR tomorrow :smile:

timkendall:
@StarpTech All good :) @Pagebakers and I were chatting and I think we'll need to further optimize the ORM codegen piece. What I have implemented now runs the transpilation/bundling at build time in Go properly, but we're still re-generating the ORM every time and operation file changes. This is very noticeable (~10+ seconds on my M1 w/the default test app). Do we have any support for caching codegen plugins already (cc. @jensneuse )?

jensneuse:
> @StarpTech All good :) @Pagebakers and I were chatting and I think we'll need to further optimize the ORM codegen piece. What I have implemented now runs the transpilation/bundling at build time in Go properly, but we're still re-generating the ORM every time and operation file changes. This is very noticeable (~10+ seconds on my M1 w/the default test app). Do we have any support for caching codegen plugins already (cc. @jensneuse )?

The code gen should actually never do anything for the ORM when operations change, as the ORM only really depends on the virtual graph. We don't yet have it, but we need a system to cache results of certain steps. E.g. if the outcome of generating the virtual graph doesn't change, there's really no need to re-generate the ORM.

timkendall:
@jensneuse Implemented the tweak for auto-selecting all scalars on object types (only one level deep). Solves the majority use-case maybe? Let me know what you guys think.

Relation/nested complex type selection still needs improvement IMO. I need to think on it some more. Would be nice to support N-level deep selections but something like this feels ambiguous `select('dragon', 'dragon.diameter', 'dragon.heat_shield')` 🤔

jensneuse:
> @jensneuse Implemented the tweak for auto-selecting all scalars on object types (only one level deep). Solves the majority use-case maybe? Let me know what you guys think.
> 
> Relation/nested complex type selection still needs improvement IMO. I need to think on it some more. Would be nice to support N-level deep selections but something like this feels ambiguous `select('dragon', 'dragon.diameter', 'dragon.heat_shield')` 🤔

When you select `dragon.diameter`, can it auto select `dragon`?

timkendall:
@jensneuse Exactly. I can whip something up for the ORM. 

Maybe a longer term solution could be to integrate a first-class build system like Turbopack or something to optimize the codegen steps optimally (maybe overkill...)? i.e WunderCtl could generate the build system config and execute that in the codegen step.

timkendall:
@jensneuse I suppose so (will have to see if this complicates the implementation too much...). Still seems ambiguous to me (i.e are we selecting _only_ `dragon.diameter`?) but maybe this isn't an issue w/IntelliSense

jensneuse:
> @jensneuse I suppose so (will have to see if this complicates the implementation too much...). Still seems ambiguous to me (i.e are we selecting _only_ `dragon.diameter`?) but maybe this isn't an issue w/IntelliSense

I thought about it a bit more. dragon is an object, so it doesn't really make sense to select it except you'd want a "select all". So either select dragon and disable all other fields, or select individual fields.

packages/sdk/src/server/plugins/functions.ts
```diff
@@ -58,24 +71,49 @@ const FastifyFunctionsPlugin: FastifyPluginAsync<FastifyFunctionsOptions> = asyn
...
+										reply.hijack();
+
+										const subscribeOnce = request.headers['x-wg-subscribe-once'] === 'true';
+										const gen = implementation.subscriptionHandler(ctx);
+
+										reply.raw.once('close', async () => {
+											// call return on the operation's `AsyncGenerator`
+											await gen.return(0);
+											// Cancel operations created by the ORM for this operation
+											asyncContext.ormOperationControllers.forEach((controller) => controller.abort());
```
jensneuse, line 98: @timkendall Is this async through the forEach or will the forEach return after all controllers are aborted?

jensneuse:
@timkendall regarding cancelling subscriptions, here's how you can do it without globals

Make a copy of the ORM (operations client) with an empty array of subscriptions similar to this:
https://github.com/timkendall/wundergraph/blob/feat/orm/packages/sdk/src/server/plugins/functions.ts#L58

Then, on unsubscribe call the cancel func:
https://github.com/timkendall/wundergraph/blob/feat/orm/packages/sdk/src/server/plugins/functions.ts#L99

Pagebakers:
We discussed this morning that we'd like to introduce a feature flag before releasing this. So users/customers can opt-in to use it while we test it and we have a solution for the build times.

I'm happy to help implementing this.

```ts
configureWunderGraphServer({
  experimental: {
    orm: true
  }
})

**Comment by StarpTech** [comment]
Hi @timkendall I did some CPU profiling for the startup performance. It doesn't look too bad. Around 25% degradation. I think there are places where we import way more than necessary or don't lazy load the feature. Maybe, we can find some quick improvements.

File: packages/sdk/src/server/plugins/functions.ts (hunk 12 above)
Comment (timkendall) [reply packages/sdk/src/server/plugins/functions.ts]: It will return after all controllers have been aborted (since `controller.abort` is synchronous).

File: packages/sdk/src/server/plugins/functions.ts [hunk 13]
```diff
@@ -40,38 +48,57 @@ const FastifyFunctionsPlugin: FastifyPluginAsync<FastifyFunctionsOptions> = asyn
...
+									const gen = implementation.subscriptionHandler(ctx);
+
+									reply.raw.once('close', async () => {
+										// call return on the operation's `AsyncGenerator`
 										await gen.return(0);
-										return reply.raw.end();
+										// Cancel operations created by the ORM for this operation
+										context.ormOperationControllers.forEach((controller) => controller.abort());
+										// @todo move to tracking in-flight operations via. our request context like we do with the ORM
+										operationClient.cancelSubscriptions();
```
timkendall, line 88: As you mentioned @jensneuse we can do more major refactoring here at some point (see comment on line 71).

timkendall:
@StarpTech yeah thats a lot of JS to load :) Put both the codegen and runtime instantiation being the `experimental.orm` flag for now as recommended. Will look into optimizing in the future (thats the fun part ;).

StarpTech (approved):
LGTM

timkendall:
@StarpTech small tweaks included in conflict resolution if you could review before I merge (nothing changing functionality, just renaming some bindings and tweaking type parameter order on `HandlerContext`)

StarpTech:
@timkendall LGTM

StarpTech (approved):
LGTM

Pagebakers:
We should set ORM to never in wundergraph.factory.ts if it's not enabled, otherwise it results in a type error

```
// Code generated by wunderctl. DO NOT EDIT.

/**
 * "Warning: `Enable the ORM by setting `experimental.orm` to `true` in your `wundergraph.config.ts`.";
 */
export type ORM = never

```

PR #786, opened 2023-04-03
chore: generate build_info.json first

Aenimus:
## Motivation and Context

Generate build_info.json first, so that if anything afterwards goes wrong, we at least have the build_info to debug.
Extracted synchronous file creation to a generic function.
Removed pointless logging.

#### Checklist

- [ ] run `make test`


fiam (commented):
Can we refactor that to a function?

StarpTech (approved):
LGTM

packages/sdk/src/configure/index.ts
```diff
@@ -1053,15 +1021,15 @@ const mapRecordValues = <TKey extends string | number | symbol, TValue, TOutputV
...
+	if (doneCount === total) {
 		setTimeout(() => {
 			Logger.info(`code generation completed`);
 			process.exit(0);
 		}, 10);
 	}
 };
 
-const ResolvedWunderGraphConfigToJSON = (config: ResolvedWunderGraphConfig): string => {
+const ResolvedWunderGraphConfigToStringifiedJSON = (config: ResolvedWunderGraphConfig): string => {
```
Pagebakers, line 1032: Stringified json? JSON is always a string

Pagebakers (approved):

packages/sdk/src/configure/index.ts
Aenimus: Let me explain:
This method produces a POJO, and then `return JSON.stringify(pojo, null, 2);` is called.
We compare this stringified object to the original contents of config.

I wanted to make it clear that this function returns a string and not a POJO, because we name some POJOs as JSON.

Aenimus: Actually, maybe I am just mistaken?

packages/sdk/src/configure/index.ts
```diff
@@ -943,11 +946,9 @@ export const configureWunderGraphApplication = <
 					await GenerateCode({
 						wunderGraphConfig: resolved,
 						templates: gen.templates,
-						basePath: gen.path || 'generated',
+						basePath: gen.path || generated,
 					});
 				}
-				done();
-				Logger.info(`Code generation completed.`);
```
Aenimus, line 950: What was the point of this logging?

fiam (approved):
LGTM

packages/sdk/src/configure/index.ts
```diff
@@ -78,6 +78,11 @@ import { resolveServerOptions, serverOptionsWithDefaults } from '../server/util'
 import { loadNodeJsOperationDefaultModule, NodeJSOperation } from '../operations/operations';
 import zodToJsonSchema from 'zod-to-json-schema';
 
+const utf8 = 'utf8';
```
StarpTech, line 81: A bit overoptimized but still good to go :smile:

PR #787, opened 2023-04-03
fix: translating oas spec in the v2 oas introspection

devsergiy:
## Motivation and Context

Fixes problems with translating OAS specs (slack in swagger format, spotify with extensions)

- Do not try to convert swagger 2.0 to OAS 3 as omnigraph works with both formats
- Remove extensions from OAS spec, as omnigraph do not expect extensions be in the first place of other fields

#### Checklist

- [ ] run `make test`


jensneuse (approved):

fiam (approved):
LGTM

PR #788, opened 2023-04-04
feat: add ability to enable inspect for the server script runner

devsergiy:
## Motivation and Context

User needs a way to debug ts operations

PR adds `--inspect` param to the node process in case wunderctl was launched with `--debug` flag

#### Checklist

- [ ] run `make test`

### DIFF
--- a/cli/helpers/server_runner.go
+++ b/cli/helpers/server_runner.go
@@ -11,8 +11,9 @@ import (
 type ServerRunConfig struct {
 	WunderGraphDirAbs string
 	ServerScriptFile  string
-	Production        bool
 	Env               []string
+	Production        bool
+	Debug             bool
 }
 
 func NewServerRunner(log *zap.Logger, cfg *ServerRunConfig) *scriptrunner.ScriptRunner {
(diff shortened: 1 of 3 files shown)


cli/helpers/server_runner.go
```diff
@@ -25,11 +26,17 @@ func NewServerRunner(log *zap.Logger, cfg *ServerRunConfig) *scriptrunner.Script
 		hooksEnv = append(hooksEnv, "NODE_ENV=production")
 	}
 
+	scriptArgs := make([]string, 0, 2)
+	if cfg.Debug {
+		scriptArgs = append(scriptArgs, "--inspect")
```
ggeijteman-wg, line 31: Can we pick either `inspect` as the variable name or use `--debug` as the CLI parameter? This would keep things more consistent.

ggeijteman-wg (changes requested):

cli/helpers/server_runner.go
devsergiy: in this case, it is a node js `--inspect` flag, which we provide when wunderctl is launched with `--debug` flag

e.g. `wunderctl up --debug` leads to such command to launch server `node --inspect generated/bundle/server.js`

fiam (approved):
LGTM

cli/helpers/server_runner.go
ggeijteman-wg: In that case, please take my approval :them

ggeijteman-wg (approved):

cli/helpers/server_runner.go
devsergiy: thank you sir :)

cli/helpers/server_runner.go
```diff
@@ -25,11 +26,17 @@ func NewServerRunner(log *zap.Logger, cfg *ServerRunConfig) *scriptrunner.Script
 		hooksEnv = append(hooksEnv, "NODE_ENV=production")
 	}
 
+	scriptArgs := make([]string, 0, 2)
+	if cfg.Debug {
```
StarpTech, line 30: Could you also add `--inspect-brk` otherwise you can't debug the initialization / bootstrap phase because you aren't fast enough to add a breakpoint.

devsergiy: Sorry, I have merged it before your comment

hmm, not sure about what initialization you mean

as we have source maps, you should be able to add `.wundergraph` directory and put a breakpoint right in the file

and TS operation launched when you call it

StarpTech: > The only difference between the two is that —inspect-brk stops the execution at the start of the script and the —inspect does not.

Therefore allows you to measure startup time easier.

PR #791, opened 2023-04-04
feat: relay client

DaniAkash:
## Motivation and Context

Draft implementation of WunderGraph relay client that can be used with react & next.js projects.

#### Checklist

- [ ] run `make test`


packages/relay-nextjs/src/createRelayApp.ts
```diff
@@ -0,0 +1,215 @@
...
+	};
+
+	const subscribe: SubscribeFunction = (params, variables) => {
+		return Observable.create((sink) => {
+			const { id } = params;
+			const abort = new AbortController();
+			client
+				.subscribe(
+					{
+						operationName: `relay/${id}`,
```
jensneuse, line 49: I'm not sure if it will be a problem if we hardcode the `relay` prefix.
If there's no way out, we might go with this approach, but if it's possible, I'd prefer if you can make this configurable.

packages/relay-nextjs/package.json
```diff
@@ -0,0 +1,58 @@
+{
+  "name": "@wundergraph/relay-nextjs",
+  "version": "0.0.1",
+  "license": "Apache-2.0",
+  "description": "WunderGraph Solid Query Integration",
```
jensneuse, line 5: Copy Pasta error

packages/relay-nextjs/src/createRelayApp.ts
DaniAkash: Will look for a way

testapps/nextjs-relay/pages/index.tsx
```diff
@@ -40,9 +40,9 @@ function Home({ preloadedQuery }: RelayProps<{}, IndexQueryType>) {
 }
 
 export default withRelay(Home, IndexQuery, {
```
Pagebakers, line 42: You could abstract this also `withWunderGraphRelay` for example, that handles setting up the environment, but also make sure it can be extended.

testapps/nextjs-relay/lib/createRelayApp.ts
```diff
@@ -0,0 +1,4 @@
+import { createRelayApp } from '@wundergraph/relay-nextjs';
+import { client } from '../components/generated/nextjs';
+
+export const { createClientEnvironment, createServerEnvironment, useLivePreloadedQuery } = createRelayApp(client);
```
Pagebakers, line 4: We can also create a codegen template for this file, what do you think?

DaniAkash: Yup would be a good idea!

testapps/nextjs-relay/pages/index.tsx
DaniAkash: Have created the withWunderGraphRelay HOC 👍🏽

packages/relay-nextjs/src/createWunderGraphRelayApp.ts
```diff
@@ -0,0 +1,273 @@
...
+	Store,
+	SubscribeFunction,
+	createOperationDescriptor,
+	getRequest,
+} from 'relay-runtime';
+import { WiredOptions, WiredProps } from 'relay-nextjs/wired/component';
+
+// To avoid the error: The inferred type of X cannot be named without a reference to Y due to dependencies between relay-nextjs & next packages
+// Reference: https://github.com/microsoft/TypeScript/issues/47663#issuecomment-1270716220
+import type {} from 'next';
```
Pagebakers, line 22: I think there is another way, you can add the next types to the tsconfig.json

There was a similar issue for zod some time ago reported in the issues.

packages/relay-nextjs/src/createWunderGraphRelayApp.ts
```diff
@@ -0,0 +1,273 @@
...
+			}
+		);
+
+		// Adding a display Name to help with debugging with React Dev Tools
+		WrappedComponent.displayName = `withWunderGraphRelay(${Component.displayName || Component.name})`;
+
+		return WrappedComponent;
+	};
+
+	const useLivePreloadedQuery = <TQuery extends OperationType>(
```
Pagebakers, line 222: Now everything is a live query by default. Can you infer the query type from the parameters? eg if it's query or subscription?

LiveQueries should be enabled using `liveQuery: true`

Subscriptions should support `subscribeOnce: true` parameter, which makes them behave as a regular query.

We  can also simply use `usePreloadedQuery` as the name I think.

Pagebakers (changes requested):

packages/relay-nextjs/src/createWunderGraphRelayApp.ts
```diff
@@ -0,0 +1,273 @@
...
+	onAbort?(): void;
+}
+
+export interface UseSubscribeToProps extends SubscriptionRequestOptions {
+	enabled?: boolean;
+	onSuccess?(response: ClientResponse): void;
+	onError?(error: ResponseError): void;
+}
+
+export const createWunderGraphRelayApp = (client: Client) => {
```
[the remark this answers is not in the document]
DaniAkash: I'm not sure if the sdk itself will need any changes for making this work 🤔

packages/relay-nextjs/src/createWunderGraphRelayApp.ts
DaniAkash: I tried changing it but as per relay docs, `usePreloadedQuery` is used on component level for getting the data. But if we keep the name usePreloadedQuery with subscription status - we'll have to force existing users of relay to refactor their components. Taking into account, it's a safer to keep a separate hook for live queries. If the user doesn't need livequery they don't have to refactor any existing components

packages/relay-nextjs/src/createWunderGraphRelayApp.ts
DaniAkash: Tried adding
```
"include": ["src/**/*", "next-env.d.ts", "tests"],
```
to tsconfig.json & created `next-env.d.ts` in root similar to next.js project but the build still fails :(

packages/relay-nextjs/src/createWunderGraphRelayApp.ts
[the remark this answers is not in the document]
Pagebakers: Also not sure, in any case it's better to make it so that it accepts an options object, so we can easily extend it later when adding more features.

```
export const createWunderGraphRelayApp = (options: CreateWunderGraphRelayOptions) => {
```

packages/relay-nextjs/src/createWunderGraphRelayApp.ts
Pagebakers: As discussed we will add subscriptions support later.

packages/relay-nextjs/src/createWunderGraphRelayApp.ts
DaniAkash: Fixed after removing nextjs specific dependencies completely

packages/relay-nextjs/src/createWunderGraphRelayApp.ts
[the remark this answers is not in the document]
DaniAkash: Solved by making the options an object based config

testapps/nextjs-relay/lib/createRelayApp.ts
DaniAkash: will create codegen after first version release

cli/commands/loadoperations.go
```diff
@@ -38,7 +37,7 @@ var loadoperationsCmd = &cobra.Command{
 		if err != nil {
 			return err
 		}
-		return ioutil.WriteFile(outFile, []byte(out), os.ModePerm)
+		return os.WriteFile(outFile, []byte(out), os.ModePerm)
```
Pagebakers, line 40: This was required for relay?

docs-website/src/pages/docs/getting-started/index.jsx
```diff
@@ -52,6 +53,13 @@ export default function GettingStartedPage() {
 						href="/docs/getting-started/remix-quickstart"
 						variant="plain"
 					/>
+					<QuickLink
+						title="Relay"
+						description="Learn how to use WunderGraph with Remix."
```
Pagebakers, line 58: Typo: remix -> relay

docs-website/src/pages/docs/getting-started/relay-quickstart.md
```diff
@@ -0,0 +1,192 @@
...
+5. **Colocation**: The `relay-compiler` enforces the colocation principle, which means keeping GraphQL fragments and their corresponding React components in the same file. This makes it easier to reason about and manage data dependencies in your application.
+
+In summary, the `relay-compiler` is a crucial part of the Relay ecosystem, as it generates artifacts, optimizes queries, validates queries against the schema, generates TypeScript types, and enforces colocation. These features help Relay applications to be more efficient, reliable, and maintainable.
+
+The relay-compiler will need to be configured to function with the WunderGraph client. Instead of writing client side operations in `.wundergraph/operations` like all the other clients, we will instead write them alongside the react components and have relay generate the code using `relay-compiler`, while using WunderGraph to serve as the relay's client side environment for making network requests.
+
+### References
+
+- [Thinking in Relay](https://relay.dev/docs/principles-and-architecture/thinking-in-relay/)
+- [Official Tutorial](https://relay.dev/docs/tutorial/intro/)
```
Pagebakers, line 192: Also link to the reference docs

packages/react-relay/src/createWunderGraphRelayApp.tsx
```diff
@@ -0,0 +1,267 @@
...
+			data,
+			error,
+			...state,
+		};
+	};
+
+	const useLiveQuery = <TQuery extends OperationType>({
+		query,
+		queryReference,
+		options = { liveQuery: true },
```
Pagebakers, line 176: I'd move this up in the root object

pkg/loadoperations/loader.go
```diff
@@ -120,6 +121,9 @@ func (l *Loader) readOperations() error {
 			l.readTypescriptOperation(filePath)
 		case ".graphql":
 			l.readGraphQLOperation(filePath)
+		case ".json":
```
Pagebakers, line 124: Should we be more specific here? eg not all .json files are relay files?

Pagebakers (changes requested):
🙌 a few minor comments

pkg/loadoperations/loader.go
DaniAkash: @Pagebakers users can define any name for the persisted json file created by relay. Either we'll have to read the relay config or let it be ".json" files for now

cli/commands/loadoperations.go
jensneuse: ioutil is deprecated

Pagebakers (approved):

jensneuse (approved):

PR #813, opened 2023-04-09
feat: tui + log handling

StarpTech:
## Motivation and Context

In wunderctl we have very verbose logging by default. This makes it hard to understand the actual context and especially find the entry point for beginners.

This PR introduces a TUI mode that simplifies the work with wundergraph fundamentally. Instead of showing logs, the user can see helpful information about the state of the WunderNode and only get errors when he has to intervene. This is how most tools have solved it (see frontend tooling like vite, next.js ...).

I removed several logs but also change some from `info` to `debug`. This results in far fewer logs when running `wunderctl generate` or `wunderctl start`.

```
> wunderctl generate

{"level":"info","time":1681119875485,"pid":51259,"hostname":"pop-os","component":"@wundergraph/sdk","msg":"Building ..."}
{"level":"info","time":1681119875902,"pid":51259,"hostname":"pop-os","component":"@wundergraph/sdk","msg":"Build completed."}

```

```
> wunderctl start

{"level":"info","time":1681120102641,"msg":"Starting WunderNode","hostname":"pop-os","pid":53559,"component":"@wundergraph/wunderctl","version":"dev","commit":"d30e6023","date":"Mon Apr 10 11:47:21 AM CEST 2023","builtBy":"dev"}
{"level":"info","time":1681120102647,"msg":"Listening on","hostname":"pop-os","pid":53559,"component":"@wundergraph/wunderctl","component":"@wundergraph/node","addr":"127.0.0.1:9991"}
{"level":"info","time":1681120102814,"pid":53576,"hostname":"pop-os","component":"@wundergraph/server","msg":"Server listening at http://127.0.0.1:9992"}

```

```
> wunderctl up --logs

{"level":"info","time":1681120920380,"msg":"Starting WunderNode","hostname":"pop-os","pid":67513,"component":"@wundergraph/wunderctl","version":"dev","commit":"28c0ed3e","date":"Mon Apr 10 12:01:37 PM CEST 2023","builtBy":"dev"}
{"level":"info","time":1681120921525,"pid":67536,"hostname":"pop-os","component":"@wundergraph/sdk","msg":"Building ..."}
{"level":"info","time":1681120921939,"pid":67536,"hostname":"pop-os","component":"@wundergraph/sdk","msg":"Build completed."}
{"level":"info","time":1681120921963,"msg":"Listening on","hostname":"pop-os","pid":67513,"component":"@wundergraph/wunderctl","component":"@wundergraph/node","addr":"127.0.0.1:9991"}
{"level":"info","time":1681120922126,"pid":67559,"hostname":"pop-os","component":"@wundergraph/server","msg":"Server listening at http://127.0.0.1:9992"}
```

### Log mode

It's important to note that the TUI mode doesn't replace the log mode. The log mode can be toggled in multiple ways:

1. `wunderctl up --ui=false` Disable TUI mode and switch to regular logs
2. `wunderctl up --logs` Disable TUI mode and switch to regular logs
3. `wunderctl up --debug` Request logging is not compatible with TUI, therefore it is equivalent to 1)
4. Stdout is not a Terminal. We will fallback to logging.

### TUI mode

image

#### Checklist

- [ ] run `make test`


Pagebakers:
I don't think we should block the entire server from starting if one or more services isn't configured correctly. EG in most cases right now I don't care if my minio service is running or not, if I'm not working on uploads right now for example. Now I need to either start minio or comment out the config so I'm not blocked, but the error message is very unclear, i have no idea what is broken and how to fix it. 

Screenshot 2023-04-11 at 13 09 40

Pagebakers:
This will be more user friendly. Except for the error message, this needs imporvement.

Screenshot 2023-04-11 at 13 19 36 copy

jensneuse:
> I don't think we should block the entire dev server from starting if one or more services isn't configured correctly or is offline. EG in most cases right now I don't care if my minio service is running or not, if I'm not working on uploads right now for example. Now I need to either start minio or comment out the config so I'm not blocked, but the error message is very unclear, i have no idea what is broken and how to fix it.
> 
> Especially in case of uploads it's not very important to block the server, since the types and client will still be generated. Only uploading won't work.
> 
> Also the menu disappears when you have an error, so no links to docs or discord/ github when they are most important.
> 
> Screenshot 2023-04-11 at 13 09 40

Minio not running should be a warning.

jensneuse:
> This will be more user friendly. Except for the error message, this needs imporvement.
> 
> Screenshot 2023-04-11 at 13 19 36 copy

Some polishing needed, but in general good idea.

StarpTech:
> I don't think we should block the entire dev server from starting if one or more services isn't configured correctly or is offline. EG in most cases right now I don't care if my minio service is running or not, if I'm not working on uploads right now for example. Now I need to either start minio or comment out the config so I'm not blocked, but the error message is very unclear, i have no idea what is broken and how to fix it.

This PR doesn't change the behavior when a server is functional. If you see an error then we handle it as such. Your server might still be functional, but it's an error, and the user has to fix it.

Returning proper errors with enough context is a big process we need to address.

> Minio not running should be a warning.

This is not related to the TUI but to some of my changes. This is the error that the server returns when it instantiate the s3 clients. I'll revert it.

StarpTech:
> https://github.com/wundergraph/wundergraph/pull/813#issuecomment-1503158976

Error messages aren't always one-lined. It would break the view and does not allow for pretty errors.

Pagebakers:
> > I don't think we should block the entire dev server from starting if one or more services isn't configured correctly or is offline. EG in most cases right now I don't care if my minio service is running or not, if I'm not working on uploads right now for example. Now I need to either start minio or comment out the config so I'm not blocked, but the error message is very unclear, i have no idea what is broken and how to fix it.
> 
> This PR doesn't change the behavior when a server is functional. If you see an error then we handle it as such. Your server might still be functional, but it's an error, and the user has to fix it.

Yeah in that case we should log certain errors as warnings, instead of errors, so the server keeps running.

> 
> Returning proper errors with enough context is a big process we need to address.

Yup
> 
> > Minio not running should be a warning.
> 
> This is not related to the TUI but to some of my changes. This is the error that the server returns when it instantiate the s3 clients. I'll revert it.

👍

The errors can be multi line, but we can still show the menu with links to docs etc if there is an error no?

Pagebakers (approved):
Let's ship it, but still missing the menu on the error screen, this is where links to docs/discord/help are much more relevant.

StarpTech:
> Let's ship it, but still missing the menu on the error screen, this is where links to docs/discord/help are much more relevant.

I tried but requires more effort. If the content is bigger than the console the view is moved to the top.

jensneuse:
> > > I don't think we should block the entire dev server from starting if one or more services isn't configured correctly or is offline. EG in most cases right now I don't care if my minio service is running or not, if I'm not working on uploads right now for example. Now I need to either start minio or comment out the config so I'm not blocked, but the error message is very unclear, i have no idea what is broken and how to fix it.
> > 
> > This PR doesn't change the behavior when a server is functional. If you see an error then we handle it as such. Your server might still be functional, but it's an error, and the user has to fix it.
> 
> Yeah in that case we should log certain errors as warnings, instead of errors, so the server keeps running.
> 
> > 
> > Returning proper errors with enough context is a big process we need to address.
> 
> Yup
> > 
> > > Minio not running should be a warning.
> > 
> > This is not related to the TUI but to some of my changes. This is the error that the server returns when it instantiate the s3 clients. I'll revert it.
> 
> 👍
> 
> The errors can be multi line, but we can still show the menu with links to docs etc if there is an error no? 
> 

I agree. We need to improve logs a lot, especially error logging. Lots of errors could be warning or below.

PR #815, opened 2023-04-10
fix: add a check that token string is a real jwt token

devsergiy:
#### Checklist

- [ ] run `make test`


thisisnithin (approved):

Pagebakers (approved):
Perhaps we can make it a bit more efficient? eg we can easily check if a string is a jwt token or something else with a regex.

PR #829, opened 2023-04-13
fix: svelte-query livequery config

DaniAkash:
## Motivation and Context

Fixes an issue with svelte query's liveQuery which gets connected even when livequery config is set to false

#### Checklist

- [x] run `make test`

### DIFF
--- a/packages/svelte-query/src/lib/createSvelteClient.ts
+++ b/packages/svelte-query/src/lib/createSvelteClient.ts
@@ -129,17 +129,16 @@ export function createSvelteClient<Operations extends OperationsDefinition>(clie
 			refetchOnWindowFocus: liveQuery ? false : refetchOnWindowFocus,
 		});
 
-		const subscriptionState = createSubscribeTo({
-			queryHash,
-			operationName,
-			input,
-			liveQuery,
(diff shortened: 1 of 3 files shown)


jensneuse (approved):

packages/svelte-query/src/lib/createSvelteClient.ts
```diff
@@ -129,17 +129,16 @@ export function createSvelteClient<Operations extends OperationsDefinition>(clie
...
-			operationName,
-			input,
-			liveQuery,
-			enabled: options.enabled !== false && liveQuery,
-			onSuccess: options.onSuccess,
-			onError: options.onError,
-		});
-
 		if (liveQuery) {
+			const subscriptionState = createSubscribeTo({
```
Pagebakers, line 133: This doesn't give any issues when liveQuery is switched on? like in React?

Eg, there is already a check here: `enabled: options.enabled !== false && liveQuery,`

DaniAkash: The enabled check in the subscribeTo logic was missed - I have added it back

Pagebakers (approved):

PR #835, opened 2023-04-13
fix: issue where webhooks with payload would fail

Pagebakers:
#### Checklist

- [ ] run `make test`


packages/sdk/src/server/plugins/webhooks.ts
```diff
@@ -35,7 +35,7 @@ const FastifyWebhooksPlugin: FastifyPluginAsync<FastifyWebHooksOptions> = async
 					try {
 						const operationClient = new OperationsClient({
 							baseURL: config.nodeURL,
-							clientRequest: (request.body as any)?.__wg.clientRequest,
+							clientRequest: (request.body as any)?.__wg?.clientRequest,
```
StarpTech, line 38: I debug a bit. I recommend setting `clientRequest` to `null` to make it clear that we haven't implemented it yet. The node handles it gracefully.

StarpTech (changes requested):

StarpTech (approved):
LGTM

PR #836, opened 2023-04-14
fix: include config headers in graphiql

jensneuse:
## Motivation and Context

This fix includes user-defined headers in graphiql. Previously, we've omitted them in the custom GraphQL fetcher.


DaniAkash (approved):

fiam (approved):
LGTM

PR #854, opened 2023-04-17
feat: add support for using proxies in data sources

fiam:
## Motivation and Context

In some environments, it's desirable for the WunderGraph node to only talk to the public Internet using intermediate machines (security, accountability, logging, etc..). To support these use cases, we now allow setting a global HTTP(S)
proxy that will be used when querying the data sources. Additionally, the global proxy can be overridden on a per-source basis.

## Marketing Changelog

WunderGraph can now use an HTTP(S) proxy to talk to its connected APIs.


packages/sdk/src/configure/options.ts
```diff
@@ -66,6 +68,11 @@ export interface NodeOptions {
 	 * @defaultValue 10 seconds
 	 */
 	defaultRequestTimeoutSeconds?: number;
+	/** Default HTTP(S) proxy to use
+	 *
+	 * @defaultValue undefined (no proxy)
+	 */
+	defaultHttpProxyUrl?: InputVariable;
```
StarpTech, line 75: Two things I couldn't find in this PR.

1. Did you cover the introspection HTTP client as well?
2. It should be possible for cloud that our environment variable precedes the users.

packages/sdk/src/configure/options.ts
```diff
@@ -18,6 +18,7 @@ export enum WgEnv {
 	ServerUrl = 'WG_SERVER_URL',
 	ServerHost = 'WG_SERVER_HOST',
 	ServerPort = 'WG_SERVER_PORT',
+	ProxyUrl = 'WG_PROXY_URL',
```
StarpTech, line 21: Can we call it `WG_HTTP_PROXY` it's more concret.

packages/protobuf/wgpb/wundernode_config.ts
```diff
@@ -1172,6 +1173,7 @@ export interface NodeOptions {
   listen: ListenerOptions | undefined;
   logger: NodeLogging | undefined;
   defaultRequestTimeoutSeconds: number;
+  defaultHttpProxyUrl: ConfigurationVariable | undefined;
```
StarpTech, line 1176: Why do you prefix it with `default` ?

packages/protobuf/wgpb/wundernode_config.ts
```diff
@@ -1027,6 +1027,7 @@ export interface FetchConfiguration {
   mTLS: MTLSConfiguration | undefined;
   baseUrl: ConfigurationVariable | undefined;
   path: ConfigurationVariable | undefined;
+  proxyUrl: ConfigurationVariable | undefined;
```
StarpTech, line 1030: Here as well. `httpProxyUrl`

StarpTech (changes requested):

packages/protobuf/wgpb/wundernode_config.ts
fiam: This is the default for all data sources, then each data source can override it

packages/sdk/src/configure/options.ts
fiam: 1 - No, the issue only mentioned the Go side, but you make a good point. I think this is not going to be very useful if we don't support both the Go and the TS sides
2 - Should we always override it with an environment variable we control?

StarpTech: 2) Yes, the same as it works with WG_NODE_PORT etc...

fiam: I was already overridden with `WG_PROXY_URL` when running in cloud, so renaming it to `WG_HTTP_PROXY_URL` should cover that. Keeping the conversation unresolved until I handle point 1.

packages/protobuf/wgpb/wundernode_config.ts
fiam: Thanks. Renamed.

packages/protobuf/wgpb/wundernode_config.ts
StarpTech: Ah ok, ty.

packages/sdk/src/configure/options.ts
fiam: I've added support for using proxies during introspection in the previous to last commit. I'll ask for a re-review once the tests are in, no rush for now.

packages/sdk/src/definition/graphql-introspection.ts
```diff
@@ -254,15 +267,41 @@ interface GraphQLIntrospectionResponse {
...
+				protocol: proxyUrl.protocol,
+			};
+			if (proxyUrl.username || proxyUrl.password) {
+				proxyConfig.auth = {
+					username: proxyUrl.username,
+					password: proxyUrl.password,
+				};
+			}
+		} catch (e: any) {
+			throw new Error(`invalid HTTP proxy URL '${proxyUrlString}': ${e}`);
```
StarpTech, line 298: Can we be more specific about the datasource?

fiam: I've added the upstream URL to the message

packages/sdk/src/configure/index.ts
```diff
@@ -94,11 +101,24 @@ export interface WunderGraphCorsConfiguration {
 	allowCredentials?: boolean;
 }
 
+/**
+ * ApiIntrospector<T> is a function type which the API generators must conform to.
+ * Given an ApiIntrospectionOptions, they should return a Promise that resolves
+ * to an Api<T>
+ */
+export type ApiIntrospector<T> = (options: ApiIntrospectionOptions) => Promise<Api<T>>;
```
StarpTech, line 109: What was the reason for introducing it?

StarpTech (changes requested):

packages/sdk/src/configure/index.ts
fiam: To document a bit better what the introspect.something() are expected to return

fiam: Re-reading it, maybe you were asking why we changed the return type. It was to be able to pass some data to the introspection functions (specifically, the global proxy to use, but we can extend that if needed)

fiam:
Just noticed that some of the code that generates snapshots needs to be updated

fiam:
> Just noticed that some of the code that generates snapshots needs to be updated

Updated

packages/sdk/src/configure/index.ts
StarpTech: Alright thanks.

StarpTech (approved):
Great! LGTM

fiam:
@StarpTech I know you already approved this, but I added a test using a local proxy as well as a workaround for supporting HTTPS over HTTP proxies (due to a bug in axios). If possible, PTAL. Thanks!

packages/testsuite/apps/generation/index.test.ts
```diff
@@ -0,0 +1,137 @@
...
+
+// Invalid proxy used to make generation fail
+const invalidProxy = 'http://this.does.not.exist';
+const validProxy = proxyFromEnvironment || proxyFromSquid;
+
+const itIfValidProxy = !!validProxy ? it : it.skip;
+
+describe('Proxy configurations', () => {
+	it('uses an invalid global proxy for generation and fails', async () => {
+		process.env.WG_HTTP_PROXY = invalidProxy;
```
StarpTech, line 81: Can we extend the client in a way that this is configurable through the interface instead env?

packages/testsuite/apps/generation/test-setup.ts
```diff
@@ -0,0 +1,5 @@
+import { Setup } from '../../setup/generate';
+
+export default async function () {
```
StarpTech, line 3: This is not necessary when you generate in the test.

packages/testsuite/apps/generation/docker-compose.yml
```diff
@@ -0,0 +1,8 @@
+version: '3'
+services:
```
StarpTech, line 2: What prevents us from starting a http server with node.js in the tests instead starting a container?

StarpTech (changes requested):

packages/testsuite/apps/generation/test-setup.ts
fiam: We do for the initial imports

packages/testsuite/apps/generation/docker-compose.yml
fiam: It doesn't support CONNECT for sending requests to HTTPS hosts through HTTP proxies

StarpTech: For reference: https://github.com/berstend/straightforward

fiam: Looking into it, thanks!

packages/testsuite/apps/generation/index.test.ts
fiam: I'm now passing them through the functions instead of using `process.env` PTAL

packages/testsuite/apps/generation/docker-compose.yml
fiam: Replaced the container with [berstend/straightforward](https://github.com/berstend/straightforward). It's significantly faster now. Great idea!

fiam:
@StarpTech @jensneuse I've addressed the requested changes and I've also added a new guide. PTAL. Thanks!

packages/testsuite/apps/generation/index.test.ts
```diff
@@ -0,0 +1,134 @@
...
+	await wg.stop();
+};
+
+const proxyFromEnvironment = process.env.HTTP_PROXY;
+const localProxyPort = 3138;
+const localProxy = `http://127.0.0.1:${localProxyPort}`;
+const sf = new Straightforward();
+
+beforeAll(async () => {
+	if (!proxyFromEnvironment) {
```
StarpTech, line 61: Why is it configurable?

packages/testsuite/apps/generation/index.test.ts
```diff
@@ -0,0 +1,134 @@
...
+
+		it('uses an invalid global proxy for queries and fails', async () => {
+			await regenerate({});
+			await expectQueryFailure({
+				WG_HTTP_PROXY: invalidProxy,
+			});
+		});
+
+		it('uses data source proxy to override invalid global proxy and generates', async () => {
+			await regenerate({
```
StarpTech, line 93: Can't we make it part of a beforeEach?

docs-website/src/pages/docs/guides/using-http-proxies.md
```diff
@@ -0,0 +1,50 @@
+---
+title: Using HTTP Proxies
+pageTitle: WunderGraph - HTTP Proxies
+description: This guide explains how to configure WunderGraph to use HTTP proxies
+---
+
+WunderGraph supports using HTTP or HTTPS proxies both when introspecting remote APIs as
+well as when performing operations. This affects APIs that are transported through HTTP,
```
jensneuse, line 8: I'd rephrase this somehow to "...proxies for both introspection and to make HTTP requests at runtime".

docs-website/src/pages/docs/guides/using-http-proxies.md
```diff
@@ -0,0 +1,50 @@
...
+namely GraphQL (either directly or via federation) and OpenAPI based data sources.
+
+Both HTTP and HTTPS proxies are supported. When talking to an HTTPS API through an HTTP
+server, WunderGraph will open an socket using and HTTP CONNECT request.
```
jensneuse, line 12: "a socket"

docs-website/src/pages/docs/guides/using-http-proxies.md
```diff
@@ -0,0 +1,50 @@
...
+The proxy can be set to either a string or an environment variable, using `EnvironmentVariable`.
+
+By default, `defaultHttpProxyUrl` reads the global proxy URL from the `WG_HTTP_PROXY` environment
+variable.
+
+This global proxy definition can be overridden on every data source.
+
+## Configuring a proxy for a single data source
+
+All data sources that use HTTP a transport have an `httpProxyUrl` field that can be set when
```
jensneuse, line 36: "that use a HTTP transport"

docs-website/src/pages/docs/guides/using-http-proxies.md
```diff
@@ -0,0 +1,50 @@
...
+defining them:
+
+```ts
+const countries = introspect.graphql({
+  apiNamespace: 'countries',
+  url: 'https://countries.trevorblades.com/',
+  httpProxyUrl: 'http://my.proxy.server:3128',
```
jensneuse, line 43: Should we name this httpProxyURL or just proxyURL?
In any case, I'd suggest to use URL instead or Url.

packages/sdk/src/configure/index.ts
```diff
@@ -672,12 +700,17 @@ const resolveUploadConfiguration = (
...
+	apiIntrospectionOptions: ApiIntrospectionOptions,
 	cors: CorsConfiguration,
 	s3?: S3Provider,
 	hooks?: HooksConfiguration
 ): Promise<ResolvedApplication> => {
-	const resolvedApis = await Promise.all(apis);
+	// Generate the promises first, then await them all at once
+	// to run them in parallel
+	const generators = await Promise.all(apis);
+	const resolvedApis = await Promise.all(generators.map((generator) => generator(apiIntrospectionOptions)));
```
jensneuse, line 712: Why async twice? Feels weird. Is the creation of the generator async?

packages/testsuite/apps/generation/index.test.ts
```diff
@@ -0,0 +1,134 @@
...
+	});
+
+	expect(result.error).toBeDefined();
+	expect(result.data).toBeUndefined();
+
+	await wg.stop();
+};
+
+const proxyFromEnvironment = process.env.HTTP_PROXY;
+const localProxyPort = 3138;
```
jensneuse, line 56: Should we use freeport() here?

fiam: Not worth it IMHO. The `freeport` package doesn't support promises and I would need to write our own function (like we did for the testing SDK). This is only used for testing, so it should only matter in environments we control. Additionally, 3128 (I'll fix the number) is Squid's default port and unlikely to be used by something else.

docs-website/src/pages/docs/guides/using-http-proxies.md
fiam: Last time this came up we agreed to use this convention for TypeScript/Javascript https://google.github.io/styleguide/jsguide.html#naming-camel-case-defined - According to that, it would be `Url` - I know in the Go side it would be `URL`. Regarding the full name, the initial draft had it as `proxyUrl` but it was requested that I changed it to `httpProxyUrl`. I don't have a strong opinion on that and I mind changing it back, you just need to make a deal with @StarpTech :-)

packages/testsuite/apps/generation/index.test.ts
fiam: Not all tests run regenerate with the same arguments, and some don't run it at all.

packages/testsuite/apps/generation/index.test.ts
fiam: In case you want to spin up a separate proxy and test against it. It's handy if you have to debug something (like it happened with Axios not support CONNECT)

docs-website/src/pages/docs/guides/using-http-proxies.md
fiam: Updated, PTAL

docs-website/src/pages/docs/guides/using-http-proxies.md
fiam: Updated

packages/testsuite/apps/generation/index.test.ts
```diff
@@ -0,0 +1,134 @@
...
+import { Setup } from '../../setup/generate';
+import { createTestServer } from './.wundergraph/generated/testing';
+
+// XXX: This test runs code generation, so we can't run multiple test files in parallel
+
+const regenerate = async (env: Record<string, string>) => {
+	await Setup('apps/generation', { env });
+};
+
+const expectQuerySuccess = async (env: Record<string, string>) => {
```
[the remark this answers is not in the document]
fiam: I've moved them to a separate file, PTAL

docs-website/src/pages/docs/guides/using-http-proxies.md
StarpTech: How do we name things right now? Do we already uppercase acronyms? If not, I'd not follow that convention because it's not common in javascript ecosystem as alberto pointed out with that example.

packages/testsuite/apps/generation/index.test.ts
StarpTech: ok

packages/testsuite/apps/generation/index.test.ts
StarpTech: You can group those again in `describe` to separate them.

StarpTech (approved):
Great work! LGTM

jensneuse (approved):
LGTM,.great work. One comment, does it work with the openapiv2 data source?

fiam:
> LGTM,.great work. One comment, does it work with the openapiv2 data source?

Thanks for the heads up! I had assumed it would just work, but I tested it and it does work. The code path used to send the request to load the schema is different than what we use for loading the schema from a "real" GraphQL server, so the former will never use a proxy.

packages/testsuite/apps/generation/index.test.ts
fiam: They're just 2 test that call `regenerate()` with the same arguments. I feel like moving them to a separate describe to add a `beforeEach()` breaks the logical organization (valid/invalid proxy, test generation, then test queries) and it ends up less readable..

PR #857, opened 2023-04-18
feat: inconsistencies in ts clients api

### LINKED ISSUE #810: Support token based authentication for SSE subscriptions
## Problem

Subscriptions support token based authentication only if using the fetch based implementation.
Currently I'm deleting `EventSource` from `globalThis` to prevent the client from using `SSE`.

## Suggested solution

I am aware that `EventSource` doesn't support headers, but this should be documented.

## Alternatives

Take in consideration EventSource polyfills which support custom headers.

## Additional context

* [EventSource](https://developer.mozilla.org/en-US/docs/Web/API/EventSource/EventSource) 
* [polyfill 1](https://github.com/EventSource/eventsource)
* [polyfill 2](https://github.com/Yaffle/EventSource/)

Pagebakers:
## Motivation and Context

We use different implementations of clients across the sdk, this PR consolidates all implementation, so all clients use the typescript client as the base implementation.

The internalClient has been deprecated for backwards compatibility.

#### Checklist

- [x] run `make test`

closes #810


packages/sdk/src/server/server.ts
```diff
@@ -72,15 +73,24 @@ if (process.env.START_HOOKS_SERVER === 'true') {
...
-export const configureWunderGraphServer = <
-	GeneratedHooksConfig extends HooksConfiguration,
-	GeneratedClient extends InternalClient,
-	GeneratedWebhooksConfig extends WebhooksConfig = WebhooksConfig
->(
-	configWrapper: () => WunderGraphServerConfig<GeneratedHooksConfig, GeneratedWebhooksConfig>
-): WunderGraphHooksAndServerConfig => {
-	return _configureWunderGraphServer<GeneratedHooksConfig, GeneratedWebhooksConfig>(configWrapper());
-};
+function configureWunderGraphServer(configWrapper: () => WunderGraphServerConfig): WunderGraphServerConfig;
```
Pagebakers, line 76: Switch this to using overloads, which are augmented in the generated code, this way you don't need to annotate `createWunderGraphServer` anymore in the config.

packages/sdk/src/server/operations-client.ts
```diff
@@ -56,6 +58,100 @@ export class OperationsClient<Queries = any, Mutations = any, Subscriptions = an
 		});
 	};
 
+	private convertGraphQLResponse(resp: GraphQLResponse, statusCode: number = 200): ClientResponse {
```
StarpTech, line 61: What needs to be handled differently here? Navily, I'd assume the typescript client already implements the full spec.

packages/sdk/src/codegen/templates/typescript/server.template.ts
```diff
@@ -20,6 +22,11 @@ export interface OutputConfig {
 }
 
 export interface GraphQLExecutionContext {
-    wundergraph: BaseRequestContext<User, InternalClient>;
+    wundergraph: BaseRequestContext<User, InternalClient, InternalOperations>;
 }
+
+declare module "@wundergraph/sdk/server" {
```
StarpTech, line 28: It looks odd to overwrite a module from the same package, but in that way, we can skip to configuring the types ourselves, correct? Can we apply this to `configureWunderGraphOperations`. Very nice.

packages/sdk/src/server/server.ts
StarpTech: I haven't read this. Check my comment above :smile:

packages/sdk/src/server/server.ts
```diff
@@ -72,15 +73,24 @@ if (process.env.START_HOOKS_SERVER === 'true') {
...
+
+function configureWunderGraphServer(configWrapper: () => any): any {
+	return _configureWunderGraphServer(configWrapper());
+}
+
+export { configureWunderGraphServer };
+
+// export const configureWunderGraphServer = <
```
StarpTech, line 84: Remove?

packages/sdk/src/server/server.ts
```diff
@@ -242,6 +252,13 @@ export const createServer = async ({
 					method: req.body.__wg.clientRequest?.method || 'GET',
 				},
 				internalClient: clientFactory({ 'x-request-id': req.id }, req.body.__wg.clientRequest),
+				operations: new OperationsClient({
```
StarpTech, line 255: Reminder: We need to update OTEL integration as well.

packages/sdk/src/codegen/templates/typescript/server.template.ts
```diff
@@ -20,6 +22,11 @@ export interface OutputConfig {
...
+    export function configureWunderGraphServer<GeneratedHooksConfig = HooksConfig, GeneratedInternalClient = InternalClient, GeneratedWebhooksConfig = WebhooksConfig>(configWrapper: () => WunderGraphServerConfig<GeneratedHooksConfig, GeneratedWebhooksConfig>) : WunderGraphHooksAndServerConfig
+}
+
```
StarpTech, line 31: Could you create an issue to remove the internal client on 1st June?

packages/sdk/src/server/server.ts
StarpTech, line 76: Please add tests to our test suite after the implementation has been settled.

StarpTech (changes requested):

StarpTech (changes requested):

packages/sdk/src/codegen/templates/typescript/server.template.ts
Pagebakers: Exactly, it gets overwritten in the generated code, so 'outside' the sdk package. 

I'll apply the same to configureWunderGraphOperations

packages/sdk/src/server/operations-client.ts
Pagebakers: I managed to extend the ts client so this is not needed anymore.

packages/sdk/src/codegen/templates/typescript/server.template.ts
Pagebakers: Done

packages/sdk/src/client/client.ts
```diff
@@ -355,30 +360,48 @@ export class Client {
...
+	): Promise<any> {
 		if (options.subscribeOnce) {
 			const result = await this.query<RequestOptions, Data, Error>(options);
-			cb(result);
+			cb?.(result);
 			return result;
 		}
-		if ('EventSource' in globalThis) {
+
+		if (cb && 'EventSource' in globalThis && options.sse !== false) {
```
jensneuse, line 383: I'd rename "sse" to "disableSSE", then it needs to be explicitly set to true.

packages/sdk/src/client/types.ts
```diff
@@ -102,8 +102,18 @@ export type SubscriptionRequestOptions<
...
+	 */
 	liveQuery?: Boolean;
+	/**
+	 * Receive the initial response and then stop the subscription
+	 */
 	subscribeOnce?: Boolean;
+	/**
+	 * Set to false to use fetch streaming instead of EventSource (SSE)
+	 */
+	sse?: Boolean;
```
jensneuse, line 116: see comment above

packages/sdk/src/codegen/templates/typescript/hooks.template.ts
```diff
@@ -4,6 +4,7 @@ import { {{ modelImports }} } from "./models"
 import type { BaseRequestContext, WunderGraphRequest, WunderGraphResponse, AuthenticationResponse, AuthenticationHookRequest, HooksConfiguration, WsTransportOnConnectionInitResponse, PreUploadHookRequest, PreUploadHookResponse, PostUploadHookRequest, PostUploadHookResponse } from "@wundergraph/sdk/server";
 import type { InternalClient } from "./wundergraph.internal.client"
 import type { User } from "./wundergraph.server"
+import { InternalOperationsClient } from "./wundergraph.internal.operations.client";
```
jensneuse, line 7: One question regarding code generation. I'm not sure how exactly you've solved it, but I think what we could do is generate types for "public" operations as well as types for "internal" operations, all in one file. Then, for the external client, we only inject the public operations, while for the internal client, we inject the union of public and internal operations types. This way, it should be quite clean and organized. WDYT?

packages/sdk/src/codegen/templates/typescript/internal.operations.client.template.ts
```diff
@@ -2,42 +2,45 @@
...
 //language=handlebars
 export const template = `
-import type { OperationsClientType } from "@wundergraph/sdk/server";
+import type { OperationsClient, InternalOperationsDefinition } from "@wundergraph/sdk/server";
+import type { ClientOperationErrors } from "@wundergraph/sdk/client";
 import type { OperationErrors } from "./ts-operation-errors";
 import { {{ modelImports }} } from "./models"
 
-export interface Queries {
+export type Queries = {
```
jensneuse, line 10: See comment above. I think we can just have one "operations client template", then generate types for internal and public, create a union and have one exported interface for the external client, one for the internal client. This could keep the code generation clean and reduce files.

packages/sdk/src/configure/index.ts
```diff
@@ -63,6 +63,7 @@ import {
 	WebhookConfiguration,
 	WunderGraphConfiguration,
 } from '@wundergraph/protobuf';
+import hash from 'object-hash';
```
jensneuse, line 66: Why is this here? I think @StarpTech replaced it in another place with a lightweight function. Is it needed at all? I don't see any usage.

jensneuse (changes requested):

packages/sdk/src/configure/index.ts
StarpTech: It is used. You mixed it up with something else. https://github.com/wundergraph/wundergraph/blob/9b9675a94fc6b61d1d72b35725a30716e18a7a50/packages/sdk/src/configure/index.ts#L1284

docs-website/src/pages/docs/wundergraph-server-ts-reference/pre-resolve-hook.md
```diff
@@ -26,21 +27,15 @@ e.g. to talk to a database or another service to enrich a response or manipulate
 
 ```typescript
 // wundergraph.server.ts
-export default configureWunderGraphServer<HooksConfig, InternalClient>(() => ({
```
StarpTech, line 29: Very nice, that we simplify it further.

packages/sdk/src/codegen/templates/typescript/hooks.template.ts
Pagebakers: Not sure that'll work, since public operations have different properties. We can investigate, but i'd open another issue for that, as this PR is big enough already.

packages/sdk/src/configure/index.ts
Pagebakers: Actually it can be removed again, I added it since applicationHash was required, but I made it optional instead.

docs-website/src/pages/docs/wundergraph-server-ts-reference/pre-resolve-hook.md
Pagebakers: Yep, simplify the API and gives us more freedom to extend or change the types later without breaking anything.

docs-website/src/pages/docs/upgrade-guides/internal-client-deprecated.md
```diff
@@ -0,0 +1,85 @@
...
+### Subscriptions
+
+Subscriptions are not supported by the internal client, but can be used with the operations client.
+
+```ts
+export default createOperation.subscription({
+  handler: async function* ({ operations }) {
+    const updates = await operations.subscribe({
+      operationName: 'federation/PriceUpdates',
+      subscribeOnce: true,
```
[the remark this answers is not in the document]
Pagebakers: That shouldn't be there indeed :D

packages/sdk/src/client/client.ts
```diff
@@ -355,30 +360,48 @@ export class Client {
...
+	>(options: RequestOptions): Promise<Response>;
 	public async subscribe<RequestOptions extends SubscriptionRequestOptions, Data = any, Error = any>(
 		options: RequestOptions,
-		cb: SubscriptionEventHandler<Data, Error>
-	) {
+		cb?: SubscriptionEventHandler<Data, Error>
+	): Promise<ClientResponse<Data, Error> | void>;
+	public async subscribe<RequestOptions extends SubscriptionRequestOptions, Data = any, Error = any>(
+		options: RequestOptions,
+		cb?: SubscriptionEventHandler<Data, Error>
```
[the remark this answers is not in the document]
Pagebakers: It has to be, since in the operations client we don't have a callback, it's just a generator.

In the ts client, you can omit the cb now and it will return a generator.

jensneuse (approved):

testapps/nextjs/.wundergraph/wundergraph.server.ts
```diff
@@ -15,7 +13,7 @@ const testEnum = new GraphQLEnumType({
 	},
 });
 
-export default configureWunderGraphServer<HooksConfig, InternalClient>(() => ({
```
StarpTech, line 18: How does it work when the type definitions aren't imported? :thinking:

StarpTech (changes requested):

testapps/nextjs/.wundergraph/wundergraph.server.ts
Pagebakers: This is a requirement in tsconfig.json

```
  "include": [".wundergraph/**/*.ts"]
```

If this is not added, it will only work by adding `HooksConfig, InternalClient` generics.

packages/sdk/src/server/plugins/hooks.ts
```diff
@@ -247,6 +247,7 @@ const FastifyHooksPlugin: FastifyPluginAsync<FastifyHooksOptions> = async (fasti
 				throw new Error(errorMessage);
 			}
 			req.ctx.internalClient = req.ctx.internalClient.withHeaders({ 'Wg-Cycle-Counter': body.cycleCounter });
+			req.ctx.operations = req.ctx.operations.withHeaders({ 'Wg-Cycle-Counter': body.cycleCounter });
```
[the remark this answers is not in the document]
Pagebakers: Yes, this works as well.

testapps/nextjs/.wundergraph/wundergraph.server.ts
StarpTech: Should we document it somewhere? Docs?

StarpTech (approved):
LGTM but please document the tsconfig changes.

PR #863, opened 2023-04-19
fix: temporarily removed ESM package to allow bundling in next.js

DaniAkash:
## Motivation and Context

This PR fixes bundling issue of `@wundergraph/react-relay` package when used with Next.js. 

The `@wundergraph/react-relay` exposes an esm bundle with `.mjs` extension which imports packages from `react-relay` package which only has cjs bundle with `.js` extension. This is leading to Next.js build failures as highlighted by @Pagebakers in https://github.com/vercel/next.js/issues/39375#issuecomment-1207917866

To solve the issue temporarily we are completely removing ESM bundle from `@wundergraph/react-relay` and making it cjs only

#### Checklist

- [ ] run `make test`

### DIFF
--- a/packages/react-relay/package.json
+++ b/packages/react-relay/package.json
@@ -4,12 +4,10 @@
   "license": "Apache-2.0",
   "description": "WunderGraph Relay Integration",
   "main": "./dist/index.js",
-  "module": "./dist/index.mjs",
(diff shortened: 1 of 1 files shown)


Pagebakers (approved):

StarpTech (approved):
LGTM

PR #866, opened 2023-04-20
fix: don't override request context after assigning client request (#866

jensneuse:
This fixes a bug where we first assign the client request to the request context and then override the context again, omitting the client request, which is then not properly injected into the transport, resulting in the client request missing in internal GraphQL requests to internally mounted/embedded GraphQL servers.

### DIFF
--- a/packages/testsuite/apps/basic/index.test.ts
+++ b/packages/testsuite/apps/basic/index.test.ts
@@ -58,4 +58,20 @@ describe('Operations', () => {
 		expect(badRequestError?.code).toBe('BadRequest');
 		expect(badRequestError?.statusCode).toBe(400);
 	});
+
+	it('Should return a client request header', async () => {
+		const client = wg.client();
+		client.setExtraHeaders({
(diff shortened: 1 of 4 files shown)


DaniAkash (approved):

Pagebakers (approved):

fiam (approved):
LGTM. Nice detective skills 🕵️‍♂️

packages/testsuite/apps/basic/index.test.ts
```diff
@@ -58,4 +58,20 @@ describe('Operations', () => {
 		expect(badRequestError?.code).toBe('BadRequest');
 		expect(badRequestError?.statusCode).toBe(400);
 	});
+
+	it('Should return a client request header', async () => {
```
StarpTech, line 62: It's not obvious that this test covers a resolver from a custom graphql server. Please put it next time in a custom app or separate it with a `describe(` that describes the test case better.

PR #877, opened 2023-04-21
feat: initial implementation for the operations code generation

jensneuse:
## Motivation and Context

Users want to be able to quickly turn a data source into API Endpoints.
Using auto-generated operations, they can generate (customizable) Operations with as little as 2 lines of code. (see docs)

## Marketing Changelog

We've got feedback from our user base that it can be tedious to import a large amount of OpenAPI Endpoints and then have to write out every Operation, although 90% of all Operations can be exposed as is.

With auto-generated Operations, you can filter API Namespaces and root fields in 2 lines of code, and WunderGraph will generate Operations for all selected fields automatically. And even if the generated Operation is not exactly what you needed, there's an easy way to customize it.


docs-website/src/pages/docs/auto-generated-operations-reference/index.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Example configuration
+
+```typescript
+// .wundergraph/wundergraph.generate.ts
+import { configureWunderGraphGeneration } from '@wundergraph/sdk';
+
+export default configureWunderGraphGeneration({
+  codeGenerators: [], // abbreviated for brevity
+  operationsGeneration: (config) => {
```
Pagebakers, line 35: `operationsGenerator` would be more consistent with `codeGenerators`

docs-website/src/pages/docs/auto-generated-operations-reference/index.md
```diff
@@ -0,0 +1,115 @@
...
+  apis: [weather, federated],
+  generate,
+});
+```
+
+## Customization of auto-generated Operations
+
+Here's an example of how a generated Operation looks like:
+
+```graphql
```
Pagebakers, line 87: docs don't mention where the operation is generated, in the root of operations/ or in a sub directory eg operations/namespace?

Pagebakers (changes requested):
Massive productivity gain, a few comments on the docs.

Also should mention this in the wundergraph.config.ts reference.

docs-website/src/pages/docs/auto-generated-operations-reference/index.md
fiam: I'd say these are different things, I think there don't need to be named in the same way.

docs-website/src/pages/docs/auto-generated-operations-reference/index.md
```diff
@@ -0,0 +1,115 @@
...
+
+export default configureWunderGraphGeneration({
+  codeGenerators: [], // abbreviated for brevity
+  operationsGeneration: (config) => {
+    // include all root fields from the "weather" & "federated" API Namespace
+    config.includeNamespaces('weather', 'federated');
+    // we could also exclude the root field "me" from the "federated" API Namespace
+    config.filterRootFields(
+      (field) => field.operationType === 'query' && field.apiNamespace === 'federated' && field.name === 'me'
+    );
```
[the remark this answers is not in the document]
jensneuse: yes

docs-website/src/pages/docs/auto-generated-operations-reference/index.md
jensneuse: changed

jensneuse:
> Massive productivity gain, a few comments on the docs.
> 
> Also should mention this in the wundergraph.config.ts reference.

Once released, I'll write a guide and we can also update some examples.

Pagebakers (approved):

PR #893, opened 2023-04-25
feat: implement mocking capabilities

StarpTech:
## Motivation and Context

This PR adds the functionality to allow mocking any datasource or external HTTP API. It extends the testing library to write HTTP mocks in TypeScript. I demonstrated how to mock the countries API and OpenSearch client in tests.

```ts
// Mock the search endpoint of OpenSearch
// Mock can only be used once. If no URL matches an error is thrown on `scop.done()`. This ensure that the test runner
// can handle the error (pretty-print, stack-trace)
const scope = ts.mockServer.mock<Record<string, any>>(
  async ({ url, method }) => {
    return url.path === "/books/_search" && method === "POST";
  },
  async ({ json }) => {
    expect(await json()).toEqual({
      query: {
        match: {
          title: {
            query: "The Outsider",
          },
        },
      },
    });

    return {
      body: mockSearchResponse,
    };
  }
);

const result = await ts.testServer.client().query({
  operationName: "search",
});

// Re-Throws the assertion error or when handler wasn't called
scope.done();

expect(result.error).toBeUndefined();
expect(result.data).toBeDefined();
expect(result.data).toEqual(mockSearchResponse);

```

https://linear.app/wundergraph/issue/ENG-1944/introduce-an-easy-way-to-mock-datasources-in-testing-sdk

## TODO

- [x] Add documentation

#### Checklist

- [x] run `make test`


packages/sdk/src/client/client.ts
```diff
@@ -128,11 +129,25 @@ export class Client {
 			...init.headers,
 		};
 
-		return fetchImpl(input, {
+		let timeout: NodeJS.Timeout | undefined;
+
+		if (!init.signal && this.options.requestTimeoutMs && this.options.requestTimeoutMs > 0) {
```
StarpTech, line 134: @Pagebakers any objections that could break clients? As far I know, we already assume a polyfill of `AbortController`. https://docs.wundergraph.com/docs/clients-reference/typescript-client#browser

.github/actions/wunderctl/action.yaml
```diff
@@ -3,8 +3,22 @@ description: 'This action install wunderctl with race detection enabled and sets
...
   using: 'composite'
   steps:
-    - name: Build wunderctl with race detection enabled and make it available in $PATH
+    - name: Make bin path available in $PATH
       shell: bash
       run: |
-        cd cmd/wunderctl && go install -race
         echo WUNDERCTL_BINARY_PATH=$(go env GOPATH)/bin/wunderctl >> $GITHUB_ENV
+
+    - uses: actions/cache@v3
```
StarpTech, line 11: Cache wunderctl. This saves us up to 2min per run.

.github/workflows/packages-ci.yaml
```diff
@@ -76,6 +76,9 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Run testsuite
```
StarpTech, line 79: Splitting it from `pnpm run` fixes the tests and also make the output more readable.

docs-website/src/pages/docs/guides/index.md
```diff
@@ -13,20 +13,30 @@ Here's an overview of all the available guides.
...
 
+{% quick-link title="Token based authentication with NextAuth.js" icon="core" href="/docs/guides/token-based-authentication-with-next-auth" description="This Guide helps you to configure Token based authentication." /%}
+
 {% quick-link title="Signing origin requests" icon="core" href="/docs/guides/signing-origin-requests" description="Similarly to injecting a short-lived token, this guide helps you to sign requests, which is a requirement for some APIs." /%}
 
 {% quick-link title="Configure Authorization" icon="core" href="/docs/guides/configure-authorization" description="This Guide helps you to configure Authorization with RBAC." /%}
 
-{% quick-link title="TypeSafe API Integrations" icon="core" href="/docs/guides/typesafe_api_integrations_with_typescript_operations_and_graphql" description="Easily integrate APIs with TypeScript Operations & GraphQL" /%}
+{% quick-link title="TypeSafe API Integrations" icon="core" href="/docs/guides/typesafe_api_integrations_with_typescript_operations_and_graphql" description="Easily integrate APIs with TypeScript Operations & GraphQL." /%}
```
StarpTech, line 30: Fixed several links

.github/workflows/packages-ci.yaml
jensneuse: Are you ensuring that we exclude this pipeline (and any other) from the oss metrics?

docs-website/src/pages/docs/guides/mocking.md
```diff
@@ -0,0 +1,145 @@
...
+
+configureWunderGraphApplication({
+  apis: [countries],
+  server,
+  operations,
+  codeGenerators: [
+    {
+      templates: [
+        // Generate all TypeScript based templates, which includes the testing library
+        ...templates.typescript.all,
```
jensneuse, line 33: I suggest that we generate the required typescript templates always, by default.
Can we just always append them to the templates config?
There's no harm in doing so.
It would make the config leaner and examples less verbose.

docs-website/src/pages/docs/guides/mocking.md
```diff
@@ -0,0 +1,145 @@
...
+
+```ts
+import { createTestAndMockServer, TestServers } from '../.wundergraph/generated/testing';
+
+let ts: TestServers;
+
+beforeAll(async () => {
+  ts = createTestAndMockServer({
+    // The directory where your wundergraph directory is located
+    dir: join(__dirname, '..'),
```
jensneuse, line 59: Can you explain `__dirname` for noobs?

docs-website/src/pages/docs/guides/mocking.md
```diff
@@ -0,0 +1,145 @@
...
+    },
+  ],
+});
+```
+
+### Test setup
+
+Within a test, you can import `createTestAndMockServer()` to create the test and mock server instances.
+Note that depending on your project's settings, the path might be slightly different depending on where you generate your templates.
+We use `beforeAll` to start the test and mock server instances. The `ts.start()` method returns a cleanup function to shutdown all servers. In vitest you can return an async function to do a cleanup. We also use `mockedAPIs` to replace the environment variables with the mock server URL.
```
jensneuse, line 44: Add a single line break before "In vitest".

docs-website/src/pages/docs/guides/mocking.md
```diff
@@ -0,0 +1,145 @@
...
+
+beforeAll(async () => {
+  ts = createTestAndMockServer({
+    // The directory where your wundergraph directory is located
+    dir: join(__dirname, '..'),
+  });
+
+  return ts.start({
+    // Environment variables replaced by the test mock server URL
+    mockedAPIs: ['COUNTRIES_URL'],
```
jensneuse, line 64: I understand what you're doing here, but it doesn't feel intuitive.
Is it actually required? Could the "mock" function simply decide by checking the host if it wants to handle the request and if not, we use the "real" origin.

docs-website/src/pages/docs/guides/mocking.md
```diff
@@ -0,0 +1,145 @@
...
+We recommended creating as few testing server instances as possible. These minimizes the number of times the server starts and stops, making your tests faster.
+
+{% callout type="warning" %}
+Due to the nature of the mock server, it is not possible right now to run multiple tests `it` concurrently.
+If you have a demanding test suite, you can create multiple test files and vitest will run them in parallel by default.
+{% /callout %}
+
+#### Mocking HTTP datasources
+
+When you setup a mock with `mock` and the request match, the interceptor is removed internally from the list. This means that you can intercept **2** or more calls to the same URL and return different things on each of them. It also means that you must setup one interceptor for each request you are going to have otherwise it will throw an error because that URL was not present in the interceptor list.
```
jensneuse, line 80: I don't feel like this is intuitive.
The fact that it needs this level of explanation might be a smell that it's too complex.

E.g. when I define a mock, will it get multiple requests? or one? Or one in case it's matched? I'm not sure.

docs-website/src/pages/docs/guides/mocking.md
```diff
@@ -0,0 +1,145 @@
...
+```ts
+const scope = ts.mockServer.mock(
+  async ({ url, method }) => {
+    return url.path === '/' && method === 'POST';
+  },
+  async ({ json }) => {
+    const body = await json();
+
+    expect(body.variables.code).toEqual('ES');
+    expect(body.query).toEqual(
```
jensneuse, line 92: Question: When this fails, will the stack trace point to exactly this position?
I'm using this technique to intentionally "fail" at a certain line of code to identify what the correct assertion should look like to then correct the test.

packages/sdk/src/testing/mock-server.test.ts
```diff
@@ -0,0 +1,190 @@
...
+		expect(resp.headers.get('x-foo')).toEqual('Bar');
+		expect(data).toEqual({
+			getUser: {
+				id: '123',
+			},
+		});
+	});
+
+	test('Should be able to mock the same request multiple times', async () => {
+		let scope = server.mock(
```
jensneuse, line 48: Should be const.

packages/sdk/src/testing/mock-server.test.ts
```diff
@@ -0,0 +1,190 @@
...
+	beforeEach(async () => {
+		server = new WunderGraphMockServer();
+		await server.start();
+	});
+	afterEach(async () => {
+		await server.stop();
+	});
+
+	test('Should be able to mock an endpoint', async () => {
+		const scope = server.mock(
```
jensneuse, line 14: I'm wondering if it's possible to define with the mocks a certain order of requests.
E.g. is it possible with this structure to mock in a way that the request chain MUST be "/foo" and then "/bar" and not the other way around?

packages/sdk/src/testing/mock-server.test.ts
```diff
@@ -0,0 +1,190 @@
...
+
+		const resp = await fetch(`${server.url()}/test`);
+
+		expect(resp.status).toBe(404);
+		expect(scope.isDone).toBe(false);
+		expect(scope.error).toBeDefined();
+		expect(() => scope.done()).toThrow('no interceptor matched for request GET /test');
+	});
+
+	test('Should try the next interceptor when the previous does not match or throws an error', async () => {
```
jensneuse, line 148: In addition, is it possible to use this to mock an "embedded" GraphQL server (wundergraph.server.ts) or an omnigraph REST API (running as part of the fastify server, making requests to OpenAPI origins, but the fetch happens within fastify, not go)

jensneuse (changes requested):

.github/workflows/packages-ci.yaml
StarpTech: No, but I can provide a fix.

docs-website/src/pages/docs/guides/mocking.md
StarpTech: Yes, please create an issue

docs-website/src/pages/docs/guides/mocking.md
StarpTech: It's a global constant that points to the absolute directory of the current file.

docs-website/src/pages/docs/guides/mocking.md
Pagebakers: Yeah I was thinking the same, just generate all required files in the .wundergraph/generate dir by default. 
Out of scope for this though.

docs-website/src/pages/docs/guides/mocking.md
StarpTech: That's exactly how nock works, and it is the most popular mocking solution for Node.js

docs-website/src/pages/docs/guides/mocking.md
Pagebakers: We have this method, we could just check if url starts with `https://countries.trevorblades.com/` for example?

```
 async ({ url, method }) => {
    return url.path === '/' && method === 'POST';
  },
```

StarpTech: Could you elaborate? The node has to request the running mock server. If you would handle all the same then you will lose the functionality to mock individual origins. This approach also does not work when you want to mock an external API like an opensearch server.

docs-website/src/pages/docs/guides/mocking.md
Pagebakers: Does it support methods from nock? like twice() trice()?

packages/sdk/src/testing/mock-server.test.ts
StarpTech: No, I overwrite in the following lines.

docs-website/src/pages/docs/guides/mocking.md
StarpTech: It's very simple maybe we should rephrase it. You can always mock one request. If the request does not match with any mock or the mock throws after it matches an error is thrown when calling `scope.done()`.

docs-website/src/pages/docs/guides/mocking.md
StarpTech: Yes, that's way `scope.done` rethrows them original error.

packages/sdk/src/client/client.ts
Pagebakers: I checked, it's supported everywhere except IE11, which we don't need to support.

packages/sdk/src/testing/mock-server.test.ts
StarpTech: If the order matters, I'd assume it was implemented like that, and the mocks can be written in that order. Other than that, the library does not ensure that, but you can simply use a variable in your tests to validate when each request arrived and check that afterward.

packages/sdk/src/testing/mock-server.test.ts
StarpTech: That's the power of this generic approach. You can mock any HTTP data source as long as you can point your service  URL to an environment variable.

packages/sdk/src/testing/mock-server.test.ts
Pagebakers: scope is set again after this one completes.

docs-website/src/pages/docs/guides/mocking.md
```diff
@@ -0,0 +1,145 @@
...
+Some testing frameworks like Vitest or Jest provides a `globalSetup` hook that allows you to run a script before running the tests. In that way, you can automate it.
+{% /callout %}
+
+```ts
+import { createTestAndMockServer, TestServers } from '../.wundergraph/generated/testing';
+
+let ts: TestServers;
+
+beforeAll(async () => {
+  ts = createTestAndMockServer({
```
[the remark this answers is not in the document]
StarpTech: Yes, it could be. What are your thoughts?

packages/testsuite/apps/basic/openapi-generation.test.ts
```diff
@@ -1,20 +1,7 @@
-import { beforeAll, describe, expect, it } from 'vitest';
-import { createTestServer } from './.wundergraph/generated/testing';
+import { describe, expect, it } from 'vitest';
```
Pagebakers, line 1: Let's move everything to vitest.

docs-website/src/pages/docs/guides/mocking.md
StarpTech: I rephrased it. Please check

packages/testsuite/apps/mock/test/mock-datasource.test.ts
```diff
@@ -0,0 +1,92 @@
...
+	return ts.start({
+		// Environment variables replaced by the test mock server URL
+		mockedAPIs: ['COUNTRIES_URL', 'OS_NODE_URL'],
+	});
+});
+
+describe('Mock http datasource', () => {
+	it('Should mock countries origin API', async () => {
+		// Mock the countries origin API
+		const scope = ts.mockServer.mock(
```
Pagebakers, line 23: Regarding types, we could potentially do something like:
```ts
const scope = ts.mockServer.from('countries_GetCountries').mock(async (request) => {
   const body = await request.json() // typed from the GetCountries input schema
   
   return {
     body: {} // typed from the models?
   }
}
```

We can get the URL from the api namepsace countries, and the path and types from the operation name.

docs-website/src/pages/docs/guides/mocking.md
Pagebakers:
```ts
ts.start({ apiNamespaces: ['countries'] }) // this can be typed
````


Also in one tests you do the above already:
```ts
		const scope = ts.mockServer.assertHTTPConnect(async ({ url }) => {
			return url === 'countries.trevorblades.com:443';
		});
```

Pagebakers (changes requested):

docs-website/src/pages/docs/guides/mocking.md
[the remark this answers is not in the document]
Pagebakers: simply createTestServer

if you configure `mockedAPIs: ['COUNTRIES_URL'],` it also sets up the mock server, for example.

docs-website/src/pages/docs/guides/mocking.md
jensneuse: I could imagine that we add a "middleware" into the apitransport.go which gets activated e.g. when we set an evironment variable, like WG_MOCK_SERVER, which we use to inject the mock server.
We will then call the mock server for every request and run it against the "match" function.
If it matches, the mock server can return the response from the mock function.
If it doesn't match, we can indicate that we don't want to mock the response, e.g. by returning a special response status code or header. In this case, the transport middleware will do a regular request.

With this approach, we could simplify the config as we don't have to mock specific namespaces or URLs, which forces the user to use environment variables for URLs, which might not be desired or not always possible.

Simply match the URL in the matcher function and decide whether you'd like to mock the response or pass it back to the transport to do a regular request.

packages/testsuite/apps/mock/test/mock-datasource.test.ts
jensneuse: We might be confusing something here. These mocks should work on "origin" requests, right?
There are no models for origin requests (yet).
Operations are only known to the client.
One client Operations doesn't necessarily map to one single origin request, it could be many different origin requests.

.github/actions/wunderctl/action.yaml
jensneuse: really nice!

docs-website/src/pages/docs/guides/mocking.md
jensneuse: I suggest to make this clear in the documentation.

packages/sdk/src/testing/mock-server.test.ts
jensneuse: I'm not sure if I see it as a feature that I have to make every URL an environment variable.

StarpTech: How do you configure your node for different environments?

docs-website/src/pages/docs/guides/mocking.md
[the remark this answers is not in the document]
StarpTech: mockedEnvUrls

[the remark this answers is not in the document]
fiam: I would also prefer this approach.

docs-website/src/pages/docs/guides/mocking.md
fiam: Since the namespace needs to be unique, I think a better approach would be referencing the APIs to be mocked by their namespaces, then sending it up to wunderctl via environment variables that override the URL. e.g. `WG_URL_OVERRIDES=countries=http://our-internal-proxy.url:1234`

docs-website/src/pages/docs/guides/mocking.md
```diff
@@ -0,0 +1,148 @@
...
+We recommended creating as few testing server instances as possible. These minimizes the number of times the server starts and stops, making your tests faster.
+
+{% callout type="warning" %}
+Due to the nature of the mock server, it is not possible right now to run multiple tests `it` concurrently.
+If you have a demanding test suite, you can create multiple test files and vitest will run them in parallel by default.
+{% /callout %}
+
+#### Mocking HTTP datasources
+
+When you setup a mock with `mock()` and the request match, the mock server will return the response. If the request does not match, the mock server will return a 404 response and the call to `scope.done()` will fail the test.
```
fiam, line 81: `match` => `matches`

docs-website/src/pages/docs/guides/mocking.md
```diff
@@ -0,0 +1,148 @@
...
+You have also the ability to throw an error inside the response function to fail the mock. This is useful if you want to verify with test assertion that the request is correct. A thrown error is handled as an unmatched request and the next mock will be checked.
```
fiam, line 82: I don't like this, if a handler function throws I think we should fail right there rather than failing later.

docs-website/src/pages/docs/guides/mocking.md
```diff
@@ -0,0 +1,148 @@
...
+
+The first argument of the `mock` function is a predicate that is used to match the request. The second argument is the function that returns the response.
+
+```ts
+const scope = ts.mockServer.mock(
+  async ({ url, method }) => {
```
fiam, line 88: Can we support non-async functions here? It feels like too much ceremony.

packages/sdk/src/client/client.ts
fiam: I would leave this entirely to the caller, via custom fetch fu

Pagebakers (approved):
LGTM

docs-website/src/pages/docs/guides/mocking.md
StarpTech: Done, ty

docs-website/src/pages/docs/guides/mocking.md
StarpTech: We throw directly, but because the request is handled asynchronously, we can't throw it in the handler. It would also make it impossible to add multiple mocks because the first mock that throws would fail the test. We use `expect` (which throws) to validate if a request matches.

StarpTech: We rethrow the error as soon as the user calls `.done` the moment when the user expects that the mock was called.

docs-website/src/pages/docs/guides/mocking.md
StarpTech: Yes, we can. Good point.

packages/testsuite/apps/mock/test/mock-datasource.test.ts
StarpTech: I like this and will keep it in mind when reworking the implementation to consider the Virtual Graph separately.

docs-website/src/pages/docs/guides/mocking.md
StarpTech: Done

docs-website/src/pages/docs/guides/mocking.md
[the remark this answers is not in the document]
StarpTech: We abandon this because it does not work with mocking apis that aren't part of the virtual graph. We use a simpler approach to cover both cases.

packages/sdk/src/testing/mock-server.test.ts
StarpTech: Can we resolve it?

.github/actions/wunderctl/action.yaml
StarpTech: thx

jensneuse (approved):

fiam (approved):
LGTM

PR #894, opened 2023-04-25
fix: nextjs-file-uploads example, re-enable in in tests

fiam:
## Motivation and Context

Script for setting up minio was outdated and didn't work with the current release. Unfortunately,
minio doesn't provide binaries of past mc releases, so we can't pin that.

#### Checklist

- [ ] run `make test`


StarpTech (approved):
LGTM great catch!

jensneuse (approved):

PR #897, opened 2023-04-26
fix: incorrect env read for public node url

thisisnithin:
## Motivation and Context

The env we read for public node url is incorrect. This also broke vercel integration because we read a different env than we set.

## Marketing Changelog

Fixed a bug where the incorrect env was read for the public node url.

#### Checklist

- [x] run `make test`


StarpTech:
Please make a release after merge.

ggeijteman-wg (approved):

PR #901, opened 2023-04-26
feat: add support for operations with hyphens in their names

fiam:
## Motivation and Context

Users have asked to be able to use hyphens in their operation names.

## Marketing Changelog

We now have an alternative to the boring underscore.

#### Checklist

- [ ] run `make test`

### DIFF
--- a/pkg/loadoperations/loader.go
+++ b/pkg/loadoperations/loader.go
@@ -341,7 +341,7 @@ func isValidOperationName(s string) bool {
 		if i == 0 && !unicode.IsLetter(r) {
 			return false
 		}
-		if !unicode.IsLetter(r) && !unicode.IsDigit(r) && r != '/' && r != '_' {
+		if !unicode.IsLetter(r) && !unicode.IsDigit(r) && r != '/' && r != '_' && r != '-' {
 			return false
 		}
 	}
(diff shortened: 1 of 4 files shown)


pkg/loadoperations/loader.go
```diff
@@ -341,15 +341,16 @@ func isValidOperationName(s string) bool {
 		if i == 0 && !unicode.IsLetter(r) {
 			return false
 		}
-		if !unicode.IsLetter(r) && !unicode.IsDigit(r) && r != '/' && r != '_' {
+		if !unicode.IsLetter(r) && !unicode.IsDigit(r) && r != '/' && r != '_' && r != '-' {
```
Aenimus, line 344: surely a regex at this point

fiam:
We want to expand this a bit, changing to draft in the meantime

pkg/loadoperations/loader.go
fiam: ```Some people, when confronted with a problem, think "I know, I'll use regular expressions." Now they have two problems.```

I do agree that the ifs were getting a bit hairy with all the negations, so I've refactored the checks to separate functions. PTAL.

PR #902, opened 2023-04-26
fix: variable corruption

jensneuse:
Copy prepared variables before merging user-defined variables to avoid corruption.

In very rare cases, it's possible that we're overriding the shared variables object in a way that will result in corruption. To avoid this, we're creating a copy first.


StarpTech (approved):
LGTM

PR #911, opened 2023-04-27
feat: add support for custom claims of any type

fiam:
- Define a new ANY type for ValueType
- Allow serializing slices in User data and cookies
- Add a custom claim with an array value in a test


packages/testsuite/apps/auth/.wundergraph/wundergraph.server.ts
```diff
@@ -23,10 +23,13 @@ export default configureWunderGraphServer(() => ({
 						message: 'denied by mutatingPostAuthentication',
 					};
 				}
+				const customClaims = hook.user.customClaims ?? {};
+				customClaims.aTestValue = ['anArray'];
```
jensneuse, line 27: Does it also work with an array of objects?

jensneuse (approved):

packages/testsuite/apps/auth/.wundergraph/wundergraph.server.ts
fiam: It does but you do lose type safety in the handlers if you use `any`, though. Maybe we can consider supporting JSON schema here instead of a reduced list of hardcoded types?

jensneuse: No, I think it's good enough.

PR #914, opened 2023-04-28
feat: context factories

fiam:
This pull request adds support for using a custom context in hooks and TS functions. To use it, the application must declare a factory function when configuring the server:

```typescript
// wundergraph.server.ts
import { GraphQLObjectType, GraphQLSchema, GraphQLString } from 'graphql';
import { configureWunderGraphServer } from '@wundergraph/sdk/server';

export class MyContext {
	greet () { console.log('say hello')}
}

// Function doesn't need to have any specific name nor to be exported
const createContext = async() => {
	return new MyContext();
}

export default configureWunderGraphServer(() => ({
	hooks: {
		queries: {
			Countries: {
				preResolve:async ({operations, context}) => {
					// context here is fully typed
					context.greet();
				}
			}
		},
		mutations: {},
	},
	context: {
         request: {
             create: createContext,
         }
	},
}));
```

This makes the context instance available for functions too:

```typescript

import { createOperation, z } from '../../generated/wundergraph.factory';

export default createOperation.query({
	input: z.object({
		id: z.string(),
	}),
	handler: async ({ input, context }) => {
		context.greet();
		return {
			id: input.id,
			name: 'Jens',
			bio: 'Founder of WunderGraph',
		};
	},
});
```

## Motivation and Context

Several customers have expressed their desire to inject arbitrary objects into the request pipeline. This provides
support for both a global context and request context, satisfying those use cases.

## Marketing Changelog

WunderGraph now supports dependency injection and IoC using context factories. This enables new patterns, like
opening and keeping stateful connections to remote services that can't be mapped to data sources (e.g. a Redis instance, Temporal, ...)

#### Checklist

- [ ] run `make test`


fiam:
@jensneuse @Pagebakers Code is ready, but we need some new docs. How should we document this? A new guide?

fiam:
I went ahead and added a guide

docs-website/config/navigation.tsx
```diff
@@ -437,6 +437,10 @@ const navigation = [
 						title: 'Extend third party GraphQL APIs',
 						href: '/docs/guides/extend-third-party-graphql-apis',
 					},
+					{
+						title: 'Use a custom context',
```
Pagebakers, line 441: Maybe call it `Extend the WunderGraph server context` ?

docs-website/src/pages/docs/guides/custom-context.md
```diff
@@ -0,0 +1,100 @@
...
+    queries: {
+      Countries: {
+        preResolve: async ({ operations, context }) => {
+          context.greet(); // Use your context from a hook
+        },
+      },
+    },
+    mutations: {},
+  },
+  context: makeContext,
```
Pagebakers, line 44: typo, `createContext`

Pagebakers (changes requested):
Found a few small issues.

In the future it'd be great if we can use the same context factory to build the internal context (log, operations, etc), instead of re-using the same code everywhere.

Pagebakers (changes requested):
One more thing, please add it to the wundergraph.server.ts reference too

https://docs.wundergraph.com/docs/wundergraph-server-ts-reference

docs-website/config/navigation.tsx
fiam: I was just thinking, what if we rename the whole document as "Using a Context Factory"?

docs-website/src/pages/docs/guides/custom-context.md
fiam: Thanks! Fixed

docs-website/src/pages/docs/guides/custom-context.md
```diff
@@ -0,0 +1,100 @@
...
```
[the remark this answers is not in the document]
fiam: 👍

StarpTech (changes requested):
The context should be available in all server components.

- [ ] Functions
- [ ] Webhook
- [X] Hooks
- [x] GraphQL servers

jensneuse:
@fiam make sure that the following resources are present and accessible in the create context factory:
- client request
- user (if authenticated)
- operations (client)

docs-website/config/navigation.tsx
fiam: I've renamed using the title you suggested

fiam:
> The context should be available in all server components.
> 
> * [x]  Functions
> * [x]  Webhook
> * [x]  Hooks
> * [x]  GraphQL servers

It was only missing from the webhooks, but I've now fixed that.

fiam:
> @fiam make sure that the following resources are present and accessible in the create context factory:
> 
> * client request
> * user (if authenticated)
> * operations (client)

It now receives the same context as hooks/functions, minus the custom context field

fiam:
> One more thing, please add it to the wundergraph.server.ts reference too
> 
> [docs.wundergraph.com/docs/wundergraph-server-ts-reference](https://docs.wundergraph.com/docs/wundergraph-server-ts-reference)

I've added it

packages/testsuite/apps/basic/.wundergraph/wundergraph.server.ts
```diff
@@ -19,15 +31,33 @@ export default configureWunderGraphServer(() => ({
...
+			// 	mutatingPostResolve: async ({ context }) => {
+			// 		return {
+			// 			data: context.hello(),
+			// 		};
+			// 	},
+			// },
 		},
 		mutations: {},
 	},
+	// This works, declaring return type makes no difference
```
fiam, line 44: @Pagebakers Could you please take a look at this one? Quite the head scratcher. I've outlined when it fails.

fiam: Got it. It was because the import at the top of a type also named `ContextFactoryContext`, but different

fiam: Well, false alarm. That makes it work by passing the explicit parameter but if I remove the explicit parameter it goes back to `any`- I'll continue investigating

fiam: Thanks for the help @Pagebakers, now it's working 👍

StarpTech (changes requested):

docs-website/src/pages/docs/guides/context-factory.md
```diff
@@ -0,0 +1,137 @@
...
+        preResolve: async ({ operations, context }) => {
+          // Use your context from a hook
+          context.greet();
+        },
+      },
+    },
+    mutations: {},
+  },
+  createContext: createContext(async (req) => {
+    return new MyContext();
```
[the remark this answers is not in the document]
fiam: It was requested that the context creation had optional access to the user, the client request and the operation client. Keep in mind that you can have like a "master" context and then make a copy for every request which holds additional per-request data. The block in your comment will work if you wrap `new MyContext()` in a function.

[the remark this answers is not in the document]
StarpTech: Why should we pass all the data again through `createContext` if the user already has access to data where context is available?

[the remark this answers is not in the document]
jensneuse: As a user, I'd like to be able to define per-request dependencies. If we don't enable this, someone will ask for it eventually.

From an implementation point of view, we can simply call the "createContext" function from within the fastify handler before we call an actual hook or a TypeScript operation. This means, the client request and user object is already present. If it's already present, we can just expose it to the user to be able to have more choices in terms of how they want to build the dependencies.

Hook flow: req -> node -> fastify -> create context -> hook
TS Operation flow: req -> node -> fastify -> create context -> ts operation

If we don't expose the data that's already present in the same request to the create context function, you might not be flexible enough in building the dependencies. So, I don't see the point in omitting this.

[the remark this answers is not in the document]
jensneuse: Yes, I think we need the following things:

1. One function to create a global context, persistent across ALL requests, this gets called when the fastify server starts
2. One function that gets called before the fastify server is shut down to clean up the global context
3. One function that gets called right before the we make a function call into a hook or typescript operation
4. One function that gets called right after we make a function call into a hook or typescript operation (for cleanup purposes)

The global persistent hook functions will, obviously, have no request or user, as they are run outside of the request lifecycle.
The request local hooks run immediately before and after calling a hook or typescript function, so we can (and should) pass the client request, user, and operations client, just in case a user needs it.

OnFastifyStart: global create context
OnBeforeFastifyStop: global cleanup context
OnBeforeCallHookOrFunction: local create context
OnAfterCallHookOrFunction: local cleanup context

In addition to the hooks, there should be a centralized way to define and extend the types of the context. Not sure if the hooks themselves can do this, or if we do it in one place.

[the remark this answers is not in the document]
jensneuse: Oh, please don't take the function names literally.

packages/sdk/src/server/server.ts
```diff
@@ -248,9 +286,14 @@ export const createServer = async ({
 						'x-request-id': req.id,
 					},
 				}),
+				context: await createContext(globalContext),
 			};
 		});
 
+		fastify.addHook<{ Body: FastifyRequestBody }>('onResponse', async (req) => {
+			releaseContext(req.ctx.context);
```
StarpTech, line 294: You forgot to await. That result in an unhandled promise rejection on error.

StarpTech (changes requested):

packages/sdk/src/server/server.ts
fiam: Good catch! Fixing

fiam: Fixed now. Thanks!

StarpTech (approved):
LGTM

PR #916, opened 2023-05-01
feat: add support for reading OpenAPI error responses

### LINKED ISSUE #822: Propagate errors from Upstream REST services to the client
This request is based on an older conversation which happened around version "0.138" (before the implementation of openApiV2).

---

The diagram below shows a setup with a Wundergraph-based API gateway and two REST microservices which I'll call "upstream" services.

```
                     |                             +-------------+
                     |                        404  |Microservice1|
+----------+         |          +-----------+ <----+             |
|          |   404              |           |      +-------------+
| Client   | <----------------- | WG        |
|          |         |          |           |      +-------------+
+----------+         |          +-----------+      |Microservice2|
                     |                             |             |
                     |                             +-------------+                 
```

The response code 404 in the diagram demonstrates the idea that an application-specific error generated in the Microservice1 is a "useful" result of an operation which should be somehow propagated to the client.

Our requirements are:
- we want to detect errors from a client (that's possible now)
- we want to distinct between application-specific and system errors (4XX vs 5XX) in the client
- we want to get messages of application-specific errors in the frontend
- we want to get messages of internal server errors in the frontend during the development stage.

---

There are several ways how our HTTP microservices (aka "upstream" services) can fail:
- 4XX statuses - application errors that should have meaningful messages (we want frontend to read those messages)
- 5XX statuses - internal server error; should have clues for debugging (we want the frontend to read those messages during the development phase of the application; in production, we would probably want to replace the actual message with a reference which can be attached to a bug report)

On top of that Wundergraph server might experience difficulties communicating with downstream services in different ways: DNS resolution error, response timeout, and response schema error. Which we probably should treat similarly to 5XX microservices errors.

---

Currently, for 4XX and 5XX errors WG client returns a "null" value and status 200. That way the frontend can understand that there was an error, but there is no additional information about the problem.

fiam:
- Add getHttpResponseError for retrieving upstream HTTP errors
- Add tests for OpenAPI error handling

Fixes #822

Error messages are now available as part of `response.error`, both in hooks and in client responses in the frontend. There are a few examples in the included tests:

```typescript
// client
expect(result.error).toBeInstanceOf(ResponseError);
const error = result.error! as ResponseError;
const httpError = getHttpResponseError(error);
expect(httpError).toBeDefined();
expect(httpError?.statusCode).toBe(404); // HTTP response status code
expect(httpError?.text).toBe(`noteID ${id} not found`); // HTTP response body
expect(httpError?.headers?.['x-note-id']).toBe(id.toString()); // HTTP response headers
```

```typescript
// hook
export default configureWunderGraphServer(() => ({
	hooks: {
		queries: {
			NoteByID: {
				postResolve: async ({ response }) => {
					if (response.errors) {
						const httpError = getHttpResponseError(response.errors);
						console.log('HTTP status code', httpError?.statusCode);
						console.log('HTTP response headers', httpError?.headers);
						console.log('HTTP response body', httpError?.text);
					}
				},
			},
		},
		mutations: {},
	},
}));
```

## Motivation and Context

Reading error responses allows the upstream to send meaningful error messages and
response codes which make their way down to the client.

## Marketing Changelog

OAS support in augmented with extended support for error handling over OAS data sources

#### Checklist

- [x] run `make test`


jensneuse:
How will this work with multiple nested requests? One RPC Operation might map to N root requests and M nested requests. Will it be possible to understand the response status codes and headers for all of them and what field/resolve path they belong to?

fiam:
> How will this work with multiple nested requests? One RPC Operation might map to N root requests and M nested requests. Will it be possible to understand the response status codes and headers for all of them and what field/resolve path they belong to?

We return the extra information from the first error that has it. However, the caller can also iterate over ResponseError.errors and pass an specific one.

packages/sdk/src/client/errors.ts
```diff
@@ -6,6 +6,59 @@ export interface GraphQLErrorLocation {
...
+	 * request sent by the gateway to the OpenAPI upstream
+	 */
+	request?: GraphQLErrorExtensionRequest;
+	/**
+	 * response text sent by the OpenAPI upstream
+	 */
+	responseText?: string;
+}
+
+interface GraphQLErrorExtensions {
```
Pagebakers, line 55: Twice the same interface name?

Pagebakers (changes requested):

packages/sdk/src/client/errors.ts
```diff
@@ -74,13 +127,59 @@ export class InternalError extends OperationError<'InternalError'> {
...
+};
+
 /**
  * The base error class for all client operation errors.
  * This error is thrown when the client receives a response that is not ok.
  * This error should only be used on the client
  */
 export class ResponseError<Code extends ClientOperationErrorCodes | string = string> extends OperationError<Code> {
 	public readonly errors?: GraphQLError[];
+	/**
```
StarpTech, line 180: Remove?

StarpTech (changes requested):

packages/sdk/src/client/errors.ts
fiam: Typo and autocomplete. Fixed.

packages/testsuite/apps/openapi/index.test.ts
```diff
@@ -0,0 +1,108 @@
...
+import { createTestServer } from './.wundergraph/generated/testing';
+import { createOpenAPITestServer } from './test-server';
+import { ResponseError } from '@wundergraph/sdk/client';
+
+const wg = createTestServer({
+	dir: __dirname,
+});
+
+beforeAll(async () => {
+	createOpenAPITestServer(8090);
```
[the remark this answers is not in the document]
fiam: Because I also want to easily run all components manually when debugging

packages/sdk/src/client/errors.ts
fiam: Good catch! Deleted.

packages/testsuite/apps/openapi/index.test.ts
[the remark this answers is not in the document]
StarpTech: Can we improve our current solution? That might be something that our users want as well.

Pagebakers (approved):

packages/sdk/src/client/errors.ts
```diff
@@ -74,6 +127,49 @@ export class InternalError extends OperationError<'InternalError'> {
...
+	if (extensions) {
+		return {
+			statusCode: extensions.http?.status,
+			headers: extensions.http?.headers,
+			text: extensions.responseText,
+		};
+	}
+};
+
+export const getHttpResponseError = (error?: ResponseError | GraphQLError) => {
```
StarpTech, line 160: The error should not be optional if you want to extract something from it.

fiam: This is just for ergonomics. The idea is that you can pass something that’s possibly undefined and only check the return value, to avoid having to check for undefined twice.

StarpTech: I don't like it because it expresses that you get something even in case of undefined, but I will also not block it.

fiam: No strong feelings about it. I'd made it required. PTAL.

packages/testsuite/apps/openapi/.wundergraph/wundergraph.config.ts
```diff
@@ -0,0 +1,46 @@
...
+import operations from './wundergraph.operations';
+
+const notes = introspect.openApiV2({
+	id: 'notes',
+	apiNamespace: 'notes',
+	source: {
+		kind: 'file',
+		filePath: './api.yaml',
+	},
+	baseURL: 'http://localhost:8090/',
```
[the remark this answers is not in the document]
StarpTech: We will address it in a separate PR.

packages/testsuite/apps/openapi/.wundergraph/wundergraph.server.ts
```diff
@@ -0,0 +1,23 @@
+import { configureWunderGraphServer } from '@wundergraph/sdk/server';
+import { getHttpResponseError } from '@wundergraph/sdk/dist/client/errors';
```
StarpTech, line 2: wrong import

packages/sdk/src/client/errors.ts
```diff
@@ -74,6 +127,49 @@ export class InternalError extends OperationError<'InternalError'> {
...
+	if (extensions) {
+		return {
+			statusCode: extensions.http?.status,
+			headers: extensions.http?.headers,
+			text: extensions.responseText,
+		};
+	}
+};
+
+export const getHttpResponseError = (error: ResponseError | GraphQLError) => {
```
StarpTech, line 160: I think you should accept an array of errors in that case you can call without iterate over all yourself.

StarpTech (changes requested):

packages/testsuite/apps/openapi/.wundergraph/wundergraph.server.ts
fiam: Good catch, updated.

packages/sdk/src/client/errors.ts
fiam: I've made it accept a `GraphQLError[]` too PTAL

packages/testsuite/apps/openapi/.wundergraph/wundergraph.server.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export default configureWunderGraphServer(() => ({
+	hooks: {
+		queries: {
+			NoteByID: {
+				postResolve: async ({ response }) => {
+					const error = response?.errors?.[0];
```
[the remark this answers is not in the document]
fiam: I'm now passing the whole array to `getHttpResponseError()`

StarpTech (approved):
LGTM

PR #920, opened 2023-05-03
fix: graphql schema extensions

jensneuse:
## Motivation and Context

This PR enables users to manually extend an introspected GraphQL Schema and resolve it via a mutatingPostResolve hook.

## Marketing Changelog

When dealing with 3rd party GraphQL APIs that are not under your control, you might want to extend the introspected GraphQL schema and add custom logic to resolve a computed field. This is now possible using the `schemaExtension` property on the `introspect.graphql` config, and the `mutatingPostResolve` hook to resolve the field.


Aenimus (approved):

Pagebakers (changes requested):
Good stuff, but no documentation.

Pagebakers (approved):

PR #924, opened 2023-05-03
feat: soap datasource

devsergiy:
## Motivation and Context

This PR adds introspect.soap functionality to be able to include soap data sources

Current limitations:
- WSDL could be supplied only as file
- Endpoint should be specified in the wsdl itself

## Marketing Changelog

Adds the ability to introspect and consume SOAP apis

#### Checklist

- [ ] run `make test`


jensneuse (changes requested):
The code seems LGTM.
2 things I'd like to ask:

- add minimal docs for someone to try it out
- can we get rid of the requirement for the user to set an ID? Could we just generate an id, use a hash, or just a serial number? I feel like it adds unnecessary complexity for the config as we already set the namespace, but don't enforce it to be unique, which is ok. Maybe just generate it?

packages/sdk/src/definition/index.ts
```diff
@@ -57,6 +57,7 @@ export interface ApiIntrospectionOptions {
 	 * Global proxy URL, which might be overridden at the data source level
 	 */
 	httpProxyUrl?: string;
+	apiID?: string;
```
fiam, line 60: Glad to see `ApiIntrospecitonOptions` put to good use 😃

PR #925, opened 2023-05-04
fix: avoid generated typenames conflicts between <opName> and <opName>Internal

fiam:
Refactor the typenames to a single place, then rename the generated types using
a different naming schema that's less likely to cause clashes.


packages/sdk/src/codegen/templates/typescript/internal.client.template.ts
```diff
@@ -9,15 +9,15 @@ import { {{ modelImports }} } from "./models"
...
+        {{operationName}}: ({{#if hasInternalInput}}options: OperationArgsWithInput<{{internalInputTypename}}>{{/if}}) => Promise<{{#if isTypeScriptOperation}}{ data?: {{responseDataTypename}}, errors?: OperationErrors['{{operationPath}}'][] }{{else}}{ data?: {{responseTypename}}['data'], errors?: Required<{{responseTypename}}>['errors'] }{{/if}}>;
     {{/each}}
 {{/if}}
 }
 
 export interface Mutations  {
 {{#if hasInternalMutations}}
     {{#each internalMutations}}
-        {{operationName}}: ({{#if hasInternalInput}}options: OperationArgsWithInput<{{operationName}}Input>{{/if}}) => Promise<{{#if isTypeScriptOperation}}{ data?: {{operationName}}ResponseData, errors?: OperationErrors['{{operationPath}}'][] }{{else}}{ data?: {{operationName}}Response['data'], errors?: Required<{{operationName}}Response>['errors'] }{{/if}}>;
```
fiam, line 20: @jensneuse This was using `<operationName>Input` but that seemed incorrect to me and I changed it to use the type for the internal input (i.e. `Internal<operationName>Input`). PTAL

packages/sdk/src/codegen/templates/typescript/helpers.ts
```diff
@@ -63,11 +71,21 @@ export const liveQueries = (application: ResolvedApplication, includeInternal: b
...
+				injectedInputTypename: operationInjectedInputTypename(op),
+				internalInputTypename: operationInternalInputTypename(op),
 				liveQuery: true,
 				requiresAuthentication: op.AuthenticationConfig.required,
+				responseDataTypename: operationResponseDataTypename(op),
+				responseTypename: operationResponseTypename(op),
 			};
 		});
 
+export const operationInputTypename = (op: GraphQLOperation) => `${op.Name}Input`;
```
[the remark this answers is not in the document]
jensneuse: Have you considered using `#` as part of the name? It makes it extremely unlikely that someone uses it.

packages/testsuite/apps/basic/.wundergraph/operations/nested/InternalWrapper.ts
```diff
@@ -2,6 +2,6 @@ import { createOperation } from '../../generated/wundergraph.factory';
 
 export default createOperation.query({
 	handler: async ({ internalClient }) => {
-		return internalClient.queries.NestedInternal();
+		return internalClient.queries.NestedInternalInternal();
```
jensneuse, line 5: This can be refactored to use the operations client, right?

jensneuse (approved):
great improvements, one suggestions regarding generated names, see comment

packages/testsuite/apps/basic/.wundergraph/operations/nested/InternalWrapper.ts
fiam: We could, but I think it's a good idea to keep it around until we delete the internal client.

packages/sdk/src/codegen/templates/typescript/helpers.ts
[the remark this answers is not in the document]
fiam: I don't think `#` would be valid, but you've made me realize `$` would be a very good choice. I'll update it.

jensneuse (approved):

PR #930, opened 2023-05-04
fix: don't skip second level operations in Postman collections, skip internal

fiam:
Rewrite the algorithm to use a simpler method to build the directories and the
child/parent relationships. Skip internal operations while generating the
collection.


packages/sdk/src/postman/builder.test.ts
```diff
@@ -111,10 +111,49 @@ const operations = [
...
+				count += countPostmanOperations(element);
+			}
+		} else {
+			count += countPostmanOperations(item);
+		}
+	}
+	return count;
+};
+
+test.only('PostmanBuilder', async () => {
```
StarpTech, line 147: I'd remove `only` otherwise, the next person wonders why his test isn't running :smile:

packages/sdk/src/postman/builder.test.ts
```diff
@@ -111,10 +111,49 @@ const operations = [
...
+			type: 'object',
+			properties: {},
+			additionalProperties: false,
+			definitions: {},
+		},
+	},
 ] as any;
 
-test('PostmanBuilder', async () => {
+const countPostmanOperations = (obj: any) => {
```
StarpTech, line 129: Can we replace any with some basic types?

StarpTech (changes requested):
Looks good! Only a few nuts.

packages/sdk/src/postman/builder.test.ts
fiam: 🤦‍♂️

fiam: Thanks! Fixed

packages/sdk/src/postman/builder.test.ts
fiam: Unfortunately, it seems the types are quite messed up. Every field that has a declaration in the root `Collection` is `undefined`. The data seems to be stored inside an `item` property (note the singular), but it's not declared in the type.

jensneuse: Postman SDK types are not in good shape. Let's keep it as is.

jensneuse (approved):
I've tested it locally, good work! LGTM

packages/sdk/src/postman/builder.test.ts
StarpTech: ok

StarpTech (approved):
LGTM

PR #932, opened 2023-05-05
feat: add support for caching in function handlers

fiam:
- Add a cache argument to functions for creating TS operations
- Refactor cache support from QueryHandler into separate types
- Use these new types to support caches from FunctionHandler
- Add test for checking the response headers

## Motivation and Context

GraphQL based functions had configurable cache behaviour. This adds the same features
to TS based operations.

## Marketing Changelog

Added server side support for typescript operations

#### Checklist

- [ ] run `make test`

### DIFF
--- a/packages/testsuite/apps/basic/.wundergraph/operations/functions/greeting.ts
+++ b/packages/testsuite/apps/basic/.wundergraph/operations/functions/greeting.ts
@@ -7,4 +7,9 @@ export default createOperation.query({
 	handler: async ({ input }) => {
 		return 'Hello I am ' + input.name;
 	},
+	cache: {
+		public: true,
+		maxAge: 60,
+		staleWhileRevalidate: 120,
+	},
 });
--- a/packages/testsuite/apps/basic/apicache.test.ts
+++ b/packages/testsuite/apps/basic/apicache.test.ts
@@ -1,0 +1,40 @@
+import { beforeAll, describe, expect, it } from 'vitest';
+import { createTestServer } from './.wundergraph/generated/testing';
+
+const wg = createTestServer({
+	dir: __dirname,
+});
+
+beforeAll(async () => {
+	await wg.start();
+
+	return async () => {
+		await wg.stop();
+	};
+});
+
+const operationUrl = (operationPath: string) => {
+	return wg.url(`/operations/${operationPath}`);
+};
+
+const fetchOperationUrl = (operationPath: string) => {
+	return fetch(operationUrl(operationPath));
+};
+
+describe('API Cache', () => {
+	it('should not have cache headers in handlers without cache', async () => {
+		const resp = await fetchOperationUrl('functions/simple');
+		expect(resp.status).toBe(200);
+		expect(resp.headers.get('cache-control')).toBeNull();
+	});
+
+	it('should have cache headers in handlers cache', async () => {
+		const resp = await fetchOperationUrl('functions/greeting?name=Pepe');
+		expect(resp.status).toBe(200);
+		const cacheControl = resp.headers.get('cache-control');
+		expect(cacheControl).not.toBeNull();
+		expect(cacheControl).toMatch(/public/);
+		expect(cacheControl).toMatch(/max-age=60/);
+		expect(cacheControl).toMatch(/stale-while-revalidate=120/);
+	});
+});
--- a/pkg/apihandler/cache.go
+++ b/pkg/apihandler/cache.go
@@ -1,0 +1,175 @@
+package apihandler
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/cespare/xxhash"
+	"github.com/wundergraph/wundergraph/pkg/apicache"
+	"github.com/wundergraph/wundergraph/pkg/wgpb"
+)
+
+type cacheConfig struct {
+	Enable               bool
+	MaxAge               int64
+	Public               bool
+	StaleWhileRevalidate int64
+}
+
+// cacheHandler holds the cache configuration and data structures for an operation handler.
+// To create a cacheHandler, use newCacheHandler()
+type cacheHandler struct {
+	config cacheConfig
+	cache  apicache.Cache
+	// configHash is the hash of the full application configuration. We use it as part
+	// of the cache keys to ensure any changes to the config invalidate the cache
+	configHash []byte
+}
+
+// newCacheHandler returns a new cacheHandler from an apicache.Cache, an wgpb.OperationCacheConfig representing
+// the cache configuration for an operation and the API configuration hash (typically from Api.ApiConfigHash)
+func newCacheHandler(cache apicache.Cache, config *wgpb.OperationCacheConfig, configHash string) *cacheHandler {
+	return &cacheHandler{
+		config: cacheConfig{
+			Enable:               config.Enable,
+			MaxAge:               config.MaxAge,
+			Public:               config.Public,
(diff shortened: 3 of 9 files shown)


packages/testsuite/apps/basic/.wundergraph/operations/functions/greeting.ts
```diff
@@ -7,4 +7,9 @@ export default createOperation.query({
 	handler: async ({ input }) => {
 		return 'Hello I am ' + input.name;
 	},
+	cache: {
```
StarpTech, line 10: Let's assume the developer makes a decision based on the authenticated user. This would make the result not cacheable, or worse; it could expose data to the public because the user data is not part of the cache.

pkg/apihandler/cache.go
StarpTech, line 1: Please add some tests.

pkg/apihandler/cache.go
```diff
@@ -0,0 +1,231 @@
...
+	configHash []byte
+}
+
+// newCacheHandler returns a new cacheHandler from an apicache.Cache, an wgpb.OperationCacheConfig representing
+// the cache configuration for an operation and the API configuration hash (typically from Api.ApiConfigHash)
+func newCacheHandler(cache apicache.Cache, config *wgpb.OperationCacheConfig, configHash string) *cacheHandler {
+	if config == nil {
+		// XXX: Lots of tests have a nil OperationCacheConfig. For now, patch this
+		// up here.
+		config = &wgpb.OperationCacheConfig{}
```
StarpTech, line 42: This couples the generic cache implementation to the protobuf config. From a design perspective, it would be good if the implementation could work standalone.

pkg/apihandler/cache.go
```diff
@@ -0,0 +1,231 @@
...
+}
+
+func (c *requestCacheHandler) isEnabled() bool {
+	return c != nil && c.handler != nil && c.handler.config.Enable
+}
+
+// ETag returns a en ETag derived from the config hash and the received data
+func (c *requestCacheHandler) ETag(data []byte) string {
+	hash := xxhash.New()
+	_, _ = hash.Write(c.configHash)
```
StarpTech, line 105: Why does the config hash has an impact on the etag? Cache-wise, we don't want to invalidate the cache just because someone changed the config. The cache decision is based on HTTP headers.

pkg/apihandler/internalapihandler.go
```diff
@@ -286,9 +286,14 @@ func (i *InternalBuilder) registerNodeJsOperation(operation *wgpb.Operation, api
 
 	route := i.router.Methods(http.MethodPost).Path(apiPath)
 
+	// Internal operations don't use the cache to store data, but they need it
+	// to calculate proper etags. Create a wgpb.OperationCacheConfig
+	// with all fields set to false/zero.
+	cache := newCacheHandler(nil, &wgpb.OperationCacheConfig{}, i.api.ApiConfigHash)
```
StarpTech, line 292: Any reason not to enable it for internal operations too?

StarpTech (changes requested):

packages/testsuite/apps/basic/apicache.test.ts
```diff
@@ -0,0 +1,40 @@
...
+
+	it('should have cache headers in function handlers with cache', async () => {
+		const resp = await fetchOperationUrl('functions/greeting?name=Pepe');
+		expect(resp.status).toBe(200);
+		const cacheControl = resp.headers.get('cache-control');
+		expect(cacheControl).not.toBeNull();
+		expect(cacheControl).toMatch(/public/);
+		expect(cacheControl).toMatch(/max-age=60/);
+		expect(cacheControl).toMatch(/stale-while-revalidate=120/);
+	});
```
StarpTech, line 39: Could you add a test that validates the staleness?

StarpTech (changes requested):

packages/testsuite/apps/basic/.wundergraph/operations/functions/greeting.ts
fiam: We have no way to detect if an endpoint returns private data, it is the responsibility of the user to not enable public caching in those cases.

pkg/apihandler/cache.go
fiam: I'll introduce an intermediate type

pkg/apihandler/cache.go
fiam: Because changes in the config could change the response. It's the same approach we use in the frontend `applicationHash` - However, I was reviewing if those two were the same value but it seems to me we're not correctly passing that hash back to the `Api`. I'll run a few tests and I'll get back to you.

packages/testsuite/apps/basic/apicache.test.ts
fiam: I'll add it

packages/testsuite/apps/basic/.wundergraph/operations/functions/greeting.ts
StarpTech: As far as I saw, we cache the data regardless of whether it is private or public. We build the cache key only from configHash + requestURI.

pkg/apihandler/cache.go
StarpTech: I can't see any new go tests.

pkg/apihandler/cache.go
StarpTech: Thanks

pkg/apihandler/cache.go
StarpTech: ty

pkg/apihandler/internalapihandler.go
StarpTech: Make sense.

pkg/apihandler/cache.go
fiam: I added integration tests

StarpTech: Those don't help during development. There are very slow compared to unit tests. This module should be shipped with his own tests. Caching is complex.

packages/testsuite/apps/basic/.wundergraph/operations/functions/greeting.ts
fiam: That's right, but we do not know if what's returned and/or cached contains any private information. I understand that having authentication plus a public cache is usually a mistake, but I think whether we should do something about it is out of the scope of the issue/PR.

pkg/apihandler/cache.go
fiam: We're definitely not passing it, but we need to change a few things in the SDK because we generate multiple hashes at different places. I'll open a separate issue to address it.

packages/testsuite/apps/basic/.wundergraph/operations/functions/greeting.ts
fiam: I think you might have misunderstood what was decided for this PR. We're removing the cache storage, so there's no cache keys to calculate.

pkg/apihandler/cache.go
fiam: Added, PTAL

packages/testsuite/apps/basic/apicache.test.ts
fiam: Only thing we can test now is the cache headers, because the storage has been removed

StarpTech (approved):
LGTM

jensneuse (approved):

PR #934, opened 2023-05-07
feat: improve mocking library, support N times and persistent mocks

StarpTech:
This PR simplifies the `.mock` interface to a single object and introduces two new features:

- `match` - A function that returns true if the request matches
- `handler` - A function that returns the response or throws an error
- **`times`** - The number of times the mock should be called. Defaults to 1.
- **`persist`** - If true, the mock will not be removed after the test. Defaults to false.

- Perssitent

```ts
let scope = server.mock({
  times: 2,
  persist: true,
  match: async ({ url, method }) => {
    return url.path === "/test" && method === "GET";
  },
  handler: async () => {
    return {
      body: {
        getUser: {
          id: "123",
        },
      },
    };
  },
});

```

#### Checklist

- [ ] run `make test`


jensneuse (approved):

fiam (approved):
LGTM

PR #937, opened 2023-05-08
feat: add projectID and environmentID to logs

thisisnithin:
## Motivation and Context

We need access to these 2 variables for application logs on the cloud.

### DIFF
--- a/pkg/logging/logging.go
+++ b/pkg/logging/logging.go
@@ -17,6 +17,10 @@ const (
 
 	// logger field name must be aligned with fastify
 	requestIDField = "reqId"
+
+	//	environment variables
+	WgCloudEnvironmentID = "WG_CLOUD_ENVIRONMENT_ID"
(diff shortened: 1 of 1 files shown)


pkg/logging/logging.go
```diff
@@ -50,6 +54,30 @@ func zapConsoleEncoder() zapcore.Encoder {
...
+	if err != nil {
+		host = "unknown"
+	}
+
+	logger = logger.With(
+		zap.String("hostname", host),
+		zap.Int("pid", os.Getpid()),
+	)
+
+	environmentID := os.Getenv(WgCloudEnvironmentID)
```
StarpTech, line 68: Would you like to add the deployment ID here as well?

StarpTech (changes requested):

ggeijteman-wg (approved):

StarpTech (approved):
LGTM

PR #944, opened 2023-05-09
fix: clientrequest not passed to subscriptions in operation client

Pagebakers:
## Motivation and Context

This fixes an issue where the clientRequest is not passed to subscriptions in the operations client.

#### Checklist

- [ ] run `make test`


packages/sdk/src/client/client.ts
```diff
@@ -486,17 +484,24 @@ export class Client {
...
+		subscription: SubscriptionRequestOptions
+	): AsyncGenerator<ClientResponse<Data, Error>> {
+		const response = await this.fetchSubscription(subscription);
 
 		if (!response.ok || response.body === null) {
 			yield {
 				error: new ResponseError({
 					code: 'ResponseError',
-					message: `Response is not ok. Failed to subscribe to '${url}'`,
+					message: `Response is not ok. Failed to subscribe to '${subscription.operationName}'`,
```
StarpTech, line 504: maybe `Failed to subscribe to the subscription with name:` ?

Pagebakers: I think it's better to be short and concise, it clear that you subscribe to an operation (name) when using the client.

StarpTech: ok

PR #947, opened 2023-05-10
feat: re-enable plan cache for graphql handler

jensneuse:
@fiam I'd like to ask you for 3 things before we switch from draft to open
- [x] add testing capabilities to the testing framework for the GraphQL Endpoint
- [x] add 2 basic tests for the GraphQL handler
- [x] add another test with concurrency that ensures the gateway behaves correctly with concurrent requests

## Motivation and Context

Re-enables a planning cache for the GraphQL handler.

## Marketing Changelog

More and more users leverage the GraphQL Endpoint.
We wanted to make sure that using the Endpoint will scale, so we've implemented a planning cache with LRU capabilities, allowing the planning cache to grow until a certain point and then garbage collect less used plans to keep memory usage limited.

Fixes ENG-2088


fiam:
@jensneuse Anything we should add in this PR?

devsergiy (approved):

PR #948, opened 2023-05-10
feat(sdk): add custom fetch options

Hebilicious:
# Add custom fetch options

## Motivation and Context

The wundergraph client is incompatible with cloudflare workers/pages. Some runtime like workers [don't support](https://github.com/cloudflare/workerd/blob/902b65ace3c5c0f5ae4b432fae976f28ddaec1ec/src/workerd/api/http.h#L510-L517) some of the fetch options that are injected by the wundergraph client and that can't be overwritten by a customFetch.

This can be tested easily in the cloudflare workers [playground](https://cloudflareworkers.com/#76e201bbfa9a01b3fc7cc29e97096465:about:blank)

This PR adds a way to pass custom fetch options to the wundergraph client to give more flexibility.

## Marketing Changelog

Add custom fetch options. 
This makes the wundergraph client compatible with cloudflare workers/pages.

#### Checklist

- [x] run `make test`

### DIFF
--- a/packages/sdk/src/client/client.ts
+++ b/packages/sdk/src/client/client.ts
@@ -121,6 +121,7 @@ export class Client {
 
 	private async fetch(input: RequestInfo, init: RequestInit = {}): Promise<globalThis.Response> {
 		const fetchImpl = this.options.customFetch || globalThis.fetch;
+		const customFetchOptions = this.options.customFetchOptions || {};
 
 		init.headers = {
 			...this.baseHeaders,
@@ -138,9 +139,8 @@ export class Client {
 
 		try {
 			const resp = await fetchImpl(input, {
-				credentials: 'include',
-				mode: 'cors',
 				...init,
+				...customFetchOptions,
 			});
 			return resp;
 		} finally {
--- a/packages/sdk/src/client/client.ts
+++ b/packages/sdk/src/client/client.ts
@@ -121,7 +121,7 @@ export class Client {
 
 	private async fetch(input: RequestInfo, init: RequestInit = {}): Promise<globalThis.Response> {
 		const fetchImpl = this.options.customFetch || globalThis.fetch;
-		const customFetchOptions = this.options.customFetchOptions || {};
+		const customFetchOptions = this.options.customFetchOptions || { credentials: 'include', mode: 'cors' };
 
 		init.headers = {
 			...this.baseHeaders,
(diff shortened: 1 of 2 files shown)


jensneuse (changes requested):
With your change, you modified the default behavior. Can you change the implementation so that the default behaviour stays as is, but you're able to change the fetchOptions, e.g. by removing credentials and mode if you don't need it.

jensneuse:
> > With your change, you modified the default behavior. Can you change the implementation so that the default behaviour stays as is, but you're able to change the fetchOptions, e.g. by removing credentials and mode if you don't need it.
> 
> Done. However, I personally think it's better if it works out of the box, as many people will encounter this issue in the future. In terms of security, the browser defaults look safe enough...
> 
> Maybe this needs to be documented somewhere ?

Let me in investigate in various browsers what happens if we drop it.

Pagebakers:
How about doing this on cloudflare.

```ts
const client = createClient({
  customFetch: (init) => {
    const {mode, credentials, ..._init} = init
    return fetch(_init)
  }
})
```

Hebilicious:
@Pagebakers haven't had the chance to try this yet as I've been using my fork. But this proposed change isn't breaking, just add an easier way to pass custom fetch options.

Hebilicious:
> Can you share the fetch config you use on cloudflare workers?

I'm using a pnpm patch

```
diff --git a/dist/client/client.js b/dist/client/client.js
index d1dd3698a33f60273dbb8a7c28d5d55a4b2c724e..5b275b14a58e94419178a8606e79132a987167a2 100644
--- a/dist/client/client.js
+++ b/dist/client/client.js
@@ -50,8 +50,6 @@ class Client {
             ...init.headers,
         };
         return fetchImpl(input, {
-            credentials: 'include',
-            mode: 'cors',
             ...init,
         });
     }
```

Pagebakers:
We looked into this again and decided to go a slightly different approach for now, since we don't want to expose both customFetch and customFetchOptions. #1137 should fix support for Cloudflare workers.

Thanks again for opening this PR, we really appreciate your input. 🙏

Hebilicious:
@Pagebakers Thanks pagebreaks, the fix in #1137 should allow me to get rid of my pnpm patch 🙏🏽 
Still, I'm surprised that no-one else had the issue ... Was I the only one using the client on cloudflare ?

PR #959, opened 2023-05-13
feat: improve Cache-Control support

fiam:
- Declare cache as public by default for anonymous operations, private
for authenticated ones
- Include a must-revalidate header by default
- Allow disabling max-age and stale-while-revalidate

## Motivation and Context

By default, we were previously omitting any Cache-Control headers. This can cause problems with overeager caching by some CDN

## Marketing Changelog

Support for Cache-Control headers has been improved with better defaults, adaptability to change depending on the incoming request and more flexibility.

#### Checklist

- [x] run `make test`


jensneuse:
@fiam add marketing section to description maybe?

Pagebakers (changes requested):
Oh this is very nice!

We should also reflect this change in the docs
https://docs.wundergraph.com/docs/wundergraph-operations-ts-reference/configure-caching

jensneuse (approved):

packages/sdk/src/configure/operations.ts
```diff
@@ -13,18 +13,46 @@ export interface BaseOperationConfiguration {
...
+	/**
+	 * Mark responses as public or private. Set to undefined to omit. Notice that
+	 * any authenticated operation will override this header and set it to private.
+	 *
+	 * @default Public for operations without authentication, private otherwise
+	 */
+	public?: boolean;
+	/**
+	 * max-age component of the Cache-Control header.
+	 * Set to undefined to omit this component.
```
StarpTech, line 31: Does `0` works too?

StarpTech (approved):
LGTM

packages/sdk/src/configure/operations.ts
fiam: It does, I've added a test for it too.

Pagebakers (approved):

PR #973, opened 2023-05-16
fix: wait for the hooks server to start up before listening on the node

fiam:
Hooks server and node were started in parallel. This causes a problem
when the node is ready and starts listening before the hooks server has
had time to do its own setup and opening its port, because operations that
call into the hooks server will fail inmmediately due to the connection
to the hooks server being refused.

To fix this, wait for the hooks server to be ready (by connecting to
its port) during startup for wunderctl up and wunderctl start (iff the
hooks server is also started by us).

## Motivation and Context

This has caused problems for apps using OpenAPIv2, since its queries are routed through
the hooks server.

#### Checklist

- [ ] run `make test`


Aenimus (approved):

StarpTech:
From a architectural perspective this task should be done by the load balancer. Passing a request only to a WunderNode if the healthcheck is successful. The WunderNode healthcheck endpoint only returns 200 when the hooks server health endpoint returned with 200 as well.

If fly doesn't doesn't guarantee that at their proxy we can ensure it on our heindall.

fiam:
> From a architectural perspective this task should be handled from the load balancer. Passing a request only to a WunderNode if it has been registered with a successful healthcheck. The WunderNode healthcheck endpoint only returns 200 when the hooks server health endpoint returned with 200 as well (until max retry).
> 
> If fly doesn't doesn't guarantee that at their proxy (revisit?) we can ensure it on our heimdall.
> 
> Did you also try to increase the timeout of the healthcheck request to the hooks server?

I've verified that the fly load balancer doesn't wait for a successful  health-check when resuming the node, so that doesn't solve the problem, and in any case that wouldn't solve the problem for standalone deployments.

We can't check that in Heimdall either because we do not know when a request is going to wake up a node. We could arrange for the node to notify Heimdall before it goes to sleep, somehow keep that state and poll the health endpoint if we believe the request will wake up the node but that is more complicated than needed.

The node has all the information because it knows that it has just started and also knows the hooks server URL. This way, once we start listening on the port everything is ready to dispatch requests. The only drawback is a small startup penalty for operations that were not going to use the hooks server anyway.

I've refactored the PR to reuse the existing `hooks.Client` instead of implementing and ad-hoc check. PTAL. Thanks!

pkg/node/node.go
```diff
@@ -356,6 +365,40 @@ func (n *Node) HandleGracefulShutdown(gracefulTimeoutInSeconds int) {
...
+		return true
+	}
+	fmt.Println("HC START", time.Now())
+	initialHealthCheckCtx, cancel := context.WithTimeout(context.Background(), initialHealthCheckWaitTimeout)
+	defer cancel()
+	for {
+		select {
+		case <-initialHealthCheckCtx.Done():
+			n.log.Warn("hooks server didn't start after initial wait", zap.Duration("duration", initialHealthCheckWaitTimeout))
+			fmt.Println("HC CRAP", time.Now())
```
StarpTech, line 391: Remove

pkg/node/node.go
```diff
@@ -356,6 +365,40 @@ func (n *Node) HandleGracefulShutdown(gracefulTimeoutInSeconds int) {
...
+			return false
+		default:
+			if n.hooksServerHealthIsOk(initialHealthCheckCtx, hooksClient) {
+				fmt.Println("HC OK", time.Now())
```
StarpTech, line 395: Here too

pkg/node/node.go
```diff
@@ -356,6 +365,40 @@ func (n *Node) HandleGracefulShutdown(gracefulTimeoutInSeconds int) {
...
+	initialHealthCheckCtx, cancel := context.WithTimeout(context.Background(), initialHealthCheckWaitTimeout)
+	defer cancel()
+	for {
+		select {
+		case <-initialHealthCheckCtx.Done():
+			n.log.Warn("hooks server didn't start after initial wait", zap.Duration("duration", initialHealthCheckWaitTimeout))
+			fmt.Println("HC CRAP", time.Now())
+			return false
+		default:
+			if n.hooksServerHealthIsOk(initialHealthCheckCtx, hooksClient) {
```
StarpTech, line 394: This will make requests as fast as the machine is, right? This will flood the server. We should add a short delay of 50ms.

pkg/node/node.go
```diff
@@ -421,6 +461,8 @@ func (n *Node) startServer(nodeConfig *WunderNodeConfig) error {
 
 	hooksClient := hooks.NewClient(nodeConfig.Api.Options.ServerUrl, n.log)
 
+	n.hooksServerInitialHealthCheck(hooksClient)
```
StarpTech, line 464: Please add a comment about it.

pkg/node/node.go
```diff
@@ -87,18 +88,26 @@ type options struct {
...
-func WithHooksServerHealthCheck(timeout time.Duration) Option {
+// WithHooksServerHealthCheck enables querying the hooks server during healthchecks
+// when the /health endpoint is hit. If initialCheck is true, it also enables
+// a check during startup to ensure the node doesn't listen until the hooks server
+// is ready
+func WithHooksServerHealthCheck(timeout time.Duration, initialCheck bool) Option {
 	return func(options *options) {
-		options.hooksServerHealthCheck = true
-		options.healthCheckTimeout = timeout
+		options.hooksHealthCheck.periodic = true
```
StarpTech, line 108: Not needed if it can't be configured and if it is the default.

pkg/node/node.go
```diff
@@ -356,6 +365,40 @@ func (n *Node) HandleGracefulShutdown(gracefulTimeoutInSeconds int) {
...
+// hooksServerInitialHealthCheck performs the initial health check for the hooks server
+// if hooksHealthCheck.initial is set, otherwise it returns immediately. It returns wether
+// a successful healthcheck could be completed
+func (n *Node) hooksServerInitialHealthCheck(hooksClient *hooks.Client) bool {
+	if !n.options.hooksHealthCheck.initial {
+		// Assume ok
+		return true
+	}
+	fmt.Println("HC START", time.Now())
+	initialHealthCheckCtx, cancel := context.WithTimeout(context.Background(), initialHealthCheckWaitTimeout)
```
StarpTech, line 385: Do we need it, if we can just retry w. max-retry it until it succeeds?

StarpTech: I also found out that not all methods of the `hooksClient` use the `httpClient` of the struct.

StarpTech (changes requested):

pkg/node/node.go
fiam: Ooops. Thanks for pointing that out. I've deleted this and the other two print statements

pkg/node/node.go
fiam: I would refactor that at a later point. I remember you and I discussed this thing a few weeks ago in the context of the testing SDK (the hooks server not being health-checked in some cases)

pkg/node/node.go
fiam: The request will be retried using the underlying `retryablehttp.Client`, so it shouldn't max the CPU.

pkg/node/node.go
fiam: I've added a comment explaining what we do in case of failure. PTAL

pkg/node/node.go
StarpTech: Good point, in order to not slow down the cold start the timeout must be low but the max retry high otherwise we waste time. We put a lot of effort into this.

This also means we will wait for the server to start before the WunderNode is started. This also has an effect on the cold start. Please test it and share your results. Currently, we have cold start of around 600ms with a hello world wundergraph app.

fiam: If we're going to use that as a measure to approve PRs, the first thing we need is a documented and repeatable way of producing those numbers. I'm not getting numbers in the same ballpark when measuring TTFB without his PR applied.

StarpTech: The agreement is to not make it worse than before but having a script would be beneficial.

> I'm not getting numbers in the same ballpark when measuring TTFB without this PR applied.

What do you mean by that?

PR #978, opened 2023-05-20
feat: add orm client to typescript webhooks

leoalves:
## Motivation and Context

Currently Typescript Webhooks do not expose the new ORM.  This pulls request adds the ORM to the Webhook context

#### Checklist

- [ ] run `make test`


jensneuse (approved):
I think it looks good overall. We might want to add a test to the teatsuite package that uses the ORM within a webhook to fixate this feature.

leoalves:
> I think it looks good overall. We might want to add a test to the teatsuite package that uses the ORM within a webhook to fixate this feature.

I can try to do that. Just need to figure it out how the test suite works.  I probably won't have time this weekend but should be able to do it Monday.
Thanks

leoalves:
I added a new project to the testsuite package (`/packages/testsuite/apps`) to test the addition of the ORM to the webhook.  But when the webhook calls the GraphQL server it returns 404. My guess is that the test server doesn't have the GraphQL server.  The test just throws an error that it's not able to form the JSON. This might be a bug in the ORM package, it's not validating if the response has the `.data` key or not.  

Is there another way to test it the ORM in the webhook?  I saw that there is another folder (`/testapps/default`).  In that project webhook I added a call to the ORM.  But I am not sure how to test it.  I didn't see anywhere where those operations are called from a test file.   

Thanks

jensneuse (approved):

Aenimus (approved):

jensneuse:
@leoalves thank you for contributing this! We'll make a release soon.

pkg/apihandler/internalapihandler.go
```diff
@@ -124,38 +121,18 @@ func (i *InternalBuilder) BuildAndMountInternalApiHandler(ctx context.Context, r
 		}
 	}
 
-	if api.EnableGraphqlEndpoint {
```
leoalves, line 127: Does this mean the `/graphql` endpoint will be public now?

jensneuse: No, it's internal.

PR #982, opened 2023-05-21
feat: propagate authentication, client request and key headers between operations

fiam:
- Add authentication middleware to internal operations
- Forward Authentication and X-Request-ID headers between operations
- Decode the client request correctly before passing it to function handlers
- Fix returning falsy data from TS operations
- Add tests for all these

Fixes ENG-2216
Fixes ENG-2215
Fixes ENG-1383
Fixes ENG-1179
Fixes ENG-1073

## Motivation and Context

Internal operations didn't support these features that external operations did.
Additionally, some features didn't work properly when they involved multiple
operation calls.

## Marketing Changelog

Authentication, request logging and accesing client headers now work with internal
operations.

#### Checklist

- [ ] run `make test`


packages/sdk/src/client/client.ts
```diff
@@ -166,7 +166,7 @@ export class Client {
 			};
 		}
 
-		if (!resp.data) {
+		if (resp.data === undefined) {
```
Pagebakers, line 169: So null can be valid?

Pagebakers (approved):

packages/sdk/src/client/client.ts
fiam: I think so, an operation might return null and we should handle that gracefully. Maybe I'm missing something? 😅 Keep in mind that we use this to handle return values from TS operations too, and they don't necessarily have to abide by GQL conventions.

pkg/apihandler/authentication.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	if c := loadvariable.String(api.AuthenticationConfig.CookieBased.CsrfSecret); c != "" {
+		csrfSecret = []byte(c)
+	} else if fallback := api.CookieBasedSecrets.CsrfSecret; fallback != nil {
+		csrfSecret = fallback
+	}
+
+	if api == nil || api.HasCookieAuthEnabled() && (hashKey == nil || blockKey == nil || csrfSecret == nil) {
+		panic("API is nil or hashkey, blockkey, csrfsecret invalid: This should never have happened. Either validation didn't detect broken configuration, or someone broke the validation code")
```
StarpTech, line 46: Why not return an error?

StarpTech (changes requested):

pkg/apihandler/authentication.go
fiam: I'd say whoever added that block intended the API misuse to be very evident. You can ignore an error, but it's way more difficult to ignore a panic. I don't usually I don't think it's a bad idea.

StarpTech: Panic should be only used when there is no other way. Here we can simply return an error and let the callee decide.

StarpTech (approved):
LGTM

fiam:
Thanks @StarpTech!

PR #1007, opened 2023-05-27
fix: use Node.js fetch that supports web stream api

StarpTech:
The cross-fetch package does not provide a web-streams compatible API required by the base client implementation. This cause `response.body.getReader is not a function` when starting a subscription with the operations client. We can provide different implementations for the Web and Server or use a web-stream compatible fetch on the server as well. I choose the second option.

This PR switched to `undici` as default fetch implementation on the server. It has support for Node.js 16+. As soon as Node.js 16 is out of maintenance, we can remove the package because, starting with Node.js 18, the exact fetch implementation is exposed as global.

#### Checklist

- [ ] run `make test`

### DIFF
--- a/packages/sdk/src/server/operations-client.ts
+++ b/packages/sdk/src/server/operations-client.ts
@@ -1,4 +1,3 @@
-import fetch from 'cross-fetch';
 import {
 	Client,
 	ClientConfig,
@@ -60,10 +59,9 @@ export class OperationsClient<
 	protected readonly clientRequest: any;
 
 	constructor(options: OperationsClientConfig) {
-		const { clientRequest, customFetch = fetch, ...rest } = options;
+		const { clientRequest, ...rest } = options;
 		super({
 			...rest,
-			customFetch,
 		});
 
 		this.clientRequest = clientRequest;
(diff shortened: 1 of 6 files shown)


JivusAyrus (approved):

packages/sdk/src/server/operations-client.ts
```diff
@@ -60,10 +59,9 @@ export class OperationsClient<
 	protected readonly clientRequest: any;
 
 	constructor(options: OperationsClientConfig) {
-		const { clientRequest, customFetch = fetch, ...rest } = options;
+		const { clientRequest, ...rest } = options;
```
Aenimus, line 62: Eelco literally just fixed this a few days ago.
https://discord.com/channels/738739428314316823/738740476940910692/1110856784924659712

Aenimus: https://linear.app/wundergraph/issue/ENG-2321/customfetch-not-passed-down-to-the-base-client-in-operationclient

Pagebakers (changes requested):
It would be better to fall back to a polyfill solution if fetch is undefined, but i believe cross fetch does this already. 

We can plan to remove it and make node 18 the minimal node version, but im not sure its time for thast yet.

packages/sdk/src/server/operations-client.ts
StarpTech: Nothing has changed except using a fetch implementation that also work with web streams

StarpTech:
> It would be better to fall back to a polyfill solution if fetch is undefined, but i believe cross fetch does this already.
We can plan to remove it and make node 18 the minimal node version, but im not sure its time for thast yet.

Done. I use undici as polyfill on the server

Aenimus (approved):

Pagebakers (approved):

PR #1009, opened 2023-05-28
Temp: Bump the version of prisma binaries by using a fork (this is temporary pr as a placeholder)

rnbokade:


jensneuse:
Does this enable views? What version of Prisma does it use? Have you tested it?

jensneuse:
Already solved and released by another PR.

PR #1013, opened 2023-05-29
fix: hook server strictness

Pagebakers:
## Motivation and Context

This fix makes the hook server config stricter so any user errors in the config are highlighted.

#### Checklist

- [ ] run `make test`


testapps/nextjs/.wundergraph/wundergraph.server.ts
```diff
@@ -13,7 +13,24 @@ const testEnum = new GraphQLEnumType({
 	},
 });
 
+export class MyContext {
```
StarpTech, line 16: Anyway, how can we prevent a regression in the future? A test?

Pagebakers: There are 2 ways to make the config 100% strict so our type tests will pick up on everything, but we need to discuss it because how the server config is implemented hasn't changed in over a year (before I started). The problem is now we cannot be too strict, otherwise the initial config generation will not work.

StarpTech: Ok, let's discuss a potential solution tomorrow.

StarpTech: What exact scenario does this PR improve?

testapps/nextjs/.wundergraph/wundergraph.server.ts
```diff
@@ -44,13 +61,13 @@ export default configureWunderGraphServer(() => ({
 				},
 				mutatingPostResolve: async (hook) => {
 					console.log('###mutatingPostResolve', hook);
```
StarpTech, line 63: Please use the sdk logger

StarpTech (changes requested):

jensneuse (approved):

testapps/nextjs/.wundergraph/wundergraph.server.ts
Pagebakers: Can you open an issue for this? it's not something I changed in this PR, it's used on other tests apps also.

StarpTech (approved):
LGTM

fiam (approved):
LGTM

PR #1015, opened 2023-05-29
feat: add graphql enums to generated typescript model

### LINKED ISSUE #960: Add GraphQL enums to generated typescript model
## Problem
The code generated typescript `models.ts` file does not include enums.

My GraphQL schema has the following enum in it:
```
enum RoomStateTypesEnum {
  FLOOR
  SHADOW
  WALL
}
```
But they get inline'd in the `model.ts`:
```
export interface RoomsInsertInput {
  ...
  knownState?: "FLOOR" | "SHADOW" | "WALL";
  realState?: "FLOOR" | "SHADOW" | "WALL";
}
```

Mainly this is for cases where I need to pass in the valid options as an argument for a helper function of some sort, eg this function to determine the neighbors in a 2D array of cells (a maze):
```
const frontiers = (
        startY: number,
        startX: number,
        state: RoomStateTypesEnum,
        maze: RoomsInsertInput[][],
        distance: number = 1
    ): number[][]
```

Functionally, it's important for me to be able to pass in the state as a valid enum-conforming parameter. If this isn't emitted in the typescript model, then I need to either pull the definition from somewhere else it is used (eg, like `admin_RoomStateTypesEnumComparisonExp` in my `model.ts`) or to manually type out the parameter as `"FLOOR" | "SHADOW" | "WALL"`. Manually typing it leads to lots of upkeep and maintenance in the future since the enum contents can change.

## Suggested solution

Ideally, I'd like to see the enums represented in the `model.ts` file as standalone definitions instead of being functionally inlined.

The graphql-code-generator I was using before (from https://the-guild.dev/graphql/codegen/docs/getting-started) would emit them as typescript enums ala:
```
export enum RoomStateTypesEnum {
  Floor = 'FLOOR',
  Shadow = 'SHADOW',
  Wall = 'WALL'
}

export type RoomsInsertInput= {
  ..
  knownState?: RoomStateTypesEnum;
  realState?: RoomStateTypesEnum;
}
```

thisisnithin:
## Motivation and Context

Resolves #960

## Marketing Changelog

GraphQL enum types are now generated as standalone types which can be imported directly.

#### Checklist

- [x] run `make test`


Aenimus (approved):

StarpTech:
@thisisnithin could you add an example or extend one how it will be used?

Pagebakers (approved):

PR #1033, opened 2023-06-01
feat: make logger in hooks/functions/webhooks use the same semantics as console.log

fiam:
pino's default formatter ignores extra arguments, which is different from console.log()
and friends, which print unused arguments during formatting separated by spaces after
the message. Using a pino hook, we can do the formatting ourselves by calling util.format()
and finally pass only the formatted string to pino.

## Motivation and Context

Ignoring extra arguments might cause some of the arguments to the logging functions to
be lost. This change brings the loggers available in hooks/webhooks/operations to behave
almost in the same way as console.log() while still allowing structured logging as
provided by pino.

#### Checklist

- [ ] run `make test`

### DIFF
--- a/packages/sdk/src/logger/index.ts
+++ b/packages/sdk/src/logger/index.ts
@@ -1,3 +1,4 @@
+import { format } from 'node:util';
 import { ConfigurationVariable, LogLevel, logLevelFromJSON } from '@wundergraph/protobuf';
 import { resolveConfigurationVariable } from '../configure/variables';
 import { DestinationStream, pino } from 'pino';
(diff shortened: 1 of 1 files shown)


Aenimus (approved):

packages/sdk/src/logger/index.ts
```diff
@@ -56,6 +57,29 @@ const initLogger = (destination: DestinationStream): pino.Logger => {
 				return { level: label };
 			},
 		},
+		hooks: {
```
StarpTech, line 60: I don't like this approach. We misused the logger because the interface wasn't strict enough. [Check out](https://github.com/pinojs/pino/blob/master/examples/basic.js#L12) how powerful and straightforward Pino is after you know how it should work. Internally, we should use pino as it is, but for the hooks, operations, we can provide a simpler interface e.g.

```ts
interface LogObject extends object {
  // The default msg property in Pino
  message: string
  // The default error property in Pino
}
interface LogFn {
  <T extends LogObject>(T): void;
 (msg: string): void;
}
```

This will work without any hook.

For formatting we could expose `infof` ... methods similar to zap.

StarpTech (changes requested):

PR #1035, opened 2023-06-02
feat: add support for user-defined session durations

fiam:
Extend the User type to include an optional "expires" field, which can be
altered from the authentication hooks. Once this timestamp is reached,
revalidate the session using the revalidate hook. If any of the hooks
returns an expired user, authentication will be rejected based on the
expiry.

## Motivation and Context

This allows applications to define their own session duration, applying their custom validation and re-authentication logic

## Marketing Changelog

Session duration is now configurable instead of having a predefined duration

#### Checklist

- [x] run `make test`


pkg/authentication/authentication.go
```diff
@@ -180,6 +187,10 @@ type User struct {
 	ZoneInfo          string `json:"zoneInfo,omitempty"`
 	Locale            string `json:"locale,omitempty"`
 	Location          string `json:"location,omitempty"`
+	// Expires indicate the unix timestamp in milliseconds when this User is
+	// considered as expired. This can only be set from the authentication
+	// hooks.
+	Expires *int64 `json:"expires,omitempty"`
```
jensneuse, line 193: Is this UTC? or how do we handle time zones between Go and NodeJS?

pkg/authentication/authentication.go
```diff
@@ -263,6 +275,12 @@ func (u *User) copyWellKnownClaim(claim string, from *User) bool {
 	return true
 }
 
+// HasExpired returns true iff the user has expired, as configured by the
+// authentication hooks (via User.Expired)
+func (u *User) HasExpired() bool {
+	return u != nil && u.Expires != nil && *u.Expires > 0 && time.UnixMilli(*u.Expires).Before(time.Now())
```
jensneuse, line 281: can you refactor this to be multi-line and simpler to read?

pkg/authentication/authentication.go
fiam: Unix timestamps are [always in UTC](https://en.wikipedia.org/wiki/Unix_time). In the Go side, we use [time.UnixMilli](https://pkg.go.dev/time#UnixMilli) and in the Node side users can rely on [Date.now()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now) to obtain the current time.

jensneuse (approved):

packages/testsuite/apps/auth/.wundergraph/wundergraph.server.ts
```diff
@@ -50,6 +68,18 @@ export default configureWunderGraphServer(() => ({
...
 				}
+				if (hook.user.userId === 'expiresIn1Second') {
+					throw new Error('should not need to revalidate');
+				}
+				if (hook.user.userId === 'expiresNow') {
+					return {
+						status: 'ok',
+						user: {
+							...hook.user,
+							expires: Date.now() + 60 * 1000,
```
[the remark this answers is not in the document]
fiam: No, the idea is to make `postAuthentication` create an expired session to ensure that the `revalidate` hook extendsits duration, then in the test we [check that the session has 50 seconds or more left](https://github.com/wundergraph/wundergraph/pull/1035/files#diff-a1f1496a0d8b93e7678b3b7656d7143956394d13f86bfcbc01b956946af3def3R521) (to leave some margin for variances in timing). Regarding the name, I think `expires` is better since that's the nomenclature that cookies use.

ggeijteman-wg (approved):

packages/testsuite/apps/auth/.wundergraph/wundergraph.server.ts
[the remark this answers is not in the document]
fiam: Thanks @ggeijteman-wg!

PR #1042, opened 2023-06-03
fix: align otel attributes

StarpTech:
## Motivation and Context

This PR aligns several OTEL attributes. Previously, we added client attributes in a server context.

#### Checklist

- [ ] run `make test`

### DIFF
--- a/pkg/trace/wrap_handler.go
+++ b/pkg/trace/wrap_handler.go
@@ -5,6 +5,7 @@ import (
 
 	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
 	"go.opentelemetry.io/otel/attribute"
+	semconv "go.opentelemetry.io/otel/semconv/v1.17.0"
 	"go.opentelemetry.io/otel/trace"
 )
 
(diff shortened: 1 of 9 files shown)


pkg/trace/wrap_handler.go
```diff
@@ -14,13 +15,18 @@ func WrapHandler(wrappedHandler http.Handler, componentName attribute.KeyValue,
 	// Don't trace health check requests or favicon browser requests
 	opts = []otelhttp.Option{
 		otelhttp.WithFilter(RequestFilter),
+		// TODO: Use router path as span name, high cardinality should be avoided
```
Aenimus, line 18: Will this be address in another PR?

Aenimus (approved):

pkg/trace/wrap_handler.go
StarpTech: Nope, it's ok for now. It just doesn't follow best practices. Ideally, we would have span names like `/operations/{name*}` or `/webhooks/{name}`.

jensneuse (approved):

PR #1049, opened 2023-06-06
feat: pass internal requests from ORM into GraphQLHandler

fiam:
- When the ORM receives a clientRequest, pass it as part of the query
in __wg.clientRequest
- In GraphQLHandler, decode __wg.clientRequest when serving internal
requests

Fixes ENG-2492, ENG-2507, ENG-2508, ENG-2509

#### Checklist

- [ ] run `make test`


packages/sdk/src/server/plugins/functions.ts
```diff
@@ -89,7 +89,7 @@ const FastifyFunctionsPlugin: FastifyPluginAsync<FastifyFunctionsOptions> = asyn
 							input: (req.body as any)?.input,
 							operations: operationClient,
 							context: requestContext,
-							graph: orm,
+							graph: orm.withRawClientRequest(clientRequest),
```
Pagebakers, line 92: Seems a bit of an odd ball that. we pass the orm from the fastify config instead of initializing here?

fiam: It's because of the conditional initialization with the experimental flag

Pagebakers: Got it!

Pagebakers (changes requested):
I'd add one more test where you use setExtraHeaders on the client and also use withHeaders in a ts operation.

Pagebakers (approved):
Great!

jensneuse (approved):

PR #1057, opened 2023-06-09
feat: initial impl of nats kv

jensneuse:
Not yet complete.


docs-website/config/navigation.tsx
```diff
@@ -611,6 +612,11 @@ const navigation = [
 			},
 		],
 	},
+	{
```
Pagebakers, line 615: This should not get it's own main navigation item, move to `Databases`

Pagebakers (changes requested):

docs-website/config/navigation.tsx
jensneuse: I agree

fiam (approved):
I think it would be great to have an API to register a KV so we can use it internally in the library for storing data (e.g. sessions), but we can add that on top later. LGTM.

PR #1059, opened 2023-06-10
feat: unify authentication mechanisms, redirect back with errors

fiam:
- Unify OIDC and GH authentication moving the common parts into an OAuth2
authentication handler, which also allows us to easily add more mechanism
in the future.

- Handle authentication errors by redirecting to redirect_uri with some
parameters signaling the error added.

- In the frontend, read these new URL parameters and produce a ResponseError
after authentication fails

#### Checklist

- [ ] run `make test`


packages/sdk/src/client/client.ts
```diff
@@ -350,12 +350,45 @@ export class Client {
...
+				let nextUrl: string;
+				if (newQuery) {
+					nextUrl = `${nonQuery}?${newQuery}`;
+				} else {
+					nextUrl = nonQuery;
+				}
+				const code = decodeURIComponent(errorCode || '');
+				const message = decodeURIComponent(errorMessage || errorCode || '');
+				window.history.replaceState({}, window.document.title, nextUrl);
+				throw new ResponseError({ code, message, statusCode: 401 });
```
fiam, line 379: @jensneuse We could use the new `NoUserError` here once #1063 is merged

jensneuse (changes requested):

packages/sdk/src/client/client.ts
```diff
@@ -350,12 +350,45 @@ export class Client {
 		return this.fetchResponseToClientResponse(resp);
 	}
 
+	/**
+	 * Handles authentication errors from a failed attempt in a browser context
+	 * by examining window.location.href
+	 */
+	private handleAuthenticationError() {
```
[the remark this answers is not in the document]
fiam: Sounds good to me, let’s await for @Pagebakers’ feedback

packages/sdk/src/client/client.ts
```diff
@@ -354,12 +354,45 @@ export class Client {
...
+		}
+	}
+
 	/***
 	 * fetchUser makes a GET request to the server to fetch the current user.
 	 * The method throws an error if the request fails to reach the server or
 	 * the server returns a non-200 status code.
 	 */
 	public async fetchUser<U extends User>(options?: FetchUserRequestOptions): Promise<U> {
+		this.handleAuthenticationError();
```
Pagebakers, line 394: Don't think this should be here, should not call this automatically in the client, but let this be handled in the client implementations like @jensneuse suggested. In that case it could perhaps also have an option callback function that can be called.

Pagebakers (changes requested):

Pagebakers (approved):

packages/sdk/src/client/client.ts
fiam: Changed to `NoUserError`

jensneuse (approved):

PR #1060, opened 2023-06-12
fix: produce JSONPatch patches with removals in the right order

fiam:
- Replace github.com/mattbaird/jsonpatch with github.com/wI2L/jsondiff
- Add support for live queries in the TS client
- Make subscriptions cancellable
- Enable jsonpatch for subscriptions over fetch
- Add integration test using jsonpatch over a live query

## Motivation and Context

This caused a problem every time a live query deleted multiple entries at the end
of the list at the same time.

## Marketing Changelog

Live Queries now work with bulk deletions

#### Checklist

- [ ] run `make test`


jensneuse:
Would it not make sense to fork the upstream, provide a fix, and use our fork until it's merged?

fiam:
> Would it not make sense to fork the upstream, provide a fix, and use our fork until it's merged? I think someone mentioned it already: [mattbaird/jsonpatch#30](https://github.com/mattbaird/jsonpatch/issues/30)
> 
> So we can open a PR and help them improve their lib.

As discussed, we've added integration tests and replaced the library with another one.

packages/sdk/src/client/client.ts
```diff
@@ -512,7 +516,18 @@ export class Client {
...
 		while (true) {
-			const { value, done } = await reader.read();
+			let value: Uint8Array | undefined;
+			let done: boolean;
+			try {
+				const result = await reader.read();
+				value = result.value;
+				done = result.done;
+			} catch (e) {
+				if (!(e instanceof Error) || !e.message.indexOf('aborted')) {
```
StarpTech, line 526: Could you add a comment for which case this is required? AbortController?

StarpTech (changes requested):

StarpTech (approved):
LGTM

JivusAyrus (approved):

devsergiy (approved):

PR #1070, opened 2023-06-14
perf: improve config generation/loading performance

### LINKED ISSUE #1020: Bug: Very large schemas cause configuration init issues
### Bug description

I pointed WG at my postgres database (~65 tables, lots of relationships) and it near immediately barfed in a few places with a `RangeError: Invalid string length` error at a few of the JSON.stringify locations in the `configureWunderGraphApplication` code path.

After some googling, it looks like that error is related to v8 running out of memory (?) or otherwise faulting with a too-large input object to stringify.

Taking a look at the db introspection file at generated/introspection/database shows it is 8.2Mb.

I have the code on a forked branch, [diff](https://github.com/wundergraph/wundergraph/compare/main...AesSedai:stringify-fixes?expand=1).

I'm not sure if there are other areas where this is also an issue, but those fixes allowed my postgres schema to be successfully parsed and allowed WG to start up.

### How to reproduce

Make a large schema (50+ tables with lots of relationships) in postgres and try to compose it into Wundergraph.

Private repro available here: https://github.com/AesSedai/wundergraph-stringify-example

### Expected behavior

Wundergraph should compile and run successfully.

### WunderGraph information

N/A

### Environment & setup

- OS: Windows 11 / WSL2 Ubuntu 20
- Database: PostgreSQL
- Node.js version: Node 18

### WunderCtl Version

```
0.154.0
```

fiam:
- Use structuredCopy to copy values when available
- Use protocol buffers as a serialization format
- Use string references for storing database prisma/graphql schemas

- Reduces generation time (on M1) from ~3.8s to ~2.9s in apps/basic => -900ms, -28%
- Reduces config size from ~25MB to ~1.8MB in apps/basic => -23.2MB, -92.8%
- Reduces config loading time (on M1) from 110ms to 7ms => -103ms, -93.7%

Fixes #1020

JSON based configuration can still be produced for debugging/inspection purposes (it won't be used)
by setting the environment variable `WG_GENERATE_CONFIG_JSON=true`

## Motivation and Context

Large schemas are slow to generate/load, so we're improving them bit by bit

## Marketing Changelog

Code generation time reduced by 30%, startup time improved by several seconds with
large schemas.

#### Checklist

- [ ] run `make test`


cli/commands/root.go
```diff
@@ -24,7 +24,7 @@ import (
 )
 
 const (
-	configJsonFilename       = "wundergraph.config.json"
+	serializedConfigFilename = "wundergraph.config.wg"
```
jensneuse, line 27: How about using `.buf` to indicate that the content of the file is a protobuf?

packages/sdk/src/codegen/templates/typescript/helpers.ts
```diff
@@ -21,7 +22,7 @@ const filteredOperations = (application: ResolvedApplication, includeInternal: b
 	includeInternal ? application.Operations : application.Operations.filter((op) => !op.Internal);
 
 export const filterNodeJSOperations = (application: ResolvedApplication): ResolvedApplication => {
-	const copy = JSON.parse(JSON.stringify(application)) as ResolvedApplication;
+	const copy = deepClone(application);
```
jensneuse, line 25: Is it actually necessary to make such a copy?

jensneuse (changes requested):

cli/commands/root.go
fiam: According to one of the creators of PB, it is better to make it application specific because you do need the schema to be able to decode it https://groups.google.com/g/protobuf/c/JWZx9n8CUvw/m/bwvQWPL70Y8J - No strong feelings, though, if you'd prefer `.buf` or `.pb` (or anything else) that's fine by me.

packages/sdk/src/codegen/templates/typescript/helpers.ts
fiam: It didn't seem necessary to me, but I thought I might be missing something. I'll take a look and I'll see if we can get rid of the copy.

packages/sdk/src/configure/index.ts
```diff
@@ -1259,11 +1261,16 @@ const ResolvedWunderGraphConfigToJSON = (config: ResolvedWunderGraphConfig): str
...
 		configHash: configurationHash(config),
 	};
+	return out;
+};
 
-	return JSON.stringify(out, null, 2);
+const storeString = (stringStorage: Record<string, string>, s: string) => {
+	const key = crypto.createHash('sha1').update(s).digest('hex');
+	stringStorage[key] = s;
+	return `string://${key}`;
```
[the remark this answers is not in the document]
fiam: Good idea, I'll change it.

cli/commands/root.go
jensneuse: That makes sense. If we follow this rule, we could name it e.g. `config.wundergraphconfig`, or `file.wgconfig`, because we will also need to find a different name for operations, like `OperationName.wgoperation`. Any good ideas?

pkg/engineconfigloader/engineconfigloader.go
```diff
@@ -479,11 +492,19 @@ func (l *EngineConfigLoader) Load(engineConfig wgpb.EngineConfiguration, wgServe
 			if in.CustomDatabase == nil {
 				continue
 			}
+			prismaSchema, err := l.LoadStoredString(engineConfig, in.CustomDatabase.PrismaSchema)
+			if err != nil {
+				return nil, fmt.Errorf("could not load prisma schema for data source %s: %w", in.Id, err)
+			}
+			graphqlSchema, err := l.LoadStoredString(engineConfig, in.CustomDatabase.GraphqlSchema)
```
devsergiy, line 499: are we using this only for prisma? could we do the same for the graphql datasource for the upstream schema?

packages/sdk/src/codegen/templates/typescript/helpers.ts
fiam: It is necessary because the idea behind that function is to obtain a `ResolvedApplication` that includes only the TS operations and can be passed to https://github.com/wundergraph/wundergraph/blob/b7a0fc604d54f1711ae68ace29242eab46a2b668/packages/sdk/src/codegen/templates/typescript/helpers.ts#L30 (which filters based on operation type)

cli/commands/root.go
fiam: I've renamed it

pkg/engineconfigloader/engineconfigloader.go
fiam: Your wish is my command, PTAL

packages/sdk/src/configure/index.ts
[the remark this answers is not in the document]
fiam: I've added an `InternedString` that stores the references, PTAL

devsergiy (approved):

jensneuse (approved):

PR #1073, opened 2023-06-17
fix: federation introspection

devsergiy:
## Marketing Changelog

1. Federation fixes:
- allow to use an entity from another subgraph as a field type
- fix creating subgraph entities union
- fix merging descriptions - uses first version of description

2. Fix namespacing of schema directives

#### Checklist

- [ ] run `make test`


jensneuse (approved):

packages/sdk/src/definition/merge.ts
```diff
@@ -470,20 +487,97 @@ Without authentication, the operation will return an Unauthorized error with sta
...
+	value: string;
+}
+type DescriptionsMap = Map<string, DescriptionInfo>;
+
+const collectAndReplaceDescriptions = (descriptions: DescriptionsMap, schemaSDL: string): GraphQLSchema => {
+	const schemaAST = parse(schemaSDL);
+
+	const modifiedAst = visit(schemaAST, {
+		leave(node, key, parent, path, ancestors) {
+			if (!('description' in node && !!node.description && node.description.value.trim() !== '')) {
```
Aenimus, line 501: I really don't like how this condition reads

Aenimus: Can it not be simply `node.description?.value.trim() === ''`

devsergiy: is it better like this?

```
			const nodeHasDescription = 'description' in node && !!node.description && node.description.value.trim() !== '';
			if (!nodeHasDescription) {
				return node;
			}
```

I need a guard that description is present, otherwise, I should repeat `description?`
And don't like empty string to bool conversion

packages/sdk/src/definition/merge.ts
```diff
@@ -470,20 +487,102 @@ Without authentication, the operation will return an Unauthorized error with sta
...
+	value: string;
+}
+type DescriptionsMap = Map<string, DescriptionInfo>;
+
+const collectAndReplaceDescriptions = (descriptions: DescriptionsMap, schemaSDL: string): GraphQLSchema => {
+	const schemaAST = parse(schemaSDL);
+
+	const modifiedAst = visit(schemaAST, {
+		leave(node, key, parent, path, ancestors) {
+			const nodeHasDescription = 'description' in node && !!node.description && node.description.value.trim() !== '';
```
Aenimus, line 501: I like this much better, but you can still do `!node.description?.value.trim.()` (or ``node.description?.value.trim.() !== ''` if you want to be verbose) instead of `!!node.description &&`.

devsergiy: image

With your approach, I will have to access the description everywhere further like this `node.description?.block` and `node.description?.value`

It will work, but it is semantically wrong, cause from the beginning of the function I will know if the description is present or not. If not I return, If it exists additional checks for existence do not have any sense, cause we already know that it is exists

Aenimus: Makes sense!

Aenimus (approved):

PR #1086, opened 2023-06-21
fix: delete unknown directives from the schema generated from OAS

### LINKED ISSUE #1083: New problems with version sdk= 0.166.0 and wunderctl= 0.160.0
### Bug description

I have a proyect that was running any problems with version sdk=0.164.1, now i have upgraded to version sdk=0.166.0 and when i run my project it generate follow error:
    error: {
      "type": "Error",
      "message": "Unknown directive "@regexp".",
      "stack":
          Error: Unknown directive "@regexp".
              at assertValidSDL (...\wundergraph\proy-inicial\node_modules\graphql\validation\validate.js:135:11)
......

i have an OpenAPIv2 instropection that it run before any problem and now Not. 

By the error i presume that the problem is with a regular expression in swagger file for OpenApi introspection and by this i removed the follow line that refer the pattern and i save of the swagger file and newly the program it running ok again.

   "schemas": {
      "DATE": {
        "type": "string",
        "format": "date-time",
        "pattern": "^\d{4}-[01]\d-[0123]\dT[012]\d:[0-5]\d:[0-5]\d(.\d+)?(Z|([-+][012]\d:[0-5]\d))$"
      },
      "NUMBER": {
        "type": "number"
      },
      "VARCHAR2": {
        "type": "string"
      }
    }
 
this problem does not exist with the sdk version 0.164.1, Is This a bug as of version sdk 0.165.0 i tested this version and also it has the bug!!

### Expected behavior

_No response_

### Environment & setup

- OS: 
- Go version: 
- Database: 
- Node.js version: 

### WunderCtl Version

```

```

fiam:
Remove instances of the @regexp directive

Fixes #1083

#### Checklist

- [ ] run `make test`


thisisnithin (approved):

packages/sdk/src/graphql/schema.ts
```diff
@@ -179,6 +179,7 @@ export const cleanupSchema = (schema: GraphQLSchema): string => {
 				case 'example':
 				case 'globalOptions':
 				case 'httpOperation':
+				case 'regexp':
```
devsergiy, line 182: please do the same on line 200

it should came in pair

devsergiy (changes requested):

packages/sdk/src/graphql/schema.ts
devsergiy: overall we could replace switch with if else to not repeat it twice

```javascript
if openApiDirectives.includes(value) {
} else id federationDirectives.includes(value) {
}
```

I could do the change in case you don't have time

devsergiy (approved):

packages/sdk/src/graphql/schema.ts
```diff
@@ -167,41 +181,17 @@ export const cleanupSchema = (schema: GraphQLSchema): string => {
...
-				case 'resolveRoot':
-				case 'typescript':
-				// omnigraph/soap directives:
-				case 'soap':
-					return null;
-				default:
-					return node;
+			if (federationDirectives.includes(node.name.value) || omnigraphDirectives.includes(node.name.value)) {
+				return null;
+			} else {
```
fiam, line 186: We can drop the whole `else` here, if there's no returned value (i.e. `undefined`) the node is also kept.

PR #1099, opened 2023-07-03
feat: reimplement customJSONScalars

Aenimus:
## Motivation and Context

customJSONScalars was deprecated because it was no longer necessary to include the typenames for replacement scalars.
However, this also removed the ability for non-replaced scalars to be recognised as JSON and they would default to strings.
This functionality has been added back but now also for inputs.

#### Checklist

- [ ] run `make test`

### DIFF
--- a/.github/workflows/packages-ci.yaml
+++ b/.github/workflows/packages-ci.yaml
@@ -81,8 +81,8 @@ jobs:
       - name: Run testsuite
         run: pnpm test:suite:ci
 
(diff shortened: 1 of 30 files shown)


Pagebakers (changes requested):
LGTM but shouldn't this be documented also?

Pagebakers (approved):

fiam (approved):
LGTM

PR #1110, opened 2023-07-06
fix: customJSONScalars are ignored when introspection is fetched from cache

fiam:
Namespacing was applied twice, since it was done in the Api constructor. Move
it to the GraphQL instrospection before the GraphQLApi result is constructed.

#### Checklist

- [ ] run `make test`


packages/sdk/src/definition/index.ts
```diff
@@ -100,7 +100,7 @@ export class Api<T = ApiType> implements RenameTypes, RenameTypeFields {
 		this.Fields = fields;
 		this.Types = types;
 		this.interpolateVariableDefinitionAsJSON = interpolateVariableDefinitionAsJSON;
-		this.CustomJsonScalars = applyNameSpaceToCustomJsonScalars(namespace, customJsonScalars);
+		this.CustomJsonScalars = customJsonScalars ?? new Set();
```
Aenimus, line 103: Please could we add typing here? new Set<string>()

fiam: Sure! I've just pushed it, PTAL

ggeijteman-wg (approved):

Aenimus (approved):

PR #1121, opened 2023-07-11
feat: add support for openai

fiam:
## Motivation and Context

This adds support for OpenAI to TypeScript Operations.

## Marketing Changelog

This adds support for OpenAI To WunderGraph, allowing developers to build applications on top of WunderGraph using OpenAI. Adding AI capabilities to an existing comes with a few challenges.

1. How do I securely integrate OpenAI into my existing API landscape?
2. How do I validate user inputs before using them with OpenAI?
3. How can I allow OpenAI to interact with my existing APIs, like REST, SOAP, GraphQL, Databases and more?
4. How do I get responses from OpenAI in a structured format so that they can be used by other applications, clients, etc...?
5. How can I build OpenAI Agents to easily leverage my existing APIs and services?
6. How can I chain OpenAI Agents so they can build on top of each other?
7. How can I batch operations that would otherwise exceed the token limit of a model?

With the initial release of WunderGraph's OpenAI support, we're solving all of these problems.
We give you an easy way to integrate OpenAI into your existing APIs, allowing you to write Agents that can interact with your existing APIs and Databases in just a few lines of code. You can easily chain them, combine them, and most importantly, validate user inputs before feeding them into OpenAI. Finally, you're able to batch operations, making sure that you're not exceeding token limits while being able to easily merge batched results. Last but not least, we've made it super easy to use zod as the language to define structured outputs from your OpenAI Agents. This way, you can easily return structured data from OpenAI and we're automatically generating an OpenAPI specification for your AI-driven apps.


examples/apollo-federation/.wundergraph/operations/TopProducts.graphql
```diff
@@ -1,5 +1,5 @@
 query {
-	topProducts: federated_topProducts(random: true) {
+	topProducts: topProducts(random: true) {
```
Pagebakers, line 2: topProducts: topProducts is a bit redundant :D

packages/sdk/src/openai/index.ts
```diff
@@ -0,0 +1,330 @@
...
+						],
+					});
+					const structuredResponse = JSON.parse(completions.data.choices[0].message!.function_call!.arguments!);
+					this.outputZodSchema.parse(structuredResponse);
+					return {
+						structuredOutput: structuredResponse as any,
+					} as any;
+				}
+			}
+		} catch (e) {
```
Pagebakers, line 308: e: any

packages/sdk/src/openai/index.ts
```diff
@@ -0,0 +1,330 @@
...
+								name: 'out',
+								description: 'This is the function that returns the result of the agent',
+								parameters: this.outputJsonSchema,
+							},
+						],
+					});
+					const structuredResponse = JSON.parse(completions.data.choices[0].message!.function_call!.arguments!);
+					this.outputZodSchema.parse(structuredResponse);
+					return {
+						structuredOutput: structuredResponse as any,
```
Pagebakers, line 304: as any isn't required here I believe?

packages/sdk/src/server/plugins/functions.ts
```diff
@@ -82,15 +87,27 @@ const FastifyFunctionsPlugin: FastifyPluginAsync<FastifyFunctionsOptions> = asyn
 							tracer: fastify.tracer,
 							traceContext: req.telemetry?.context,
 						});
-						const ctx: HandlerContext<any, any, any, any, any, any, any> = {
-							log: createLogger(fastify.log),
+						const configuration = new Configuration({
```
Pagebakers, line 90: Should we use `experimental` feature flag for this, like the ORM? How does the feature affect cold starts?

Pagebakers (changes requested):

examples/apollo-federation/.wundergraph/wundergraph.config.ts
```diff
@@ -4,19 +4,18 @@ import server from './wundergraph.server';
 import operations from './wundergraph.operations';
 
 const federatedApi = introspect.federation({
-	apiNamespace: 'federated',
 	upstreams: [
 		{
-			url: 'http://localhost:4001/graphql',
+			url: 'https://wg-federation-demo-accounts.fly.dev/graphql',
```
fiam, line 9: Do we want to change these to point to the live copies or was it accidental?

packages/sdk/src/configure/index.ts
```diff
@@ -1637,13 +1639,16 @@ const loadAndApplyNodeJsOperationOverrides = async (
 // this function.
 const applyNodeJsOperationOverrides = (
 	operation: TypeScriptOperation,
-	overrides: NodeJSOperation<any, any, any, any, any, any, any, any, any, any>
+	overrides: NodeJSOperation<any, any, any, any, any, any, any, any, any, any, any>
```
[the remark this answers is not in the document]
fiam: I'd agree, but I'm not sure how that would look. If you can make it more manageable, that would be great.

packages/sdk/src/graphql/operations.ts
```diff
@@ -437,6 +442,15 @@ const jsonSchemaUpdater = (arg: ConstArgumentNode) => {
 	};
 };
 
+const loadOperationDescription = (operationFile: GraphQLOperationFile, wgDirAbs?: string): string => {
```
fiam, line 445: Could we do this on the Go side when we call `wunderctl loadoperations`? I think that would be significantly faster and require less code.

fiam (commented):
OpenAI integration looks amazing! Just a couple of comments regarding the integration with the rest of the SDK

packages/sdk/src/server/plugins/functions.ts
jensneuse: I think it's fine without experimental. We're still at 0.x
Regarding cold starts, it's just a small rest api client for the OpenAI API.

jensneuse (approved):

packages/sdk/src/graphql/operations.ts
fiam: Sounds good, no reason to rewrite it now!

PR #1139, opened 2023-07-19
feat: unify pretty logs

fiam:
- Make the Node side always emit JSON logs, prettify them in the node
- Make wunderctl start not use pretty logging by default
- Cleanup code for passing around whether logs should be prettified
- Use a more human-readable format for pretty logs
- Unify colors for levels across both sides of the SDK

Fixes ENG-3334

## Motivation and Context

More friendly logs improve the developer experience by making it easier to debug your WG application

## Marketing Changelog

Pretty logging is now even prettier with a more human readable format

#### Checklist

- [ ] run `make test`


cli/commands/root.go
```diff
@@ -139,6 +141,44 @@ func sendTelemetry(cmd *cobra.Command) error {
 	return TelemetryClient.Send(metrics)
 }
 
+func workaroundBugWithDuplicatedFlags(cmd *cobra.Command) {
```
StarpTech, line 144: Could you explain again why you need this?

packages/sdk/src/logger/index.ts
```diff
@@ -55,23 +53,6 @@ const initLogger = (destination: DestinationStream): pino.Logger => {
 		level: logLevel,
 	};
 
-	if (enablePretty) {
```
StarpTech, line 58: That means wundergraph server prettify their logs through wundergraph? That implicit that the server can't run standalone in pretty log mode. I'm ok with it.

pkg/logging/pretty_encoder.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+func (e *prettyEncoder) EncodeEntry(entry zapcore.Entry, fields []zapcore.Field) (*buffer.Buffer, error) {
+	buffer, err := e.Encoder.EncodeEntry(entry, fields)
+	if err != nil {
+		return nil, err
+	}
+	pretty, err := e.prettifier.PrettifyJSON(bytes.NewReader(buffer.Bytes()))
+	if err != nil {
+		// If prettifying fails, return the raw log line
```
StarpTech, line 37: Please add tests here as well

pkg/scriptrunner/scriptrunner.go
```diff
@@ -227,26 +284,17 @@ func newCmd(options CmdOptions) (*gocmd.Cmd, chan struct{}) {
 		for cmd.Stdout != nil || cmd.Stderr != nil {
 			select {
 			case line, open := <-cmd.Stdout:
+				fmt.Fprintln(options.stdout, line)
```
StarpTech, line 287: Did you consider in your change that the script runner is a generic module that is used to execute any kind of command?

StarpTech (changes requested):

cli/commands/root.go
fiam: It's because the `pretty-logging` flag has different values for the global flag (defaults to `false`) vs the `wunderctl up` flag (defaults to `true`). This is not really supported by cobra/viper so we have do to some fixups to make it work.

packages/sdk/src/logger/index.ts
fiam: That's right. I went for this design because it allows us to have a single prettifier implementation and it also reduces startup time for the server since we no longer import `pino-pretty`. The only downside is that the server on its own can't prettify logs, but I thought that was a reasonable tradeoff.

Pagebakers (approved):
A small step for man, a huge leap for mankind

cli/commands/root.go
StarpTech: ok, I just find it hard to understand why it was necessary.

packages/sdk/src/logger/index.ts
StarpTech: ok

pkg/logging/prettifier.go
```diff
@@ -0,0 +1,231 @@
...
+		componentKey:  componentKey,
+		callerKey:     callerKey,
+		timeKey:       timeKey,
+		stacktraceKey: stacktraceKey,
+		ignoredKeys:   ignoredKeys,
+	}
+}
+
+func (p *Prettifier) PrettifyJSON(r io.Reader) (*buffer.Buffer, error) {
+	o := orderedmap.New()
```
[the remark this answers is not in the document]
fiam: I've added tests for the prettifier

pkg/logging/pretty_encoder.go
fiam: Do you mean tests for `prettyEncoder`? Most of the functionality is in `Prettifier` while `prettyEncoder` is a very small Encoder for zap, so I'm not sure how we should test it. Could you clarify what you have in mind? Thanks!

pkg/scriptrunner/scriptrunner.go
fiam: I did. By default, it will print to stdout/stderr, but the new API allows you to redirect its outputs (e.g. to discard them for the TUI) as well as to set a pipe function that rewrites each line before it gets outputted.

pkg/logging/pretty_encoder.go
StarpTech: Fine for me. It was important to have some coverage.

StarpTech (approved):
LGTM

PR #1141, opened 2023-07-21
fix: allow setting logging levels from command line

fiam:
- Make the command line flag propagate correctly to the TS side
- Correctly update the logging level when the config is parsed
- Introduce a TRACE level for the TS side (zap doesn't support it)

Fixes ENG-3335
Fixes ENG-3357
Fixes ENG-900

## Motivation and Context

Controlling the logging level from the command line allows momentarily switching the logging level without having to alter the configuration.

#### Checklist

- [ ] run `make test`


bitcoinfinancier (approved):

jensneuse (approved):

Pagebakers (changes requested):
We need to update the docs to reflect the changed options.

Pagebakers (approved):

PR #1160, opened 2023-07-31
fix: allow svelte 3 | 4 peer-dependency in @wundergraph/svelte-query …

leoalves:
Svelte released v4.  Updated `@wundergraph/svelte-query`  peer-dependency `package.json` to allow either version.


Pagebakers (approved):
Nice!

Aenimus (approved):

PR #1194, opened 2023-08-14
fix: jsonschema template

thisisnithin:
declare module instead of type in jsonschema.ts

Also Unless we set shamefully hoist, the module is not detected with pnpm. So we just ts-ignore it in that case. Other package managers have no issue. This is done to bypass pnpm with testapps.

#### Checklist

- [x] run `make test`


StarpTech (approved):
LGTM

Aenimus (approved):

PR #1210, opened 2023-08-18
fix: use correct import on react hooks template for server imports

RayHughes:
The `@wundergraph/sdk` import does not resolve ` RequestOptions`, `MutateRequestOptions`, `SubscriptionRequestOptions`, and `Response`  functions. These functions can be found at `@wundergraph/sdk/server`

## Motivation and Context

Working on generating client and realized that the imports could not be resolved on the template.

## Marketing Changelog
- Corrected import on React Hooks template.

#### Checklist

- [ ] run `make test`


fiam:
cc @Pagebakers

Pagebakers:
Thanks for the PR @RayHughes, we don't recommend using this code any longer, it will be removed in future releases. And recommend using the SWR or React Query client (separate packages) instead.

It'd be better to also not import from /server in here or at least use `import type {}` to make sure only types get imported here and no other server code.

Pagebakers:
Definitely, please follow these instructions (apply to expo, but also regular react native)

https://docs.wundergraph.com/docs/getting-started/expo-quickstart

RayHughes:
> Thanks for the PR @RayHughes, we don't recommend using this code any longer, it will be removed in future releases. And recommend using the SWR or React Query client (separate packages) instead.
> 
> It'd be better to also not import from /server in here or at least use `import type {}` to make sure only types get imported here and no other server code.

Pagebreakers, 

WG/SWR or the actual SWR library. We're having trouble getting our generated client to work in RN

PR #1234, opened 2023-08-29
feat: add support for server ping during subscription idle times

fiam:
Some provides or middleboxes aggressively close TCP connections that
look inactive. This means a long running subscription might be end
disconnected if no new data arrives for some time. To counteract this,
introduce a ping mechanism in the server that ensures that the line is
not idle for longer than the requested interval by sending an empty
payload if no actual data has been sent.

Also, refactor all the SSE / jsonpatch support to implement it in a single
place.

Additionally, fix a bug which incorrectly shared a variable in the client
when returning subscription results to the caller.

#### Checklist

- [ ] run `make test`


Pagebakers (approved):

packages/sdk/src/configure/options.ts
```diff
@@ -55,6 +56,9 @@ const DefaultNodeOptions = {
 	},
 	defaultRequestTimeoutSeconds: 0,
 	defaultHttpProxyUrl: new EnvironmentVariable(WgEnv.HttpProxyUrl, ''),
+	subscriptions: {
+		serverPingIntervalMs: new EnvironmentVariable(WgEnv.SubscriptionServerPingIntervalMs, 0),
```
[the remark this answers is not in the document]
fiam: I was a bit wary of introducing extra data in the line because it's an incompatible change to the protocol. On the other hand, if we don't have pings by default it seems likely third party clients will not realize they have to support them, I guess your comment changed my mind. What do you think would be a good default? 10s?

docs-website/src/pages/docs/architecture/wundergraph-conventions.md
```diff
@@ -88,20 +88,21 @@ So, you're safe to use `process.env.{VARIABLE_NAME}` in your `wundergraph.server
...
+| `WG_NODE_PORT`                         | The port of the `WunderNode`.                                        | `9991`                  |
+| `WG_NODE_INTERNAL_PORT`                | The port of the `WunderNode` for internal operations.                | `9993`                  |
+| `WG_SERVER_URL`                        | The URL of the `WunderGraph Server`.                                 | `http://localhost:9992` |
+| `WG_SERVER_HOST`                       | The host of the `WunderGraph Server`.                                | `localhost`             |
+| `WG_SERVER_PORT`                       | The port of the `WunderGraph Server`.                                | `9992`                  |
+| `WG_PROMETHEUS_ENABLED`                | Whether to enable Prometheus metrics. \*Enterprise license required. | `false`                 |
+| `WG_PROMETHEUS_PORT`                   | Port used to serve Prometheus metrics.                               | `8881`                  |
+| `WG_SUBSCRIPTION_SERVER_PING_INTERVAL` | Ping interval when serving subscriptions, as a duration (e.g. `30s`) | `off`                   |
```
Pagebakers, line 105: We have a default now no?

Pagebakers (changes requested):

docs-website/src/pages/docs/architecture/wundergraph-conventions.md
fiam: I've set it to `30s`, PTAL

Pagebakers (approved):

Aenimus (approved):

PR #1247, opened 2023-09-04
fix: string email_verified fields from sign in with apple

rwest202:
Neither token nor cookie based providers work with sign in with apple via Auth0. The reason is that apple provides the email_verified field as a string. This breaks the parser which expects a boolean value. I fixed the token based flow here.

#### Checklist

- [x] run `make test`


fiam (approved):
I like this workaround, LGTM

Pagebakers (approved):

PR #1248, opened 2023-09-04
feat: propagate error messages from functions/hooks to API responses

fiam:
When a function or a hook throws an error, propagate it through the stack
until it's output by the API as an error. This allows clients to retrieve
the very same error message that was produced in the backend.

If the error is not one of the types we control, just log it to the terminal's
output without returning it through the API, since that could expose internal
implementation details.

#### Checklist

- [ ] run `make test`

### DIFF
--- a/packages/sdk/src/client/errors.ts
+++ b/packages/sdk/src/client/errors.ts
@@ -66,9 +66,40 @@ export interface GraphQLError {
 	path?: ReadonlyArray<string | number>;
 }
 
-export type OperationErrorBaseFields = {
+export type OperationErrorFields = {
 	message?: string;
-	code: string;
+	code?: string;
+	statusCode?: number;
+	stack?: string;
+	cause?: OperationErrorFields;
+};
+
+type ErrorWithCause = {
+	cause?: Error;
+};
+
+type MightBeOperationError = {
+	code?: string;
+	statusCode?: number;
+};
+
+const errorToJSON = (err: Error): OperationErrorFields => {
+	const result: OperationErrorFields = {
+		message: err.message,
+		stack: err.stack,
+	};
+	const opError = err as MightBeOperationError;
+	if (opError.code) {
+		result.code = opError.code;
+	}
+	if (opError.statusCode) {
+		result.statusCode = opError.statusCode;
+	}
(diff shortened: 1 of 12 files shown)


packages/sdk/src/client/errors.ts
```diff
@@ -66,9 +66,40 @@ export interface GraphQLError {
 	path?: ReadonlyArray<string | number>;
 }
 
-export type OperationErrorBaseFields = {
+export type OperationErrorFields = {
 	message?: string;
-	code: string;
+	code?: string;
```
Pagebakers, line 71: What's the reason for making this optional now?

Pagebakers: Shouldn't we implement codes for all our errors? This makes it much easier to handle client side error reporting.

packages/testsuite/apps/hooks/.wundergraph/wundergraph.server.ts
```diff
@@ -1,20 +1,20 @@
...
-					const result = await hook.internalClient.queries.Infinite({
-						input: hook.input,
-					});
-
-					const result2 = await hook.operations.query({
+					const result = await hook.operations.query({
 						operationName: 'Infinite',
 						input: hook.input,
 					});
+					if (result.error) {
```
Pagebakers, line 15: This test should depend on the internals to throw an error right when the dept limit is reached? should this still be here?

Pagebakers (changes requested):
Big improvement! left some comments

packages/testsuite/apps/hooks/.wundergraph/wundergraph.server.ts
fiam: If we don't throw, the hook doesn't fail. This worked because an error in the hooks produced an error message (which wasn't valid JSON) and caused the whole operation to fail. Now that we're extracting the error messages and returning them as JSON, the error is returned to the `preResolve` hook as an operation error; hence if we don't throw it, it gets swallowed.

Pagebakers (approved):

JivusAyrus (approved):

Aenimus (approved):

PR #1281, opened 2023-09-21
docs: add comment about introspect.OpenApi being phased out

fiam:
#### Checklist

- [ ] run `make test`

### DIFF
--- a/packages/sdk/src/definition/index.ts
+++ b/packages/sdk/src/definition/index.ts
@@ -496,6 +496,12 @@ export const introspect = {
 	mongodb: introspectMongoDB,
 	prisma: introspectPrisma,
 	federation: introspectFederation,
+	/**
+	 * @remarks introspect.openApi exists only for backwards compatibility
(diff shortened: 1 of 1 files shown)


Aenimus (approved):

packages/sdk/src/definition/index.ts
```diff
@@ -496,6 +496,12 @@ export const introspect = {
 	mongodb: introspectMongoDB,
 	prisma: introspectPrisma,
 	federation: introspectFederation,
+	/**
+	 * @remarks introspect.openApi exists only for backwards compatibility
```
Pagebakers, line 500: Not @deprecated?

fiam: I think @deprecated is not the right choice here, because we eventually want to make it an alias to `openApiV2` without removing it.

Pagebakers: Hm ok, at that point we can deprecate it i suppose, no need to maintain 2 implementations.

Pagebakers (approved):

PR #1285, opened 2023-09-23
fix: propagate roles to user for oidc

### LINKED ISSUE #1084: Consider populating User.roles from token
I do not know that problem exist when keycloak jwt is parsed in wundergraph....The array roles is not populate..role attribute comes with null value....For fix this situation i use this hook:
```
authentication: {
     mutatingPostAuthentication: async ({ user }) => {
              console.log("===Autenticacion===")
              console.log(user.roles)
              if  (user.roles === null) {
                if (user.accessToken) {
                    console.log(user.accessToken.roles)
                    user.roles = Object.values(user.accessToken.roles)

                }
              }
              console.log(user.roles)

              return {
                user,
                status: 'ok',
              };
            },
          },
```
With this senteces i populate the property roles in user object:
user.roles = Object.values(user.accessToken.roles)
I will want that wundergraph teams will review because the property roles is not populate if in jwt in delivered for keycloak comes with attribute roles like as arrays of strings?

I suggest using the same pattern used by Hasura configuring special claims like "x-hasura-allowed-roles" in wundergraph could be "x-wundergraph-allowed-roles". https://hasura.io/docs/latest/auth/authentication/jwt/

ShaneMaglangit:
## Motivation and Context

Fixes #1084 . Addresses an issue where user roles were not being populated correctly in Keycloak

It looks like we do get the Roles, but we missed the step of extracing and propagating it to the User instance/object.

#### Checklist

- [x] run `make test`


ShaneMaglangit:
Hi @Pagebakers, would appreciate if you can have a look at this changes. Also, would appreciate if you can clarify where/how should we handle unsupported roles. Is there a particular layer/package/file where we should check for this? Do we filter the unsupported roles out or explicitly throw an error?

ShaneMaglangit:
Hey @Pagebakers,

I'd greatly appreciate it if you could take a look at these changes. Can you verify if we still need to check for authorization config roles. It seems like our primary concern is to propagate the roles, and I assume we might already have a mechanism in place to handle unsupported roles (as part of the RBAC implementation/feature)

If we do need to address unsupported roles, what would be your recommendation? Should we filter them out or throw an explicit error? I'm also curious about which specific layer, package, or file we should target for implementing this check.

re: https://github.com/wundergraph/wundergraph/issues/1084#issuecomment-1729787693

ShaneMaglangit:
Closing in favor of https://github.com/wundergraph/wundergraph/pull/1288

PR #1292, opened 2023-09-27
Update post-logout-hook.md for auth0 signout details.

jimmiebtlr:
## Motivation and Context

I found it a bit difficult to understand how to properly logout auth0 accounts.  There didn't seem to be much in the way of details to how to achieve this.

## Marketing Changelog

Improve auth0 docs on properly logging out users.

I'm not sure if there are other places in the docs I should add this details.  I'm also not sure if this is the best approach or if there's another way to handle it.  We're waiting on the iframe to logout before calling wundergraph logout to be sure we properly logout before getting redirected.

### DIFF
--- a/docs-website/src/pages/docs/auth/cookie-based-auth/auth0/index.md
+++ b/docs-website/src/pages/docs/auth/cookie-based-auth/auth0/index.md
@@ -91,8 +91,25 @@ import { useAuth } from 'components/generated/nextjs';
 export default function Page() {
   const { logout } = useAuth();
 
-  return <button onClick={() => logout({ logoutOpenidConnectProvider: true })}>Logout</button>;
+  return <button onClick={() => logout()}>Logout</button>;
 }
+```
+
+Unfortunately, Auth0 also doesn't provide a way to call a back channel API to log out the user.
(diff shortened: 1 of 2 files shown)


jimmiebtlr:
A little worried the latest approach is a bit racey, not sure if the onLoad event will wait for the redirect to load, or just fire as soon as a response from the auth0 url.

fiam:
Hello @jimmiebtlr - Thanks for your contribution. To sign out users using auth0 you have to set `logoutOpenidConnectProvider: true` and declare your OIDC provider using `authProviders.auth0()` like in the [auth0 example](https://github.com/wundergraph/wundergraph/blob/475591d4f59ea311b30092566e6fee65d86dae39/examples/auth0-oidc-authentication/.wundergraph/wundergraph.config.ts#L39)

jimmiebtlr:
I don't believe that actually works, also in this PR you'll notice changes to some docs that confirm that.

jimmiebtlr:
I don't think my current approach in the PR is ideal, but `logoutOpenidConnectProvider: true` doesn't impact the auth0 login state.

PR #1294, opened 2023-09-28
perf: make wunderctl generate faster, add a few more trace log messages

fiam:
- Avoid using objectHash() on the whole application object. It gets
slow when there are many operations. Use an md5 of its JSON representation
which should be good enough for our purposes (generating a different URL
when the config caches)
- Avoid generating the same templates twice, use an in-memory cache during
generation
- Remove a few unneccesary copies performed during code generation
- Avoid double calculations

Reduces execution time of wunderctl generate with a test application by more than 50%

Fixes ENG-4045

#### Checklist

- [ ] run `make test`


Aenimus (approved):

packages/nextjs/src/template.ts
```diff
@@ -13,33 +13,27 @@ import {
...
 		const tmpl = Handlebars.compile(handlebarTemplate);
-		const content = tmpl({
+		const imports = modelImports(config.application, false, true);
+		const queriesWithInput = config.application.Operations.filter(filterOperation(OperationType.QUERY, true));
+		const queriesWithoutInput = config.application.Operations.filter(filterOperation(OperationType.QUERY, false));
+		const args = {
 			baseURL: config.deployment.environment.baseUrl,
 			sdkVersion: config.sdkVersion,
-			applicationHash: hash(config).substring(0, 8),
+			applicationHash: configurationHash(config),
```
StarpTech, line 30: This is a dangerous change. Object-hash respects all kinds of data types. Are you sure we cover all cases with just a hash?

StarpTech (changes requested):

StarpTech (approved):
LGTM

packages/nextjs/src/template.ts
StarpTech: Clarified internally

PR #1377, opened 2024-03-27
docs: document new nuxt `import.meta.*` properties

danielroe:
## Motivation and Context

This is a very early PR to make these docs compatible with [changes we expect to release in Nuxt v5](https://github.com/nuxt/nuxt/issues/25323).

In [Nuxt v3.7.0](https://github.com/nuxt/nuxt/releases/tag/v3.7.0) we added support for `import.meta.*` (see [original PR](https://github.com/nuxt/nuxt/pull/22428)) and we've been gradually updating docs and moving across from the old `process.*` patterned variables.

These variables are replaced at build-time and enable tree-shaking in bundled code.
This change affects _runtime_ code (that is, that is processed by the Nuxt bundler, like vite or webpack) rather than code running in Node. So it really doesn't matter what the string is, but it makes more sense in an ESM-world to use `import.meta` rather than `process`.

#### Checklist

- [x] run `make test`

### DIFF
--- a/docs-website/src/pages/docs/examples/nuxt.md
+++ b/docs-website/src/pages/docs/examples/nuxt.md
@@ -70,13 +70,13 @@ export default defineNuxtPlugin((nuxt) => {
 
   nuxt.vueApp.use(VueQueryPlugin, options);
 
-  if (process.server) {
+  if (import.meta.server) {
     nuxt.hooks.hook('app:rendered', () => {
       vueQueryState.value = dehydrate(queryClient);
     });
   }
 
-  if (process.client) {
+  if (import.meta.client) {
     nuxt.hooks.hook('app:created', () => {
       hydrate(queryClient, vueQueryState.value);
     });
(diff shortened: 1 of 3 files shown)


danielroe:
Not stale.

danielroe:
Not stale

danielroe:
Not stale

danielroe:
Not stale

danielroe:
Not stale. I would appreciate a review. 🙏

danielroe:
@StarpTech would you mind having a look?

danielroe:
not stale

danielroe:
😢

danielroe:
🙏

danielroe:
I’d appreciate a review if possible 🙏

this corrects nuxt documentation and should be totally safe

StarpTech:
Hi @danielroe, sry for the late response and thanks for the PR. This project is in maintenance mode.

PR #1383, opened 2024-04-25
fix: openapi edge case for empty payloads

acdn-tsmith:
## Motivation and Context
The generated `wundergraph.openapi.json` specification doesn't currently handle mutations that have no input. The [requestBody](https://spec.openapis.org/oas/v3.1.0#request-body-object) shouldn't be required per the OpenAPI 3.1.0 docs, but the `openapibuilder` code currently assumes that we always have an input.

Now, for mutations that have no input, instead of of a payload that shows as an empty object...
Screenshot 2024-04-25 at 12 35 19 PM
We have a payload that just shows as empty and non-required:
Screenshot 2024-04-25 at 12 35 25 PM

This will correct the generated spec for the small subset mutation operations that don't consume an input.

## Marketing Changelog
OpenAPI Generated Spec now formats POST's with empty payloads correctly

#### Checklist

- [x ] run `make test`
- [x ] PR title must follow [conventional-commit-standard](https://github.com/wundergraph/wundergraph/blob/main/CONTRIBUTING.md#conventional-commit-standard)
- [ x] tests and/or benchmarks are included
- [ x] commit message and code follows the [Code of conduct](https://github.com/wundergraph/wundergraph/blob/main/CODE_OF_CONDUCT.md)

### DIFF
--- a/packages/sdk/src/openapibuilder/index.test.ts
+++ b/packages/sdk/src/openapibuilder/index.test.ts
@@ -211,6 +211,8 @@ describe('OpenAPI builder', () => {
 				PathName: 'MutationPath',
 				OperationType: OperationType.MUTATION,
 				AuthenticationConfig: { required: true },
+				VariablesSchema: personSchema,
+				ResponseSchema: personSchema,
 			},
 			{
 				Name: 'Subscription',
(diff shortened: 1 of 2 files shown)


packages/sdk/src/openapibuilder/index.test.ts
```diff
@@ -211,6 +211,8 @@ describe('OpenAPI builder', () => {
 				PathName: 'MutationPath',
 				OperationType: OperationType.MUTATION,
 				AuthenticationConfig: { required: true },
+				VariablesSchema: personSchema,
```
acdn-tsmith, line 214: While running the tests, I noticed that this `annotate operations` test will throw an error in the changed `mutateOperations` function due to not having a `VariablesSchema`. Adding a dummy `VariablesSchema` and `ResponseSchema` now.

